PR #9, opened 2018-08-06
Rewrote the contributors page. Improved sorting.

ewels:
I refactored how the main contributor institutes are listed on the About page. Now uses YAML and looks a lot nicer. Includes colour logos as well as white (will update later to cycle the logos on the homepage).

Sorting pipelines can now be done with stars. Double click a sorting type to get the reverse.


apeltzer:
Do you want us to have a look once you#re done or are you intending to work on this more?

PR #12, opened 2018-08-09
Mobile / small screen fix for pipeline & contributors cards

ewels:


ewels:
Appended: Sort the releases by publication date so that we actually show the last release.

Issue #16, opened 2018-08-14
Missing nf-core create subcommand docs and template branch


sven1103:
Currently the docs do not mention the nf-core tools subcommand `create` and the documentation about how to setup the git repo correctly with an additional `TEMPLATE` branch is not referenced anywhere in the docs.

ToDo:

- [x] Add docs for subcommand `create`
- [x] Link to docs for proper git setup

alneberg:
I think you finished the first task and #19 should finish the second one.

PR #17, opened 2018-08-14
Reference create subcommand

sven1103:
Addresses #16

### DIFF
--- a/markdown/tools.md
+++ b/markdown/tools.md
@@ -16,6 +16,45 @@ Alternatively, you can install the development version directly from GitHub:
 pip install --upgrade --force-reinstall git+https://github.com/nf-core/tools.git
 ```
 
+## <a name="create"></a>Create a pipeline from scratch 
+To create a pipeline from scrath, use the command `nf-core create`.
+
+You should see an output like this:
+
+```
+$ nf-core create
+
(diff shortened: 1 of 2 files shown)


sven1103:
No clue, if this works :D

markdown/tools.md
```diff
@@ -16,6 +16,45 @@ Alternatively, you can install the development version directly from GitHub:
...
+```
+
+If you have already initialized a Git repository, you can pass the option flag `--no-git`. Otherwise,
+nf-core tools will create a Git repository via `git init`. You still have to add remote tracking by yourself though.
+
+One very important setup is the creation of a branch `TEMPLATE`, right after the initialisation
+from the nf-core template. We will use this branch to keep pipelines in sync with future changes
+in the template. 
+
+Please follow the [template branch](template_branch.md) description.
```
ewels: Links like this won't work on the website, as the markdown is rendered dynamically. I haven't yet added the `template_branch.md` docs onto the rendered web pages yet though, so will fix this when I do so (and have a working URL to switch it with).

ewels:
I think we should probably move the docs for the nf-core/tools commands into that repo (this is already done for the linting errors so isn't much work). We can do this in a future PR though, this looks great :)

Issue #25, opened 2018-08-21
Add a map showing location of collaborators


ewels:
Our nf-core collaborators are truly global - it would be fun to have a map indicating this!

matq007:
I think we can close the issue

MaxUlysse:
@matq007 You know you can use close #25 in your PR or even in your commits to have the issue closed when merging the PR?
cf: https://help.github.com/articles/closing-issues-using-keywords/

PR #33, opened 2018-09-08
Implemented contributors map

matq007:
I've implemented the contributors' map using Leaflet on About page as mentioned in Issue #25. See the picture below. What do you think? 
screencapture-localhost-8888-about-2018-09-08-13_13_49


public_html/about.php
```diff
@@ -12,7 +12,8 @@
 // Parse YAML contributors file
 require_once("../Spyc.php");
 $contributors = spyc_load_file('../nf-core-contributors.yaml');
-$contributors_html = '<div class="card-deck">';
+$contributors_html = '<div id="contribution_map"></div><br>';
```
ewels: Minor thing, but I have a pet-hate against using line breaks for spacing: CSS is better 😉 Especially as you're already adding CSS styles for this element.

nf-core-contributors.yaml
```diff
@@ -21,6 +21,7 @@ contributors:
       contact: Phil Ewels
       contact_email: phil.ewels@scilifelab.se
       contact_github: ewels
+      location: [59.3505174, 18.0221508]
```
ewels: Question: is it possible to locate the pins based on an address search, instead of having to specify coordinates? Minor convenience thing only, so only worth it if it can be done easily.

matq007: After some time, I've found a way how to do this.

public_html/about.php
matq007: Fixed in new commit

matq007:
Fixed:
1. map pins are now green
2. On click, you can see the full name of the institution
3. No more coordinates
With the last one, I've used [nominatim](https://wiki.openstreetmap.org/wiki/Nominatim) which is free, but we have to careful when adding the address. The API has to find it. Some addresses were modified so they could be found.

includes/header.php
```diff
@@ -21,6 +21,10 @@
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-68098153-2"></script>
     <script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);}  gtag('js', new Date()); gtag('config', 'UA-68098153-2'); </script>
+    <!-- Leaflet maps -->
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.4/leaflet.css" rel="stylesheet">
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/leaflet/1.3.4/leaflet.js"></script>
+    <script src="assets/js/yaml-0.3.0.min.js"></script>
```
ewels: Can we drop this dependency now?

public_html/about.php
```diff
@@ -7,12 +7,14 @@
...
 $markdown_fn = '../markdown/about.md';
 $no_print_content = true;
+$locations = [];
 include('../includes/header.php');
 
 // Parse YAML contributors file
 require_once("../Spyc.php");
 $contributors = spyc_load_file('../nf-core-contributors.yaml');
-$contributors_html = '<div class="card-deck">';
+$contributors_html = '<div id="contributors-map"></div><br>';
```
ewels: Cheeky `<br>` remains 😉

public_html/about.php
```diff
@@ -62,8 +70,52 @@
     $contributors_html .= '</div></div>';
 }
 $contributors_html .= '</div>';
+$contributors_html .= '<div style="display: none;" id="locations">' . json_encode($locations) . '</div>';
```
ewels: This is fine, but I'd usually do it as a `script` block instead:

```php
$contributors_html .= '<script type="text/javascript">var locations = '.json_encode($locations).'</script>';
```

Then below you can just access the `locations` variable directly, instead of having to call the `parseJSON` function.

public_html/about.php
```diff
@@ -62,8 +70,52 @@
...
+                var lon = parseFloat(marker[0].lon);
+                latitude += lat;
+                longitude += lon;
+                L.marker([lat, lon], {icon: greenIcon}).addTo(map).bindPopup(location.full_name);
+                promise.resolve();
+            });
+            promises.push(promise);
+        });
+        
+        $.when.apply(this, promises).then(function() {
```
ewels: Bonus points for proper use of deferred promises 😎👍

ewels:
The more I think about this, I'm wondering if your original solution with the lat / long coordinates is better. We don't actually use the address anywhere at the moment I think, and the lat long gives simpler and more robust code. Perhaps we can just drop the physical address and have only lat/long.

Sorry for saying this after you've spent time looking into the searching thing! What are your thoughts?

MaxUlysse:
I think the lat / long coordinates is a good idea.
Often institution prefer to have a non regular address or a PO box, so it should be easier to display on a map with just coordinates.

matq007:
Coordinates are back with some code cleanup :)

public_html/about.php
```diff
@@ -40,6 +42,12 @@
             $contributors_html .= '</a>';
         $contributors_html .= '</h6>';
     }
+    // Address
+    if(array_key_exists('address', $c) && array_key_exists('full_name', $c)){
```
ewels: Should be `array_key_exists('location', $c)`

public_html/about.php
ewels: Bump

nf-core-contributors.yaml
```diff
@@ -56,42 +58,47 @@ contributors:
       contact: Andreas Wilm
       contact_email: wilma@gis.a-star.edu.sg
       contact_github: andreas-wilm
+      location: [1.3025202, 103.7904214]
 
     - full_name: The Competence Centre for Genome Analysis Kiel
       short_name: CCGA Kiel
       region: Kiel University, Germany
-      address: Christian-Albrechts-Platz 4, 24118 Kiel, Germany
+      address: Christian-Albrechts-Platz 4, 24118 Kiel, Gernamy
```
MaxUlysse: Gernamy? not sure about this country

public_html/about.php
matq007: Good catch, totally forgot about that

PR #32, opened 2018-09-08
add SLU to contributors

HadrienG:

PR #34, opened 2018-09-09
Reduce padding on SLU logo svgs

ewels:
Update SLU logos from #32 to remove massive white padding.


ewels:
Minor change, so self-merging.. 😱

PR #35, opened 2018-09-11
About map: minor additions

ewels:
Sorry, I couldn't help myself...

* Made leaflet libraries local instead of CDN
* Moved JS include to the footer
* Added institute logo to map popover
* Added HTML anchors to the contributor cards
* Made map popovers link to the contributor card

image

Note that the anchor links don't work brilliantly because they're hidden by the floating menu. There is a CSS workaround for this but I can't remember it off the top of my head. We can always revisit this in a future PR.


public_html/about.php
```diff
@@ -68,6 +65,15 @@
     $contributors_html .= '</div>';
     // Close card div
     $contributors_html .= '</div></div>';
+
+    // Location JSON
+    if(array_key_exists('location', $c)){
+        $location['location'] = $c['location'];
+        $location['full_name'] = array_key_exists('full_name', $c) ? $c['full_name'] : '';
+        $location['image_fn'] = $img_path;
```
matq007: We should include also `array_key_exists('image_fn')` for an image, because right now The Competence Centre for Genome Analysis Kiel is missing one.

public_html/about.php
```diff
@@ -94,16 +100,19 @@
...
             attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
         }).addTo(map);
-        
+
         locations.forEach(function(marker) {
             if (marker != null) {
                 latitude += marker.location[0];
                 longitude += marker.location[1];
-                L.marker(marker.location, {icon: greenIcon}).addTo(map).bindPopup(marker.full_name);
+                L.marker(marker.location, {icon: greenIcon}).addTo(map).bindPopup(
```
matq007: Actually, the fix should be here if we are missing an image.

public_html/about.php
ewels: Good spot 👍

public_html/about.php
ewels: I moved this markup into the PHP code so that it's handled in the same logic.

MaxUlysse:
I'm sorry I'm late in this, but wouldn't it be better if we center the logo horizontally in the popover?

ewels:
Sure - feel free to open a PR 😉

PR #36, opened 2018-09-19
Update nf-core-contributors.yaml

MaxUlysse:
- Add twitter information


MaxUlysse:
Someone else just need to sort out the PHP now ;-)

PR #55, opened 2018-11-26
Ignore nf-core/configs repo

ewels:
Don't show the configs repo on the pipelines page.

### DIFF
--- a/ignored_repos.ini
+++ b/ignored_repos.ini
@@ -6,6 +6,7 @@ repos[] = "nf-co.re"
 repos[] = "tools"
 repos[] = "logos"
 repos[] = "test-datasets"
+repos[] = "configs"
 
 [ignore_topics]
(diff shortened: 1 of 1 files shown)


ewels:
Confident enough in this to break the rules and self-merge.

ewels:
Crap, it didn't work. Why not?? I'm going to leave it a little while and hope that it magically fixes itself.

ewels:
Looking at the server I think that we probably hit the API limit. The update script seems to be unable to fetch the list of URLs when it works fine for me. So hopefully this will magically fix itself tomorrow when it tries to fire again for some event.

ewels:
..changed the web hook to fire on forks, releases and new stargazers. Hopefully that will mean it runs a lot less frequently than on every push event, but still keep the website up to date.

PR #63, opened 2018-12-07
No nohup, move log file out of git and append.

ewels:
Hopefully this will fix the automatic website updates.


ewels:
Cool. Updated manually this time, but then redeployed the webhook payload and I think it worked. So fingers crossed, should work automatically again in the future 👍

PR #69, opened 2019-01-07
Deploy: split pipelines update into its own script because it's slow.

ewels:


ewels:
self-merging :shipit:

PR #91, opened 2019-04-05
Mini fixes, now all URls should be fine again :+1: 

apeltzer:


apeltzer:
Yeah, I just found this reference to a `base.config` on this page that leads to another 404: 

https://github.com/nf-core/nf-co.re/blob/master/markdown/usage/installation.md#pipeline-configuration

so thought adding one is better than nothing, though I could add such a file as markdown (with inline code) and make it this way better?

apeltzer:
Changed accordingly to use a markdown file instead 👍

ewels:
This still won't do anything though, as `conf` isn't a magic directory like `usage` and `developers`. I think it's better to put the code block in-line in the documentation file that references it.

apeltzer:
Done :-)

PR #129, opened 2019-07-11
Update FontAwesome version and development pipeline icon.

ewels:
Slightly less negative / alarming icons for pipelines that are under development.

Before:

After:


drpatelh:
Agreed!

PR #133, opened 2019-07-18
Add in the Troubleshooting section for users

apeltzer:


drpatelh:
Thanks @apeltzer ! 

Do all of the files in here have a tab?
https://github.com/nf-core/nf-co.re/tree/master/markdown/usage

May be worth adding them in too.

apeltzer:
Good question - I'm open to have some more items in there!

ewels:
Will merge now, can add additional links as well though if we want - I guess we should! May want to have a bit of a think about nice organisation of all this material at some point..

PR #135, opened 2019-07-20
First draft of new page for individual pipelines

ewels:
Refactored how some of the dynamic pages are loaded - usage and developer docs. This allows me to set up more dynamic URL routing to handle individual pages for pipelines.

Currently the individual pipeline page doesn't do much - just grabs the `README.md` file remotely from GitHub and renders that. There will probably be a bunch of bugs from people putting weird stuff in there, but it seemed to work ok on the ones I tried. I've also updated the pipelines page to link to the individual pages instead of GitHub directly.

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -147,6 +147,15 @@ if( isset($markdown_fn) and $markdown_fn){
     },
     $content);
 
+    // Prepend to src URLs if configureds and relative
+    if(isset($src_url_prepend)){
+      $content = preg_replace('/src="(?!https?:\/\/)([^"]+)"/i', 'src="'.$src_url_prepend.'$1"', $content);
+    }
+    // Prepend to href URLs if configureds and relative
+    if(isset($href_url_prepend)){
+      $content = preg_replace('/href="(?!https?:\/\/)([^"]+)"/i', 'href="'.$href_url_prepend.'$1"', $content);
+    }
+
 }
 
 if(isset($title) and $title): ?>
--- a/includes/header.php
+++ b/includes/header.php
@@ -78,7 +78,7 @@ if( isset($markdown_fn) and $markdown_fn){
     }
     // Prepend to href URLs if configureds and relative
     if(isset($href_url_prepend)){
-      $content = preg_replace('/href="(?!https?:\/\/)([^"]+)"/i', 'href="'.$href_url_prepend.'$1"', $content);
+      $content = preg_replace('/href="(?!https?:\/\/)(?!#)([^"]+)"/i', 'href="'.$href_url_prepend.'$1"', $content);
     }
     // Find and replace HTML content if requested
     if(isset($html_content_replace)){
--- a/includes/header.php
+++ b/includes/header.php
@@ -182,13 +182,13 @@ if(isset($title) and $title): ?>
           } ?>
           <h1 class="display-3"><?php echo $title; ?></h1>
           <?php if($subtitle){ echo '<p class="lead">'.$subtitle.'</p>'; } ?>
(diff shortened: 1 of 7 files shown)


drpatelh:
Amaaaazing! I have no idea what I'm reviewing here so I'll side step. Would it be worth linking `usage.md` and `output.md` somehow (tabs/links) while you're doing this?

ewels:
Ok, just done a bit more stuff: now repo sub-pages in markdown should also work - not _just_ `README.md`. So for example https://nf-co.re/hlatyping/docs/output now works as well as https://nf-co.re/hlatyping

Images should now also load, and relative links are fixed.

ewels:
> Would it be worth linking `usage.md` and `output.md` somehow (tabs/links) while you're doing this?

This could be a nice idea! They now work when you get there from a link in the readme, but this is definitely one way that we could improve the UI. Any other ideas? I also threw together what's there pretty quickly so am happy for feedback / suggestions.

ewels:
Ok, subnav thingy added! Plus a bunch of other tidying, testing and bugfixing. I think this can pretty much be merged now...

image

I think the header is a bit deep still, so maybe need to play with moving the GitHub link or something th thin it out..? But it's trivial stuff now I think.

drpatelh:
Good work. Looks really good :+1: Wouldn't expect anything less :smile:

drpatelh:
Going to merge then based on pretty pictures :+1:

ewels:
Nice! Hah, finding a couple of things with fresh eyes on the website.. Like on pipelines without a release:

image

...I'll open a new PR! 😉

drpatelh:
https://nf-co.re/atacseq/docs/output

Heads up. Quick look on phone revealed that you can't see images from `output.md` too.

drpatelh:
Wait. It would help to actually have images embedded in the page. It's fine.
https://nf-co.re/chipseq/docs/output

As you were.

ewels:
Hah, yes that does help! I was thinking that I should add a link at the bottom to "View this page on GitHub". Would make it easier to check things like this, to see what's expected.

drpatelh:
Maybe we need to change the links in the template to point to the website for `usage.md` and `output.md` too?

ewels:
Ok, I added a "Read this on GitHub" subfooter to basically all website pages over in https://github.com/nf-core/nf-co.re/pull/137 (e40ff74d6472d6b6d9c85d345421151265c6c0a2)

ewels:
> Maybe we need to change the links in the template to point to the website for `usage.md` and `output.md` too?

Not sure I understand, which links? All relative links (ie. not starting `http(s)://`) are fudged when the markdown is rendered on the website, so they _should_ all work there without changes.

drpatelh:
The ones in this section:
https://github.com/nf-core/tools/tree/master/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D#documentation

drpatelh:
Most point to the website already.

ewels:
There is no need for this - it's already set up as it should be.

These links on the readme are a mixture of "core docs" - one markdown file which all pipelines should refer to (eg. how to install nextflow) and "pipeline docs" which are markdown files that are bundled with the repo.

The core docs should have complete, absolute paths - eg. https://nf-co.re/usage/installation - this is because all pipelines need to go to exactly that page.

The pipeline docs should have relative paths - eg. `docs/usage.md` - this is because that link is the same for every pipeline but refers to the usage documentation _for that pipeline_. So it needs to be relative to the file that it's written in.

Both sets of links should already work properly on both GitHub and the nf-co.re website.

we're good

PR #136, opened 2019-07-21
Pipeline pages: Moar awesomeness

ewels:
On the live site there is a noticeable delay when loading the pipeline pages, as the web server pulls the markdown file from github before rendering. This PR adds caching, so that the file is saved to the server once pulled once. On subsequent page loads it's rendered from that file directly, meaning we don't hit GitHub as much and the load time is much much better. Caching is based on the last release timestamp, or push event if there are no releases.

Also, I put in a warning and a button for in-dev pipelines, instead of just having the text `fubar` 😆

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 config.ini
 public_html/pipelines.json
+markdown/pipelines
 update.log
 
 access-logs
--- a/README.md
+++ b/README.md
@@ -22,17 +22,27 @@ To make edits to the website, fork the repository to your own user on GitHub and
 
 **IMPORTANT:** The repo has git submodules, so remember to use the `--recursive` flag:
 
-```
+```bash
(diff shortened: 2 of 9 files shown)


ewels:
Ok, I got a bit overexcited and have extended this quite a bit. Now, it uses the GitHub API to list all files in the repo and using this, builds a tree of all markdown files (excluding the boring ones). It uses this to build a table of contents. This step is also cached, to avoid rate limiting problems.

Example:
image

---

For Sarek, which has loads of markdown files:

image

ewels:
Note that I had a future search implementation in the back of my mind for this - it should be pretty easy to extend so that it fetches all of the markdown files to the local cache when it first discovers them. They would then be easily available for the search function to grep.

ewels:
PR should be ready to merge ✨

drpatelh:
> PR should be ready to merge

Ok then :+1: :+1:

PR #137, opened 2019-07-21
New: website search functionality

ewels:
Added new search bar on the homepage:
image

This leads to a new search results page:
image

The following is currently searched:
* Pipelines
  * Title
  * Keywords
  * Description
* "Core" documentation (markdown files in `markdown/usage` and `markdown/developers`

The search string is highlighted in the results, and also when you click through to the page all instances should be highlighted too.

Note that the new pipeline markdown files from earlier today are _not_ yet searched - this would be an easy addition, but really needs a bit of refactoring so that we pull all markdown files from all pipelines so that the search is complete (currently it would only search the pages that someone, anyone, has already viewed on the website).

I'd also like to add an AJAX search results dropdown to the search bar on the homepage, so that you don't have to go to the results page but can jump directly.

I think that both of these can go in another PR as they're non-essential and I don't know when I'll get to them.

Other feedback / suggestions / bug findings welcome!

### DIFF
--- a/includes/footer.php
+++ b/includes/footer.php
@@ -1,48 +1,56 @@
-<?php if(isset($title) and $title): ?>
-      </div>
-    </div>
- <?php endif; ?>
+<?php
+if(isset($title) and $title){
+  echo '</div></div>';
+}
 
+// Subfooter, if given
+if(isset($md_github_url) and $md_github_url){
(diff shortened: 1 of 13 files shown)


drpatelh:
Thanks for doing this so quickly 👍

PR #139, opened 2019-07-22
Collect and display statistics for nf-core

ewels:
Whilst trying to collect stats about nf-core for the umpteenth time, I thought to myself: _"Why not automate this process?"_

So, I did:

localhost_8888_stats

Some cool things:

GitHub has nice graphs of visitors and clones which I often use in presentations - however, it only keeps data for the past two weeks:
image

With this PR, we now have a script which we can run every week on a cronjob to get these numbers via the GitHub API and save for perpetuity. It's these totals that are shown in the table above. In the future we can extend this to log the number of stargazers and forks _over time_, so we can see the growth of repos.

For future work, I'd like to use some of the numbers currently sitting in the background. Ideally every pipeline will have a new stats page with a load of graphs. We can also have nice contributor statistics now too, with a leaderboard of top contributors! It's going to be very pretty.

I think this PR can be merged. However, I'd like to keep it vaguely secret for a short while (eg. only visible if you know the URL: https://nf-co.re/stats) until we're sure that it's working as expected and not crippling the webserver or anything else nasty.

Phil

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 config.ini
 public_html/pipelines.json
+nfcore_stats.json
 markdown/pipelines
 update.log
 
--- a/README.md
+++ b/README.md
@@ -42,6 +42,16 @@ php update_pipeline_details.php
 This will create `public_html/pipelines.json`, which is used by the website.
 Note that this is ignored in the `.gitignore` file and will not be tracked in git history.
 
+Optionally, once you've done that, you can grab the pipeline traffic statistics:
+
+```bash
+php nfcore_stats.json
+```
+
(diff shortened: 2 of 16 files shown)


drpatelh:
Amazing! I have been using exactly this information in my presentations too to give people an idea of the usage of `nf-core` pipelines. Really nice that this will have be in one place now :fireworks:

ewels:
And always up to date too!

ewels:
Ok - I ran the update script on the server manually and the page works now: https://nf-co.re/stats

I've set up a cronjob to run it automatically once a week in the future.

PR #140, opened 2019-07-24
Stats: More awesomeness!

ewels:
Fairly major update to the stats page, with quite a bit of refactoring:

* Now pull stats for "core" (non-pipeline) repos as well as just pipelines
* Stats page refactored to show all of these stats in two separate tables
* Stats collection page now saves stuff like forks and stars with a timestamp, so we can track growth over time
* We now collect the number of slack users, including whether they are currently active. Again this is with a timestamp so we can track growth.

The update scripts will have to be run on the server manually again once this is merged. I think it should be good to go though.

Phil


ewels:
Oh, I also updated the bootstrap framework so that I could mess around with toasts.

In fairness, some of the code I added at the start to handle new pipelines shouldn't be needed now, as this is more decouples from the `pipelines.json` file than at first. I might look in to whether I can remove some of this cruft in a while.

drpatelh:
Merging without knowing what I'm merging again :+1: :+1:

ewels:
Thanks @drpatelh! I think I might start just pushing to master with these :)

PR #145, opened 2019-07-30
Fix slack url

MaxUlysse:
As @drpatelh pointed out, we're not the Near Field Core ;-)


ewels:
Oops, sorry! Just noticed and fixed this myself without a PR 😞

PR #146, opened 2019-07-30
Logo: New tool to generate repo logo PNGs

ewels:
Ok, I'm trying to be good with PRs again 😇 

New thing! For a while it's been in the back of my mind that it should be pretty easy to write a mini PHP script to generate repo logos with the correct font. So I had a quick play and did it.

Once merged, https://nf-co.re/logo/whateveryouwant will show the following image:

nf-core-whateveryouwant

By default, it sanitises the input, so https://nf-co.re/logo/MyH0rribLE//name gives:

nf-core-nf-core-myhrriblename png

But, if you _really_ want it, you can just add `?F` to the URL: https://nf-co.re/logo/MyH0rribLE//name?f

nf-core-nf-core-yrribname png

---

I figure that, once merged, we can add in functionality to the `nf-core create` command to automatically fetch the logo for the given pipeline name and add it when the pipeline is first created! 🎉


lpantano:
so cool!

drpatelh:
BOOOOOOOOM!

drpatelh:
So the logos are currently called like `nfcore-rnaseq_logo.png` in most repos. Can we keep this naming format?

Also, the million dollar question...: 🥁 Can we generate a smaller version to place in the appropriate directory to attach to emails too? Then we can finally close:
 https://github.com/nf-core/tools/issues/316

drpatelh:
Does this also mean we will need to host a wake for `nf-core/logos`?

ewels:
Sure, I’ll do both of those things tomorrow :+1: Should he easy!

No need to retire it - still good to have a place to keep the `.ai` files and versions for a dark background / square / etc etc. But maybe can remove or rewrite some of the docs and remove the scripts for logo stuff?

ewels:
Ok, default filenames now follow the pattern used elsewhere: `nfcore-xxxx_logo.png` (this wasn't really important for the create script, as it's basically only used when saving from the browser with a default filename, but it's still nice).

Also, there are two new URL flags: `?s` and `?w`. The first scales the image to be 400px wide, which is roughly the size of the email logos that we've made so far. The second lets you specify any width you want, _eg._ `?w=800` will give you a logo that's 800px wide. The height of the image is automatically calculated.

Only one can resize flag can be given at a time; `w` takes precedence. If specified with another flag such as `f`, the second needs an `&` instead of a `?`: `/logo/name?f&s`

apeltzer:
Looks fab 🥇

ewels:
Thanks!

PR #149, opened 2019-07-31
Contributors: Clean up affiliations

ewels:
As the list of contributing institutions gets longer, the use of the `affiliation` field gets more and more random. Here I've attempted to clean these up a bit. Some just repeated information (eg. the short name with the same URL), so I just deleted those. Some were a little on the verbose side 😀 

Note that when I first wrote this, my intention was for this to show the larger institute - eg. university and so on, that the contributor belongs to. See the first two: NGI and QBIC. Later on, many people have put the larger institute as the main name and used the affiliation to denote a smaller group within that institute. I've left this for now, as I figure it's fine - but we may want to standardise this at some point..? Feedback welcome.

Anyway, please sanity check my changes - especially if I've modified one of your institutes!

### DIFF
--- a/nf-core-contributors.yaml
+++ b/nf-core-contributors.yaml
@@ -112,14 +112,16 @@ contributors:
       location: [52.0797204, 0.1833987]
       twitter: sangerinstitute
 
-    - full_name: Cancer Research UK Beatson Institute & ICS University of Glasgow
+    - full_name: Beatson Institute & Institute of Cancer Sciences
       short_name: BICR
+      affiliation: Cancer Research UK, University of Glasgow
       description: >
         The Beatson Institute is one of Cancer Research UK’s core-funded institutes.
         They have built an excellent reputation for basic cancer research, including world-class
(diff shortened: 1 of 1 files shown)


lpantano:
I understand the issue. I am guessing the issue is if two departments of the same institute are here, and they do/maintain a different kind of pipelines in parallel. Is there any way to keep just the institution and in other place, people in the institution contributing? so more than one person can be listed. Or like a leader and then others. I am happy to take out my affiliation, I copy/paste from the previous one :)

ewels:
Yeah, I was wondering about adding the ability to list more than one person. That would be even more of a hassle to maintain than it already is though.. To be honest, I'm not too fussed - my thinking is that if the person listed at least knows a little about nf-core then that's sufficient credentials 😀

apeltzer:
I agree with @ewels here - it would anyways just serve as a point of contact for people interested in participating/contributing/using nf-core pipelines, so should be fine 👍

PR #156, opened 2019-08-06
Stats views/clones fixes

ewels:
* Fixed titles and axes to say day instead of month
* Added label to make clear that the totals are from a certain date
* Configured axes to go to minimum 1-day labels

Fixes nf-core/nf-co.re#153


drpatelh:
:+1: Do you think it's also worth clarifying the totals at the bottom of each plot i.e. "Unique cloners over period"?

ewels:
Hmm, maybe "Unique cloners since xxx" would be simpler 🤔

ewels:
Yeah, I think that's better:
image

PR #163, opened 2019-08-11
Issue stats fetch refactor, include issue in GH contribs

ewels:
- Cache all comment results and regenerate stats each run
- Include issues when counting contributors
- Update contributors plot to show who's making commits and who is making issues

Issues are important contributions too, right? We now have 205 contributors instead of 71.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ public_html/pipelines.json
 nfcore_stats.json
 nfcore_issue_stats.json
 markdown/pipelines
+api_cache
 contributor_stats
 update.log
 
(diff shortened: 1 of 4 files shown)


drpatelh:
> Issues are important contributions too, right? We now have 205 contributors instead of 71.

You read my mind. 71 definitely an under-representation of those who are actually just contributing via GitHub without joining the organisation 👍

ewels:
71 is the number of different github users who have at least one commit in one repository. It’s separate to the organisation members, though the number is similar (the two circles don’t quite overlap on the Venn diagram though)

PR #167, opened 2019-08-13
Stats: Make stacked area charts overlap colours nicely.

ewels:
This is a pretty trivial one purely for aesthetics. Stops stacked area plots from having the colour show through under the other datasets.

Changes this:
image

---

To this:
image


ewels:
Self-merging as I need this for figure generation

PR #174, opened 2019-09-03
Introduces versions for the parameters.schema.json

sven1103:


sven1103:
Trying to avoid https://github.com/nf-core/tools/pull/368 test failures in the future, because the schema evolves.

sven1103:
@ewels don't know how the public_html is deployed on the website server. Are the version directories I just created paths in the website?

ewels:
> @ewels don't know how the public_html is deployed on the website server. Are the version directories I just created paths in the website?

Yes - it should deploy automatically with the relevant URL path.

ewels:
Where is this schema file referred to from? I can't find it in https://github.com/nf-core/rnaseq/blob/master/parameters.settings.json or https://github.com/nf-core/tools/blob/62f9a129cf2131c5dc29d974d4b949c85ff481eb/nf_core/launch.py

ewels:
Ok, merged. Let's see what breaks...

sven1103:
> Where is this schema file referred to from? I can't find it in [rnaseq/parameters.settings.json](https://github.com/nf-core/rnaseq/blob/master/parameters.settings.json) or [tools/launch.py](https://github.com/nf-core/tools/blob/62f9a129cf2131c5dc29d974d4b949c85ff481eb/nf_core/launch.py)

Yes, something that has bothered me for quite some time. There is no standard you can stick to (at least to my knowledge). We could introduce a required property in the JSON though, that links to the URL of the implemented JSON schema. With that, we have more control and transparency. 

We could parse this property during linting and validate against the corrected schema. I don't think that we need to enforce a schema version yet. 

The more strict way would be to restrict the parameter schema version in the nf-core tools release.

PR #192, opened 2019-09-22
Turn debug statements off for issue stats.

ewels:
Minor change, quick merge appreciated. The updates log file is getting quite large, quite quickly... My bad 😳


ewels:
Some of the API calls for updating the pipeline were being rejected due to rate limiting. As we already have authentication for GitHub APIs in other scripts, I figured we should use it here too. Pushes up the number of requests we're allowed massively.

PR #201, opened 2019-10-08
Events: Allow details section as well as description.

ewels:
Because the closest event to now should be at the top of both lists.

Also make the modal window for details bigger.


ewels:
Also hopefully fixed the automatic tweets on pipeline releases..

PR #205, opened 2019-10-11
Add a section about memory limits in the introduction

sidorov-si:
I added a section about memory limits into the Introduction.

### DIFF
--- a/markdown/usage/introduction.md
+++ b/markdown/usage/introduction.md
@@ -135,7 +135,7 @@ So, the amount of memory available to this user on this machine is 8 GB.
 
 > NB. We recommend to have at least 4 GB available. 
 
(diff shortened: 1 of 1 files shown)


sidorov-si:
I had to remove the main header:
---
title: Introduction
subtitle: Get to grip with the key concepts used in nf-core pipelines.
---
and replace it with a mock top-level header to get the checks passed.
But this now needs to be fixed somehow.

ewels:
Many thanks for this @sidorov-si and sorry that it's taken so long for a response. I've just had a look over this and whilst it's a good resource, I'm not sure if it really fits with the docs. Firstly, it's quite in-depth docs about nextflow and not nf-core - generally I try to keep the nextflow stuff fairly light and recommend instead that people refer to the main nextflow site for things like this. Secondly, I'm not sure that it's general enough. For example, none of the commands work on my Mac.

Instead, I wonder if you could rework this and submit it to the Nextflow documentation? Then we can refer to that from our docs here.

Let me know what you think,

Phil

PR #211, opened 2019-10-19
Pipeline health page

ewels:
* [x] Page to display statuses
* [ ] Check all statuses
* [ ] Button to fix errors where possible

Closes https://github.com/nf-core/nf-co.re/issues/210

### DIFF
--- a/public_html/pipeline_health.php
+++ b/public_html/pipeline_health.php
@@ -22,6 +22,7 @@ define('GH_API_OPTS',
       'method' => 'GET',
       'header' => [
         'User-Agent: PHP',
+        'Accept:application/vnd.github.mercy-preview+json', // Needed to get topics (keywords) for now
         "Authorization: Basic $gh_auth"
       ]
     ]
@@ -46,7 +47,6 @@ class RepoHealth {
   // Test result variables
   public $repo_wikis;
   public $repo_issues;
-  public $repo_sponsorships;
   public $repo_merge_commits;
   public $repo_merge_rebase;
   public $repo_merge_squash;
@@ -54,7 +54,6 @@ class RepoHealth {
   public $repo_keywords;
   public $repo_description;
   public $repo_url;
-  public $repo_social_preview;
   public $team_all;
   public $team_core;
   public $branch_master_protection;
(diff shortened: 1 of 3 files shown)


ewels:
Current state of play:

localhost_23000_pipeline_health

Please don't fix the errors where they are present just yet - I need to be able to test something when I write the code to automatically fix it! 😆

ewels:
Also, although the code isn't doing a stupid number of API calls, it's still ridiculously slow. Takes about 30s to load. So need to optimise this quite a bit - probably by caching results and running on a cron job I guess. That's what I do everywhere else.

public_html/pipeline_health.php
```diff
@@ -0,0 +1,281 @@
+<?php
+$title = 'Pipeline health';
```
drpatelh:
```suggestion
$title = 'Repository health';
```

public_html/pipeline_health.php
```diff
@@ -0,0 +1,281 @@
...
+$subtitle = 'Check repository settings for all nf-core pipelines';
```
drpatelh:
```suggestion
$subtitle = 'Check GitHub repository settings for all nf-core pipelines';
```

drpatelh:
So cool :+1: Thanks! Can we include the core repos in this too? I remember we missed some of these last time we imposed the merge restrictions to `dev` and `master`.

ewels:
Yup! Back end code is already written for that, just didn’t get around the printing the results. I even did nice OOP code with class inheritance so that we can easily add pipeline-specific or core-repo-specific tests in the future if we want to :sparkles:

ewels:
Updated screenshot:

localhost_23000_pipeline_health (4)

The table scrolls sideways and is now so wide that it won't fit on my screen for a screenshot 😆 At least, not unless I zoom out super far so you can't see anything:

image

I _think_ that it now tests for everything mentioned in https://github.com/nf-core/nf-co.re/issues/210 .. though you might find some stuff that is reporting weirdly.

The table cells now have hover text to show you what they are for (useful when the table gets so big), and clicking on the table cell opens a new tab with the relevant repository settings page.

Finally, the caching is now implemented, so the page loads in a relatively normal time. However, if you click the magic blue button at the bottom it pulls everything from scratch again. This takes a long time, so I even made it so that you can pull a single repo - fewer API calls, so quite a bit quicker (pulling even one repo refreshes some tests for all repos though, such as team access). I did this after the screenshots above, so you'll have to just imagine a drop down box with repo names next to the blue button.

Ok - I think this is ready to be reviewed (?!?) and / or merged. The page doesn't actually _fix_ anything for you yet, but it does show all of the tests that we discussed which is a good starting point. I can do the code to fix the faults in a separate PR.

Phil

ewels:
Small improvement - I added an option to merge table column headings. So now there's just `Branch protection: master` and the mouse hover tooltips tell you what each specific test is for. I think it's better and it makes the table a lot more manageable:

localhost_23000_pipeline_health (6)

PR #213, opened 2019-10-21
added missing "/" to quick installation guide

FranBonath:
https://github.com/nf-core/nf-co.re/blob/master/public_html/index.php#L146
to make sure people don't just re-name nextflow to bin.


drpatelh:
Good spot @FranBonath !

PR #215, opened 2019-10-31
fix issue counter to only count open issues on pipeline pages

mashehu:
Hi,

I saw that the issue counter on the pipeline pages showed the number of all issues (open and closed). I changed the code to only count open ones. I am not very familiar with php, so I am sorry for my clumsy code.

Before:
Screenshot 2019-10-31 11 17 57

After:
Screenshot 2019-10-31 11 17 49

### DIFF
--- a/includes/pipeline_page/components.php
+++ b/includes/pipeline_page/components.php
@@ -1,4 +1,6 @@
 <?php
+ ini_set('display_errors', 'On');
+ error_reporting(E_ALL | E_STRICT);
 // Build the HTML for a pipeline documentation page.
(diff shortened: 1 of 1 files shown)


includes/pipeline_page/components.php
```diff
@@ -1,4 +1,6 @@
 <?php
+ ini_set('display_errors', 'On');
```
ewels: Please remove these lines.. 🐛

mashehu: Oops, sorry. Fixed now.

PR #216, opened 2019-11-01
Events: softlink modals

ewels:
Changes the URL in the browser when an event modal is launched, adding the new ID attribute as a `#hash`. If such a hash is present when the page is loaded, the modal is shown.

Enables a URL for a specific event to be shared.

### DIFF
--- a/nf-core-events.yaml
+++ b/nf-core-events.yaml
@@ -4,7 +4,8 @@
 
 events:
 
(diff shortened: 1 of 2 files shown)


ewels:
image

ewels:
I need this for event advertising, so will self-merge.. 😀

PR #225, opened 2019-11-08
Remove trailing slashes from URLs

ewels:
eg. https://nf-co.re/tools/ will now redirect to https://nf-co.re/tools, https://nf-co.re/join/ to https://nf-co.re/join etc.


ewels:
This one I actually tested locally. Seems to work. Also means that #224 was kind of unnecessary, as it's now impossible to visit `/join/slack/` 😆

PR #230, opened 2019-12-02
fixed name and changed twitter handle for GIS

jonathangoeke:
I fixed the name of GIS (Genome Institute of Singapore), and I added the Twitter handle for our Institute

### DIFF
--- a/nf-core-contributors.yaml
+++ b/nf-core-contributors.yaml
@@ -64,9 +64,9 @@ contributors:
       url: https://www.a-star.edu.sg/gis
       affiliation_url: https://www.a-star.edu.sg
       image_fn: GIS.svg
-      contact: Andreas Wilm
-      contact_email: wilma@gis.a-star.edu.sg
-      contact_github: andreas-wilm
+      contact: Jonathan Göke
+      contact_email: gokej@gis.a-star.edu.sg
+      contact_github: jonathangoeke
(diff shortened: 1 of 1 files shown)


drpatelh:
I think it's probably best that we add yourself as the main contact too seeing as Andreas no longer works at the Institute. Ping @ewels @apeltzer.

jonathangoeke:
> I think it's probably best that we add yourself as the main contact too seeing as Andreas no longer works at the Institute. Ping @ewels @apeltzer.

sure should I change that?

drpatelh:
Yep. Assuming the email address won't work either so probably best to update it whilst you are making these changes. Just wanted to double-check with the other core members because Andreas was quite heavily involved in setting up nf-core. I think they will agree though so should be fine :+1:

apeltzer:
Yes please adjust this. I doubt Andreas will mind that 👍

ewels:
@andreas-wilm - all ok I assume?

jonathangoeke:
I spoke with Andreas, he is fine with that. I can also add @shihcc and me if 2 main contacts are OK?

apeltzer:
Normally we have only one contact person per institute - its anyways rarely happening that you're contacted via this, so should be fine to have you on it alone 👍

jonathangoeke:
ok done!

PR #238, opened 2019-12-05
Recommand recent genomes

MaxUlysse:
_ie_ `GRCh38` instead of `GRCh37`

### DIFF
--- a/markdown/developers/guidelines.md
+++ b/markdown/developers/guidelines.md
@@ -60,6 +60,7 @@ If possible, it's great if pipelines can also have:
   * Nearly all nf-core pipelines use a conda `env` script to list their software requirements.
     The pipeline Docker images are then built using this, meaning that with a single file your pipeline can support nextflow users running with conda, docker or singularity.
   * The [nf-core template](/tools#creating-a-new-workflow) comes with all required code to support this setup.
+* All pipelines should use by default a recent reference genomes (_ie_ `GRCh38` for human...)
 * Optimised output file formats
   * Pipelines should generate `CRAM` alignment files by default, but have a `--bam` option to generate `BAM` outputs if required by the user.
 * Digital object identifiers (DOIs) for easy referencing in literature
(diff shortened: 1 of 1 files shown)


markdown/developers/guidelines.md
```diff
@@ -60,6 +60,7 @@ If possible, it's great if pipelines can also have:
   * Nearly all nf-core pipelines use a conda `env` script to list their software requirements.
     The pipeline Docker images are then built using this, meaning that with a single file your pipeline can support nextflow users running with conda, docker or singularity.
   * The [nf-core template](/tools#creating-a-new-workflow) comes with all required code to support this setup.
+* All pipelines should use by default a recent reference genomes (_ie_ `GRCh38` for human...)
```
apeltzer:
```suggestion
* All pipelines should attempt to use a recent reference genome draft(_ie_ `GRCh38` for human...)
```

ewels:
I'm a bit confused by this now that I see it. Isn't the choice of reference genome up to the user? The guidelines are more for pipeline developers..

MaxUlysse:
I think it should be there.
It's something that we developers should push, using more recent genomes, especially GRCh38 over GRCh37

ewels:
But it's not up to us developers what people use? I guess this comes down to a minor wording issue maybe. That pipelines should _support the use_ of recent reference genomes..

apeltzer:
Not sure its just for the enduser: Some pipelines we have are relatively tight to use e.g. only GRCH37, GRCH38 supported by downstream tools (`epitopeprediction` for example). 

And I think the statement here is relatively open:

```
 * All pipelines should attempt to use a recent reference genome draft(_ie_ `GRCh38` for human...) 
```

"Should attempt ..." is not enforcing it but just stating that you should aim for current drafts of references when implementing your pipeline. Could be maybe better formulated I agree ...

ewels:
Yeah, I think we're all agreeing here. I was just confused by the wording and thought it meant something else at first.

PR #239, opened 2019-12-05
Adding LNU/EEMiS to organizations

erikrikarddaniel:
I've added EEMiS, part of the Linnaeus University to the contributors.yaml file plus svg images.


ewels:
Testing is only for markdown, and we don't have any markdown in this PR 👍

PR #245, opened 2019-12-06
Fix initial issue update

### LINKED ISSUE #233: update_issue_stats.php throws undefined notices on initial run
If no nfcore_issue_stats.json is present in the project (as it is the case on an inital run) `$ php update_issue_stats.php` returns 
```PHP Notice:  Undefined variable: results in /Users/mitochondrium/nf-co.re/update_issue_stats.php on line 34```

mashehu:
fixes #233


ewels:
Looks good, but also included in the other PR already (you need to make branches earlier :wink:)

PR #246, opened 2019-12-06
Fix header id generation

### LINKED ISSUE #175: Not all links in MD are available the website
I have that in the docs: https://github.com/nf-core/sarek/blob/dev/docs/usage.md#--skipqc
But that won't work in the website: https://nf-co.re/sarek//docs/usage#--skipQC

mashehu:
fixes #175

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -77,10 +77,11 @@ if( isset($markdown_fn) and $markdown_fn){
   // Add ID attributes to headers
   $hids = Array();
   $content = preg_replace_callback(
-    '~<h([1234])>([^<]*)</h([1234])>~Ui', // Ungreedy by default, case insensitive
+    '~<h([1234])>(\S*)</h([1234])>~Ui', // Ungreedy by default, case insensitive
     function ($matches) {
       global $hids;
-      $id_match = strtolower( preg_replace('/[^\w\-\.]/', '', str_replace(' ', '-', $matches[2])));
+      $id_match = strip_tags($matches[2]);
+      $id_match = strtolower( preg_replace('/[^\w\-\.]/', '', str_replace(' ', '-', $id_match)));
       $id_match = str_replace('---', '-', $id_match);
       $hid = $id_match;
       $i = 1;
--- a/includes/header.php
+++ b/includes/header.php
@@ -77,7 +77,7 @@ if( isset($markdown_fn) and $markdown_fn){
   // Add ID attributes to headers
   $hids = Array();
   $content = preg_replace_callback(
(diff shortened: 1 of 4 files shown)


includes/header.php
```diff
@@ -77,10 +77,11 @@
   // Add ID attributes to headers
   $hids = Array();
   $content = preg_replace_callback(
-    '~<h([1234])>([^<]*)</h([1234])>~Ui', // Ungreedy by default, case insensitive
+    '~<h([1234])>(\S*)</h([1234])>~Ui', // Ungreedy by default, case insensitive
```
ewels: Does this work with headers that have whitespace? eg. `<h1>This Header</h1>`?

More generally, why the change?

public_html/assets/css/nf-core-dark.css
```diff
@@ -56,6 +56,13 @@ blockquote {
   background-color: #181A1B;
   border-color: rgba(102, 102, 102, 0.13);
 }
+/* .stats_keynumbers .bg-icon {
```
ewels: Please remove this again if it's not needed..

includes/header.php
mashehu: I had to change it, because ```[^<]*``` didn't catch ```<code>match</code>``` pieces.
Hmm, yes, this actually breaks headers with a whitespace, sorry. 
How about ```~<h([1234])>(.*?)</h([1234])>~Ui``` or is that to general?

public_html/assets/css/nf-core-dark.css
mashehu: Sorry, this was an artifact while creating the before-after screenshots, see commit 
0f06535.

includes/header.php
ewels: Yes I was going to say that it should work, as long as we have the ungreedy `U` modifier so that it doesn't match the whole file...

mashehu: I made it more general now (see 164ceee). 
While testing, I found two different problems with the links from the ToC of the Usage section of bcellmagic (https://nf-co.re/bcellmagic/docs/usage). 
- The `-profile`'s href is to `#-profile-single-dash` 
= The `<li>`s underneath shouldn't appear there, since they are no `<h>` elements. Not sure if that is coming from the repo or the backend, so I don't know where to open the respective issue or fix the problem.

ewels: `-profile` should be one hyphen (it’s a core Nextflow option, not a pipeline param) and the markdown (including ToC) comes from the pipeline. So needs to be fixed there if anywhere I guess..

PR #250, opened 2019-12-11
fix .text-dark a:hover behavior in dark-mode

mashehu:
Found another small dark mode bug.
Before:
Screenshot 2019-12-11 11 16 47
After:
Screenshot 2019-12-11 11 17 32


ewels:
Ah nice, thanks!

PR #265, opened 2020-02-05
Added publications page

ewels:
Closes nf-core/nf-co.re#261

localhost_8888_publications

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -203,6 +203,7 @@ if( isset($markdown_fn) and $markdown_fn){
               <a class="dropdown-item" href="/events">Events</a>
               <a class="dropdown-item" href="/community">Community</a>
               <a class="dropdown-item" href="/stats">Statistics</a>
+              <a class="dropdown-item" href="/publications">Publications</a>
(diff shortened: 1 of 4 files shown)


drpatelh:
Do you think we should add a DOI section for pipelines too? Would be nice to have all of this information findable on the same page rather than clicking countless links to figure out what you should be referencing. 

Could have granular details as to DOI per version too but that may involve fetching these from the Zenodo API if it exists? Or any other method you can think of that will work?

MaxUlysse:
We should already have DOI for all released pipelines

ewels:
I think that manually maintaining all pipeline Zenodo DOIs will be too much work and won't happen.

If we want this (and I think it could be a nice idea for quick reference, though not totally sure how often people will need a big list?) then I guess the best approach is to use the Zenodo API to fetch DOIs based on GitHub repository names. This is kind of unrelated to this PR, so maybe open a new issue for that 👍 

Phil

ewels:
Ok, so can I merge this..? 😃

drpatelh:
Sorry. Thought you were still working on it :+1:

PR #279, opened 2020-02-11
Add to guidelines for deleting master until release

ewels:
Currently, there is a mixture of behaviour when a pipeline is in early development before release. Often the state of the default `master` branch is that it's just an empty template left from `nf-core create`

To make the default code as useful as possible whilst adhering to existing rules of branch management, @MaxUlysse proposed that we _delete_ the `master` branch until the first release, setting `dev` as default. When preparing the first release we create `master` again and set it as the default.

In this PR I add to the guidelines to mention this, as well as adding to the nf-core tutorial.

This topic is open for debate - @nf-core/all please feel free to point out any potential problems with this new approach in the comments below.

TODO if merged:

- [ ] Need new checks on the [Pipeline Health webpage](https://nf-co.re/pipeline_health).
- [ ] Check whether `nf-core sync` will tolerate the absence of a `master` branch (especially for `--all`)

### DIFF
--- a/markdown/developers/adding_pipelines.md
+++ b/markdown/developers/adding_pipelines.md
@@ -192,12 +192,19 @@ will automatically update to list your new pipeline.
 
 ### Branch setup
 
+All nf-core pipelines use branches called `dev` and `master`.
+The `master` branch should contain the code from the latest stable release, `dev` should have the latest development code.
+Before the first release is made we set `dev` as the default repository branch instead of `master`;
+this means that the latest code runs by default up until the first release.
+After the first release we switch the default back to `master`.
+
 We want people to run the latest development code by default up until the first release.
-To do this, we set `dev` as default and delete the `master` branch.
-We will create the `master` branch again when you are ready to create the first release.
+To do this, we set `dev` as the default repository branch.
+After a release is created, we set the default branch back to `master` so that the default
+action is to run the latest stable release code.
 
 Once you have forked the repository, create a new branch called `dev` for the active development.
-In the repository settings, set `dev` to be the default branch and then delete the `master` branch.
+In the repository settings, set `dev` to be the default branch.
 
 ### Repository setup
 
@@ -207,6 +214,10 @@ Remember to configure the repository on the GitHub website with the following:
 * Issues enabled, disable Wiki and Projects
 * A protected `master` branch that requires review and passing tests
 * Write permissions for `nf-core/all` and admin permissions for `nf-core/admin`
+
+You can check that all of these settings are done correctly by referring to your pipeline
+in the nf-core [Repository health web page](https://nf-co.re/pipeline_health).
+This reports the status of various checks and also has the option of fixing errors for you via the GitHub API.
 
 ### Setting up Travis and Docker Hub
 
@@ -223,7 +234,7 @@ that. Then the tests should pass.
 
 ## Making the first release
 
-When the code is stable and ready for a release, create the `master` branch again and set this as the default branch.
+When the code is stable and ready for a release, set the `master` branch to be the default branch again.
 Bump the version numbers on `dev` (see below) and make a pull-request from the `dev` branch to `master` on the nf-core fork.
 This is a special case and the tests should pass.
 Once they do, merge the PR yourself and let the nf-core team know that you're ready.
--- a/markdown/developers/adding_pipelines.md
+++ b/markdown/developers/adding_pipelines.md
@@ -187,13 +187,17 @@ First, fork your workflow repository to the nf-core GitHub organisation by
 clicking 'Fork' at the top of the GitHub webpage. If you don't see nf-core
 as an option, please ask one of the nf-core administrators to do this for you.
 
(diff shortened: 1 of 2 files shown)


ewels:
@olgabot - one half-way method we discussed is to just make `dev` the default branch before the first release, without deleting anything. This is my preferred method - as you say, reduced churn and fewer steps (it's also something we can automate through the GitHub API easily without any risk).

The advantages are pretty much the same - latest code available as default instead of just a blank template; always developing on `dev` and never on `master` (before and after a release).

I guess the argument against this is that having `master` there could cause confusion about what should be on each branch. But as @olgabot says - not having `master` at all could also cause confusion. I'm not sure that either one approach is better than the other, and I think both are probably less confusing than what we do currently.

Thoughts?

drpatelh:
Im happy with that. Confusion either way but probably best to keep it.

Not having  `master` = confusion as to why it isnt there...
Having (pre-release) `master` = confusion as to what its doing there...

kerimoff:
@olgabot has a good point. Although I am not sure which one is the way to go, I think it depends on the target audience. 

If audience is proficient users (of git) then it does not really matter, they will understand what is the point of doing it either way
If audience is beginner users (of git) then it is better to have master there for the reason Olga mentioned. 

Therefore, for general audience I think it is better to change the default branch to dev and not delete the master.

mahesh-panchal:
I would prefer to leave it as is. The less steps one has to perform, the lower the barrier usually is for newcomers. So I agree with @olgabot and @kerimoff.

MaxUlysse:
I would say that our current issue is that we have non-released pipelines that are developed on a `master` branch, some of them have not been touched for months.

For me, if a code is deemed stable enough to be on a `master` branch, it should be released already (even if it is lacking tools or docs).
We do adhere to [semantic versioning 2.0.0](https://semver.org/spec/v2.0.0.html), and I think we should already have more released pipelines, or at least, that's how I understand this part:

> How do I know when to release 1.0.0?
 
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

I can see the problem with an empty `master` branch, or no `master` branch at all, so changing the default to `dev` is I think the least we should do.

ewels:
> I would prefer to leave it as is. The less steps one has to perform, the lower the barrier usually is for newcomers. So I agree with @olgabot and @kerimoff.

To clarify, the proposed changes (setting `dev` to default) means fewer steps for newcomers - newcomer _users_ at least. The latest development code will run by default with `nextflow run` without having to specify `-r dev`

There's two extra steps for developers - setting dev as default and back again, but this can be checked and fixed automatically via the GitHub API so we can do this via the website. Pretty low barrier.

The only barrier I think is to know which branch to make PRs against. But again changing the default to `dev` means that new PRs will select that by default, which is probably what we want.

So we all have the same objective here - to lower the barrier for newcomers. That's what the proposed change is about 😄 It's just a question of which option does that most effectively.

mahesh-panchal:
I thought the target was newcomer nf-core developers, rather than users. 

For a new user who wants to use a pre-release nf-core pipeline, I think having to explicitly set `-r dev` is useful visual signal. 

Perhaps this could also be added to the create template code instructions as well and then removed later when it's made release-worthy. 

My preference would be that the master show some kind of pre-release banner with instructions to use the dev branch, and stay the same as the create template until it's release. 

But then that comes down to teaching developers how to use the tools and work exclusively on dev until a release version is ready.

tdelhomme:
Just to give an additional possibility, in the [iarcbioinfo plateform](https://github.com/IARCbioinfo) we used to maintain 3 branches:
  * **master**=latest release (_e.g._ v1.0)
  * **dev**=development branch with minor changes
  * **vX.Yb**=version branch with major changes (_e.g._ v1.1b, this one is merged with the master to build a new release)
see an example with our [needlestack pipeline](https://github.com/IARCbioinfo/needlestack)

ewels:
Thanks @tdelhomme - that's the same structure that we use for nf-core, except we just use `patch` instead of `vX.Yb`.

@mahesh-panchal 
> For a new user who wants to use a pre-release nf-core pipeline, I think having to explicitly set `-r dev` is useful visual signal.

Yup, this is a good point. However, I worry that instead people just run non-functioning code.

> But then that comes down to teaching developers how to use the tools and work exclusively on dev until a release version is ready.

Agreed. We have said this for a while now, but it isn't really written down anywhere. The documentation changes in this PR aim to address that.

mahesh-panchal:
> Yup, this is a good point. However, I worry that instead people just run non-functioning code.

I would have suggested making the main code then exit if it's not dev or version > 1.0, but that doesn't address the current repositories in their various states. This proposed solution would address that in a simple-ish manner.

ewels:
Ok, so I think we're reaching towards consensus that this is an ok plan..? Make `dev` default on un-released pipelines. Would you be ok with this @olgabot?

ewels:
Awesome. Thank you everyone! Merging..

PR #288, opened 2020-02-14
More metrics

MaxUlysse:
Why not getting metrics for all articles?


ewels:
Nice! I think that this can be done without repeating the code each time though, just using a regex to dump in the DOI as that's all that changes.

I had a stab at doing this, and also tidied them up a bit: https://github.com/nf-core/nf-co.re/pull/289

MaxUlysse:
I was sure it was possible, but my php is quite rusty ;-)

PR #293, opened 2020-02-18
Full-text readcube link for NBT paper

ewels:
Added full-text readcube link for the Nat Biotechnol. manscript on the publications page﻿


drpatelh:
Think it's also worth adding the paper to the homepage somewhere. It will be the primary landing site for most people.

ewels:
Will do so in a separate PR

PR #295, opened 2020-02-18
Added link to NBT paper on homepage

ewels:
image


drpatelh:
Nice! I was thinking it would be nice to make it more apparent but I can see why it fits in the way you have added it in :+1:

PR #298, opened 2020-02-19
Add BovReg to community page

ewels:
Added BovReg to the community page:

image

---

NB: This is using a JPG banner image taken from twitter. Would be better if we could get a SVG version of the logo if possible. Hopefully @JoseEspinosa can help 😁


ewels:
Brilliant - thanks @JoseEspinosa!

PR #300, opened 2020-02-19
Remove SciLifeLab and The Crick from community projects

ewels:
The community page has a section called "Projects we are involved with". We have SciLifeLab and The Crick there, which made sense early on when we didn't have much else. However, now we have quite a few actual projects there and I think we should probably remove SciLifeLab and The Crick.


drpatelh:
Yep. Fine with me :+1:

PR #310, opened 2020-02-25
BCM contributor: Replace logo with true SVGs

ewels:
See nf-core/nf-co.re#301


MaxUlysse:
How can people do that???

ewels:
It's an easy mistake :)

PR #311, opened 2020-02-25
Page titles

ewels:
If we have a page title, add this to the HTML `<title>` element, so that it is used for the browser tab and bookmarks.

Before:

image

After:

image


MaxUlysse:
I think you have commits from #310 at the same time

ewels:
yeah, sorry - I was too fast with opening this PR..

Also just added in the description meta tag..

PR #312, opened 2020-02-25
Sanitise meta title and description

ewels:
Oops. Fixes the left tab to look like the right tab:
 
image

I also removed the title link from pipeline pages. I'm not sure why that was there, it's inconsistent with the rest of the site and doesn't really add anything.

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -119,12 +119,12 @@ if( isset($markdown_fn) and $markdown_fn){
 // Page title
 $page_title = 'nf-core';
 if(isset($title) && strlen($title) > 0){
-  $page_title = $title.' &raquo; nf-core';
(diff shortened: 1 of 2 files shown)


drpatelh:
Possible to have `nf-core >> sarek` instead?

ewels:
> Possible to have `nf-core » sarek` instead?

Possible to have that, yes. But it's considered bad practice in web design circles. Imagine you have a bunch of different nf-core tabs open in a browser window and they're all squished: if it says `nf-core` first, then all you can see is `nf-core` repeated over and over. If it comes last, then you have useful information in the limited characters that you can see..

drpatelh:
This is why you are developing the website and I am not!

ewels:
high five

drpatelh:
high five

PR #315, opened 2020-02-27
fix spellings of Docker Hub

mashehu:


markdown/developers/adding_pipelines.md
```diff
@@ -74,7 +74,7 @@ git push
 
 ### Set up Travis and Docker Hub
 
-The nf-core pipelines use two additional services that link to GitHub: Travis and Dockerhub.
+The nf-core pipelines use two additional services that link to GitHub: Travis and Docker Hub.
```
ewels: Hmm, Travis eh?? I guess we should probably update these docs!

Is that something you fancy doing whilst you're here? Strip out all mention of Travis and mention GitHub Actions instead (which I think just work? Or does it need activating?)

markdown/developers/adding_pipelines.md
```diff
@@ -87,7 +87,7 @@ The repository should already contain the required `.travis.yml` file, so the ne
 time you push a commit the tests will be automatically triggered.
 
 The process for Docker Hub is similar, though a little more fiddly. Docker has a number
-of  different docker websites (docker hub, docker cloud, docker swarm _etc._), but
+of  different docker websites (Docker Hub, Docker Cloud, Docker Swarm _etc._), but
```
ewels: Is this still valid? Does Docker Cloud / Swarm still exist?

apeltzer: They still exist, but the current docs say we should always use Docker Hub. (so could drop all the others)

markdown/developers/adding_pipelines.md
mashehu: I tried to rewrite the documentation to fit GitHub Actions and remove all mentions of Travis.

markdown/developers/adding_pipelines.md
mashehu: Yes, it also felt a bit outdated to me, so I removed the whole paragraph now.

ewels:
Nice, thanks @mashehu!

PR #333, opened 2020-03-18
add icon picker in json schema builder

mashehu:


drpatelh:
Wooohoooooo @mashehu you are now officially the designated reviewer for anything that involves Javascript! 

https://github.com/nf-core/nf-co.re/pull/329#issuecomment-600025042

At least Ill be able to spend some quality time with my kids at Xmas 😅

mashehu:
I want to keep the unminified CSS and JS for now, while developing (if we need to make changes in the source).

For the `if` clauses in the header, is it okay if we have one for css files and one for js, so `nf-cor.css` will be loaded last (and I can make dark mode fixes to iconpicker.css)?

ewels:
Good point about the nf-core CSS. Maybe instead of adding more `if` blocks, we can just move the nf-core CSS to the end of the head section..?

mashehu:
The latest moves the icon selection from the modal to the row itself with a dropdown:
Screenshot 2020-03-18 18 16 06
I didn't touch the css yet, so the alignment is off 🤷‍♂️

ewels:
Nice, looking good! I guess this is now saving the class names (`fa fa-icon`) instead of the full icon string (`<i class="fa fa-icon"></i>`)? Or not? (I guess it's fine either way as long as we're consistent from now-ish onwards).

Does it work when you add new groups and rows? (I can't see where you initialise new even listeners on the dynamically generated content). Should be a minor fix if missing.

It would be nice if we could give it the same appearance as the other buttons - as narrow as possible, without the dropdown caret, and with the same width / height. Should be able to do that by overriding the CSS I guess.

I had thought about removing the icon from the settings modal in this way too, I think that's nicer. We could even look in to doing the same for the help text, having that in a popover window as well (as long as we can make one big enough), or maybe a separate modal.

I'm a little concerned that you have a bunch of `undefined` strings - that looks like a javascript bug to me. How did you generate this schema?

mashehu:
I think it is more human readable if it just `fa fa-icon`, but I guess for people who are not familiar with FA, the `fa` part can be very confusing...

Since it is not really a button, but a dropdown, I would advocate to keep the caret, which should also enhance the discoverability (that you can edit the icon by clicking on it), without losing too much screen estate (which we have enough of). 

I agree to move as much as possible away from the modal into the form itself. For the help text, I would expand the text field together with the row while editing and then collapse it to only show the first few words (indicating the existence of a text with an icon is not so useful in my opinion, because you can preview if there is a text or not by just showing the first few characters. This would also help to quickly check if the help text is assigned correctly). 

I created this schema by running `$ nf-core schema build .` inside my nf-core/cageseq repo...

ewels:
If you could resolve the conflicts in `header.php` and clear out everything that's not `header.php` and `update_fontawesome_icons.py`, that would be fab 👍

ewels:
My changes to your script: https://github.com/nf-core/nf-co.re/pull/335

PR #334, opened 2020-03-19
Built font-awesome icon picker

ewels:
My version of #333, inspired by @mashehu 

Differences to #333:

* Homebrew - no additional dependencies
* Fully functional (updates schema properly)
* Support for keyboard shortcuts / behaviour you might expect (click to hide, escape to hide etc)
* Focus support - auto focus of search field, refocus of button when popover hidden

I know you like the caret @mashehu, but I'm mostly working on my small laptop screen where space is at a massive premium and I preferred it without.

Additional changes:

* Form is now full-width fluid, instead of constraining to the page container
* Stripped font-awesome icon from modal
* JSON Schema now always shown at the bottom, instead of having to click _Finished_ to see it.

Note that I used your python script to generate the JSON file @mashehu, with a couple of changes. Firstly to give a custom JSON file as I'm not using iconpicker, and also simplifying the output a bit - I managed to reduce the file size by about half by making it a simple key:value map instead of more complex arrays etc.

Demo using only keyboard-shortcuts:

icon_picker


ewels:
@mashehu - if we're happy with this implementation, I suggest merging it and then refactoring your PR. I didn't want to commit your script and take credit for it 😅 If we keep your bits like moving the imports around, the script etc. Then once that's merged I'll make a PR against your script with the changes that I made. Sound ok?

public_html/assets/css/nf-core.css
```diff
@@ -906,6 +923,25 @@ Based on https://codepen.io/wintr/pen/beBJBb */
   padding: 1rem 1rem 0.5rem;
 }
 
+.fa_icon_picker.popover {
+  width: 315px;
```
mashehu: why in `px`? don't the icons scale with the rest of the fonts, so use `rem` instead?

public_html/assets/js/fa-icons.json
```diff
@@ -0,0 +1 @@
...nner", "fas fa-ship": "boat sea water", "fas fa-shipping-fast": "express fedex mail overnight package ups", "fab fa-shirtsinbulk": "", "fas fa-shoe-prints": "feet footprints steps walk", "fab fa-shopify": "", "fas fa-shopping-bag": "buy checkout grocery payment purchase", "fas fa-shopping-basket": "buy checkout grocery payment purchase", "fas fa-shopping-cart": "buy checkout grocery payment purchase", "fab fa-shopware": "", "fas fa-shower": "bath clean faucet water", "fas fa-shuttle-van": "airport machine public-transportation transportation travel vehicle", "fas fa-sign": "directions real estate signage wayfinding", "fas fa-sign-in-alt": "arrow enter join log in login sign in sign up sign-in signin signup", "fas fa-sign-language": "Translate asl deaf hands", "fas fa-sign-out-alt": "arrow exit leave log out logout sign-out", "fas fa-signal": "bars graph online reception status", "fas fa-signature": "John Hancock cursive name writing", "fas fa-sim-card": "hard drive hardware portable storage technology tiny", "fab fa-simplybuilt": "", "fab fa-sistrix": "", "fas fa-sitemap": "directory hierarchy ia information architecture organization", "fab fa-sith": "", "fas fa-skating": "activity
```
mashehu: for entries without a search term (e.g. 500px), the value should be the second part of the key without the `fa-`. Otherwise they will not come up in the search, I think.

mashehu: Actually, we should add that, to all entries, otherwise you will not find exact matches on the name,. e.g. "x-ray" (if I understand the search function correctly)

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -469,20 +576,6 @@ $(function () {
                 e.stopPropagation();
             }
         }
-        // Check that the font-awesome icon looks right
```
mashehu: will the schema validation do that, when somebody uploads their schema with a malformed FA icon?

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -695,22 +788,23 @@ function generate_param_row(id, param){
...
-        <div class="col-sm-3">
+        <div class="d-sm-none w-100"></div>
+        <div class="col">
             <label>Description
                 <input type="text" class="param_key param_description" data-param_key="description" value="`+param['description']+`">
             </label>
         </div>
-        <div class="col-sm-auto align-self-center schema_row_help_text_icon">`+help_text_icon+`</div>
-        <div class="col-sm-1">
+        <div class="col-auto align-self-center schema_row_help_text_icon iconpicker-component">`+help_text_icon+`</div>
```
mashehu: Why is this class needed here?

ewels:
Ok, I added a new modal for the help text. So the settings modal is trimmed down a lot more now which feels better.

Help modal follows the same kind of UI as the icon picker in terms of keyboard shortcuts and focus niceties. I also added GitHub-style _Write_ and _Preview_ tabs so that it's easy to see how it will be rendered:

helptext_modal

public_html/assets/js/nf-core-schema-builder.js
ewels: Yeah, it wasn't! I noticed it as well, now removed. 👍

public_html/assets/js/nf-core-schema-builder.js
ewels: Currently there is no validation so if someone uploads something weird then it will, yeah, probably break.

public_html/assets/js/fa-icons.json
ewels: Not needed, as I made the search box look both in the search terms but also the class names: `.popover-body div button[data-classname*='"+search_term+"']`

https://github.com/nf-core/nf-co.re/pull/334/files#diff-50fe998361a51c27bb42f5df30467eeaR90

public_html/assets/css/nf-core.css
ewels: Yeah, probably should do. Got to the value by tweaking in the browser so was being lazy. I'll fix 👍

ewels: Ok, switched to `rem` 👍

public_html/assets/js/nf-core-schema-builder.js
ewels: Ok, I added a validation step when building the schema rows. Now if there's something weird it will just delete the key from the schema and treat it as if there was no icon. It throws an error message to the browser console.

public_html/assets/css/nf-core.css
```diff
@@ -906,6 +920,29 @@ Based on https://codepen.io/wintr/pen/beBJBb */
...
+  -webkit-box-shadow: none !important;
+  -moz-box-shadow: none !important;
+  box-shadow: none !important;
+  font-size: 0.8rem;
+}
+.fa_icon_picker .popover-header {
+  padding: 0;
+}
+.fa_icon_picker .popover-body {
+  height: 200px;
```
mashehu: found another one 🙂 (I can't test these changes locally and I am bad with guessing, so you have to fix that on your side)

public_html/json_schema_build.php
```diff
@@ -249,15 +256,7 @@ function validate_cache_id($cache_id){
...
-                    <div class="form-group settings_help_text_group">
-                        <label for="settings_enum">Help text</label>
-                        <textarea class="form-control" id="settings_help_text" rows="5"></textarea>
-                        <small class="form-text text-muted">
-                            Longer help text explaining how to use the parameter, typically shown on user request.
-                            Some tools may render <a href="https://www.markdownguide.org/cheat-sheet/" target="_blank">markdown</a>,
-                            some may render as plain text.
-                        </small>
-                    </div>
+                    <p class="text-muted settings_nothing_special">No special settings available.</p>
```
mashehu: cool classname 😎 😉

public_html/json_schema_build.php
```diff
@@ -322,10 +313,67 @@ function validate_cache_id($cache_id){
...
+                                        but remember that this will be shown raw on the command line. So no tables please!
+                                    </small>
+                                </div>
+                            </div>
+                            <div class="tab-pane fade" id="tab-helptext-preview" role="tabpanel">
+                                <p>Command-line:</p>
+                                <pre><span class="helptext-preview-title"></span>
+<span class="helptext-preview-description"></span>
+
+<span class="helptext-preview-helptext text-muted"></span></pre>
```
mashehu: how about if we call this class `helptext-preview-text`, sound less confusing to me 🙂

ewels: Yeah, could do. It makes more sense in the javascript though as the variable we put in here is called `helptext`

PR #335, opened 2020-03-19
Update update_fontawesome_icons.py

ewels:
Refactor to give new JSON output for the JSON Schema icon picker.

Also refactored a bit to simplify the code and make it use an explicit version of Font Awesome.


ewels:
hah, ok the diff is a little extreme 😳 But it's still basically your code 😅

PR #338, opened 2020-03-20
Convert default string to native types

ewels:
This should hopefully make the schema builder use native types for booleans, numbers / ranges (floats) and integers. For example, `"default": true` instead of `"default": "True"`.

Also handle invalid JSON schema in a better way.

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -135,8 +135,22 @@ $(function () {
     }
 
     // Build the schema builder
-    $('#schema-builder').html( generate_obj(schema['properties'], 1) );
-    init_group_sortable();
+    try {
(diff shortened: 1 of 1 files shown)


public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -324,6 +340,30 @@ $(function () {
...
+                            param['default'] = true;
+                        } else {
+                            param['default'] = false;
+                        }
+                    }
+                }
+
+                // Convert number types
+                if(['number', 'int', 'range'].indexOf(param['type']) != -1){
+                    if(param.hasOwnProperty('default') && typeof param['default'] == "string" && param['default'].trim() != ''){
```
mashehu: why do you check for `ypeof param['default'] == "string"`?

PR #339, opened 2020-03-21
🖤 fix/add dark mode for schema builder and smaller general fixes 🖤 

mashehu:
CSS changes:
- fix dark mode for schema builder
- increase contrast for labels in light mode
- make copy button narrower
- fit popover width to content width

Other changes:
- make width for default value smaller for numbers
- add border on group header
- change color of ui-selectable placeholder (the yellow was too 90s for me 😉)
- add margin to submit button, so that it doesn't touch the footer

Screenshot 2020-03-21 13 52 48
Screenshot 2020-03-21 13 51 52
Screenshot 2020-03-21 13 52 19
Screenshot 2020-03-21 13 53 19

### DIFF
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -921,8 +921,9 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 }
 
 .fa_icon_picker.popover {
-  width: 15.3rem;
-}
+  width: 17rem;
+}   
+
 .fa_icon_picker .popover-header input {
   outline: none;
   -webkit-box-shadow: none !important;
@@ -945,9 +946,6 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 
 #json_schema {
   font-size: 0.7rem;
-}
-.schema_row input[type=number]{
-  width: 4rem;
 }
 .copy-schema-btn{
   width: 15rem;
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -847,7 +847,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
   background-color: #ededed;
 }
 .schema_row label {
-  color: #ddd;
+  color: #bbb;
   margin: 0.2rem 0 0;
   font-size: 0.7rem;
   width: 100%;
@@ -921,7 +921,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 }
 
 .fa_icon_picker.popover {
-  width: 17rem;
+  width: 15.3rem;
 }
 .fa_icon_picker .popover-header input {
   outline: none;
@@ -945,4 +945,10 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 
 #json_schema {
   font-size: 0.7rem;
+}
+.schema_row input[type=number]{
+  width: 4rem;
+}
+.copy-schema-btn{
+  width: 15rem;
 }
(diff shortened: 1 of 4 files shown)


public_html/assets/css/nf-core.css
```diff
@@ -921,7 +921,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 }
 
 .fa_icon_picker.popover {
-  width: 17rem;
+  width: 15.3rem;
```
ewels: How come this change was needed? For me the size already fits the contents?

image

public_html/assets/css/nf-core.css
```diff
@@ -946,3 +946,9 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 #json_schema {
   font-size: 0.7rem;
 }
+.schema_row input[type=number]{
+  width: 4rem;
+}
```
ewels: Why do we want narrower numeric input fields?

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -962,7 +962,7 @@ function generate_group_row(id, param, child_params){
     <div class="card schema_group" data-id="`+id+`">
         <div class="card-header p-0">
             <div class="row schema_row schema_group_row mb-0" data-id="`+id+`">
-                <div class="col-auto align-self-center schema_row_grabber">
+                <div class="col-auto align-self-center schema_row_grabber border-right">
```
ewels: Good spot! 🏆

public_html/assets/css/nf-core.css
mashehu: It looks weird to me if the up-down buttons are so far away:
Screenshot 2020-03-23 12 09 12

public_html/assets/css/nf-core.css
mashehu: My Chrome:
Screenshot 2020-03-23 12 09 00
Could it maybe be because I don't hide my scrollbars?

ewels: Ah, yes - that'll be it 👍 

Maybe we could force the scrollbar to always show? Would be a good idea anyway to show that there are lots of icons to choose from and not just the ones initially visible..

https://codepen.io/stevenlewis/pen/hubpL

public_html/assets/css/nf-core.css
ewels: ...I guess. Looks weird to me if the inputs aren't all the same width though:

image

Personally I prefer it the way it was - I don't really like having a chunk of space that is wasted, even if the buttons are kind of distant. I'm not sure that people will use the buttons much anyway, I never do.

public_html/assets/css/nf-core.css
mashehu: I tried this and several other versions and at least for me it doesn't work after changing the general settings to show scroll bars only while scrolling. I guess it is best to keep you width then, which works for standard settings.

public_html/assets/css/nf-core.css
mashehu: how about putting it on the left side:
Screenshot 2020-03-23 13 48 57

public_html/assets/css/nf-core.css
ewels: ok!

public_html/assets/css/nf-core.css
ewels: I still prefer it as it was to be honest, but if you have a strong attachment to this then that's fine 😄

mashehu: Not a hill I have to die on.

PR #341, opened 2020-03-25
BIGGER icons in icon picker

mashehu:
Screenshot 2020-03-25 17 19 48


ewels:
Wow those are huge 😳 😆 

Can we maybe make the picker window a little bigger and tone down the black colour in the light styles? Both minor requests, so feel free to ignore.

PR #343, opened 2020-03-25
fix dark mode for Github graphs on stats page

mashehu:
Before:
Screenshot 2020-03-25 18 15 15
After (with fake data):
Screenshot 2020-03-25 18 14 42

+ remove blue color from twitter icon (to make it more similar to the rest)


ewels:
Nice! I think the fake data is in the PR though..

mashehu:
> Nice! I think the fake data is in the PR though..

Oops, fixed now.

PR #345, opened 2020-03-31
update font awesome 

mashehu:
Lessons learned:
I can't update the FA script in the header, because it is account based. I guess you, @ewels, have access to it. If I understand it correctly, you can just update it on the FA account page, without us needing to update the link.


ewels:
Thanks for looking into this @mashehu - I guess you're right. I remember being annoyed that I couldn't just use a bunch of static files like I had in the past.

And... turns out that I have it set to "Latest":

image

So after me whinging about FA versions for the icon picker it turns out that we're not even fixed 🤦‍♂ 😆 

Do you think I should pin 5.13? I guess it's not a problem for the imported version to be more up to date than the JSON, only the other way around..?

Means that it could also not be such a bad idea to regenerate the JSON file on a cron job if we're sure that we're on the latest version anyway..

Phil

mashehu:
It depends what you prefer: Being hassled into changing the FA kit version everytime or setting up a cron job, which would include untracking `fa-icons.json`, so you don't get any git errors when trying to deploy changes after FA updates.

ewels:
..is this a trick question? 🤔

ewels:
Merging now, can do the automated thing in a separate PR. The python env on my shared web hosting is crappy and difficult to work with, so I think the script should be re-written in PHP. I can do that now quickly if you like it'd be a nice end-of-day mini project :)

mashehu:
My php is like your python env then 😉, so go ahead!

ewels:
https://github.com/nf-core/nf-co.re/pull/346

PR #346, opened 2020-03-31
Get FontAwesome icons JSON file with PHP instead of Python

ewels:
Rewrote the FontAwewsome script in PHP so that it can be easily run on the server.

Once merged I will create a cron job.


ewels:
ok, ran the script on the server to regenerate the file and created a cron job to run once a week 👍

ewels:
Tested and working:

image

PR #348, opened 2020-04-09
Pipeline releases page - expand first release

ewels:
On the pipeline releases page, each release shows a toggle expand with details and buttons. This isn't super obvious from just looking at it. To help, this change makes the top release expanded on page load. This is usually the one that people will be interested in anyway, and also hints at the functionality for the others.

### DIFF
--- a/includes/pipeline_page/releases.php
+++ b/includes/pipeline_page/releases.php
@@ -6,6 +6,7 @@ ob_start();
 
 echo '<h1>Version history</h1>';
 
+$first = true;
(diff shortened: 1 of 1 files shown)


ewels:
Now looks like this on page load:

image


Instead of this:

image

mashehu:
Should we maybe populate these with changes of each release, like GitHub does it?

ewels:
You mean show the changelog? I think I'm happy with the _View Release_ button, which takes you to that..

PR #351, opened 2020-04-20
Update troubleshooting.md

jfy133:
Add description as to what 'sticked on revision' means, as the nf-core documentation is generally more accessible by users (than the nextflow docs which is more developer focused), even if it's a nextflow error.

Suggested for nf-core website here: https://github.com/nf-core/eager/pull/386#issuecomment-616419347

### DIFF
--- a/markdown/usage/troubleshooting.md
+++ b/markdown/usage/troubleshooting.md
@@ -7,6 +7,7 @@ subtitle: How to troubleshoot common mistakes and issues
 * [Data organization](#data-organization)
 * [Image cannot be built](#image-cannot-be-built)
 * [Cannot find input files when using Singularity](#cannot-find-input-files-when-using-singularity)
+* [Warning about 'sticked on revision'](#warning-about-sticked-on-revision)
 * [Extra resources and getting help](#extra-resources-and-getting-help)
 
 ## Input files not found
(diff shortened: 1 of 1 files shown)


markdown/usage/troubleshooting.md
```diff
@@ -78,6 +79,26 @@ bind path = /home
...
+
+For example:
+
+```bash
+nextflow run nf-core/eager -r 2.1.0 --reads '/<path>/<to>/data/*_{R1,R2}_*.fq.gz' <...>
+```
+
+Specifying the version of the run you are using is highly recommended, as it helps in full reproducibility. In the sense that if you explicitly record the whole command _with_ the version for your publication or internal reports, then anyone who wants to check your work can use the exact version you used (including all internal tools).
+
+You can see more information on the nextflow documentation [here](https://www.nextflow.io/docs/latest/sharing.html?highlight=revision#handling-revisions).
```
ewels, line 100:
```suggestion
You can see more information on the nextflow documentation [here](https://www.nextflow.io/docs/latest/sharing.html#handling-revisions).
```

jfy133:
> Looks good! Is this a new Nextflow feature? I don't remember seeing this message before.
> 
> "Sticked on revision" could maybe do with rewording a little...

Nope, been there ever since I've been using Nextflow (from 19.04 I think). I guess you're too much of a nextflow master to encounter these errors ;)

ewels:
That or I haven't really been running any analysis since about v0.32 😆

PR #353, opened 2020-04-24
Added Flomics to the list

jcurado-flomics:
The location is the same as the CRG, I hope that's not a problem.

Thanks!


jcurado-flomics:
Thanks! :)

PR #359, opened 2020-05-08
Update navigation

ewels:
Added Data Management: x-ref https://github.com/nf-core/nf-co.re/pull/349

Also updated the nav in the footer, that had drifted quite a long way out of date.


mashehu:
For my taste it looks imbalanced/too long/overwhelming now: 
Screenshot 2020-05-08 09 57 52

Also a bit confusing, that "getting started" is the third item in a list titled "getting started" 🙂
I think for example "running offline", "troubleshooting" and "helper tools" don't need to be available at that level

Another point: how actively is the google groups page used?

ewels:
Yes I know, I thought the same but figured I’d push ahead as I want to do a major rethink of the whole site structure anyway (there’s another issue open for that already).

We should kill the google groups really. Been thinking about that for a while.

ewels:
https://github.com/nf-core/nf-co.re/issues/352

PR #364, opened 2020-05-18
Rewrite how events are handled

ewels:
Now, each event has its own Markdown file, allowing for a lot more content and flexibility.


mashehu:
Just a matter of taste: but I think there should be a future event heading even if there are none scheduled yet (with a short placeholder text), to make it clearer where one would find these in the case there would be one and indicating that this is not just an archive page 🙂

ewels:
Sure, that sounds good to me too. Updated to always show the headings :+1:

PR #365, opened 2020-05-18
Add edit button to header for GitHub markdown URL

ewels:
Much of the nf-core website is built using Markdown files that are hosted in a range of different nf-core GitHub repositories.

The footer of such pages already shows a link to this markdown file. However, most people don't notice this and then struggle to know how to edit the page.

This PR adds a small button to the header of every page that has the markdown URL set to make this more obvious.

image

image

image

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -260,9 +260,14 @@ if(isset($title) and $title): ?>
     <div class="mainpage">
       <div class="mainpage-heading">
         <div class="container">
-          <?php if(isset($header_btn_url) && isset($header_btn_text)){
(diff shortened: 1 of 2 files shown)


mashehu:
Nice! To reduce redundancy should we then remove the the "Read this page on GitHub" footer?

ewels:
> To reduce redundancy should we then remove the the "Read this page on GitHub" footer?

I'd rather not - I quite like this as personally I like to check it to see where it's coming from sometimes, and I think that they are fair enough away from one another to not be a problem (like the navigation, which is also in the header and footer).

PR #369, opened 2020-05-19
Fix broken events page

ewels:
* Properly parse only the front-matter as YAML instead of the whole markdown file
* Add CSS classes to tables for nicer formatting

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1,6 +1,26 @@
 <?php
 // Common PHP functions for the website
 
+// Pull out YAML front-matter from a markdown file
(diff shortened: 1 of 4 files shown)


mashehu:
Can you maybe add one check to not show an empty "Location:" in the `header_html` if not set.

ewels:
Good spot 👍

ewels:
Ok, fixed - added a few additional tweaks to the markdown rendering too.

mashehu:
Interesting choice for location 🙂

ewels:
> Interesting choice for location 🙂

It was an aesthetic thing - I didn't like the empty space! 😅

PR #382, opened 2020-06-15
YAML lint

ewels:
New GitHub actions workflow to lint all YAML files when they are edited.

### DIFF
--- a/.github/workflows/markdown.yml
+++ b/.github/workflows/markdown.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   test_markdown:
     name: Test markdown syntax
-    runs-on: ubuntu-16.04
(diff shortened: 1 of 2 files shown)


.github/workflows/markdown.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   test_markdown:
     name: Test markdown syntax
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
apeltzer, line 11: good spot! wondered why we had any pinned ubuntu releases in Github actions at all.

ewels: Copy paste from the GHA docs I guess

PR #383, opened 2020-06-16
Added some FontAwesome to the homepage

ewels:
Messing around with the new FontAwesome Pro to add some nice new icons to the homepage:

image

---

image

### DIFF
--- a/public_html/index.php
+++ b/public_html/index.php
@@ -200,7 +200,7 @@ nf-core list
         <div class="col-md-4 px-md-5 mb-5 mb-md-0">
           <h3>See what's available</h3>
           <p>Check out the available pipelines to see if we have what you need. Each comes with release details, keywords and a description.</p>
-          <a class="btn btn-lg btn-success arrow-hover" href="/pipelines"><span><i class="fad fa-toolbox mr-1"></i> Available pipelines</span></a>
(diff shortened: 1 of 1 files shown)


mashehu:
I find the two-color toolbox icon a bit off, because in my opinion the accent colors should be switched with the main color. But that is something FA has to fix 🙂. So my proposal would be to just use the normal one color fa-toolbox, to improve readability of the symbol.

ewels:
[`fa-swap-opacity`](https://fontawesome.com/how-to-use/on-the-web/styling/duotone-icons#swapping-layers) to the rescue! 




Though actually you might be right that the solid version is still better:

ewels:
Any suggestions for a better icon?

mashehu:
Unfortunately not, the only one which came to my mind was the ball pile https://fontawesome.com/icons/ball-pile?style=regular, where the regular version looks a bit like a bunch of pipelines. but maybe I have too much fantasy there and a toolbox is actually the closest icon association.

ewels:
I like where you're going with the ball pile icon @mashehu, but after playing with a couple it feels a bit busy and not super clear.. Might work with a big icon though. I think the toolbox is ok for now..

Thanks both!

jfy133:
Ok I forgot to press comment: I think the full opacity toolbox is fine. I found the balls a bit too abstract.

ewels:
Glad we're in agreement 😄

PR #384, opened 2020-06-17
JSON schema builder - better help, icon tweaks

ewels:
Some tweaks to the JSON schema builder:

* Tidied up the help text at the top - hid more by default, added extra line with close text at the bottom of the expanded content:
Jun-17-2020 09-41-37

* Replaced the icons for _set icon_ and _add help_:
image

* Bumped up the icon size a little in the builder UI buttons

### DIFF
--- a/public_html/assets/css/nf-core-dark.css
+++ b/public_html/assets/css/nf-core-dark.css
@@ -371,8 +371,13 @@ hr {
   color: #ddd;
 }
 .param_fa_icon_missing, .help_text_icon.help_text_icon_no_text {
-  color: #999;
+  color: #616161;
 }
+.param_fa_icon:hover .param_fa_icon_missing,
+.schema_row_help_text_icon:hover .help_text_icon.help_text_icon_no_text {
+  color: #ddd;
+}
+
 .popover,
 .popover-header,
 .popover-body{
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -906,6 +906,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 }
 .schema_row > button {
   line-height: 3rem;
+  font-size: 1rem;
   vertical-align: middle;
   border: 0;
   background-color: transparent;
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -7,8 +7,8 @@
 var schema = '';
 var new_param_idx = 1;
 var new_group_idx = 1;
-var help_text_icon_template = '<i class="fas fa-comment-slash help_text_icon help_text_icon_no_text" data-toggle="tooltip" data-html="true" data-placement="right" data-delay="500" title="Does not have any help text"></i>';
-var no_help_text_icon = '<i class="fas fa-comment-dots help_text_icon" data-toggle="tooltip" data-html="true" data-placement="right" data-delay="500" title="Has help text"></i>';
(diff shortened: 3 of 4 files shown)


ewels:
Link to current page: https://nf-co.re/json_schema_build

jfy133:
Book for help is good. 
I personally find the icon one is also a bit random because of the contents (even if it's called 'icon'). We don't have music objects, for example ;). Maybe just a 'file-image' or 'image-polaroid' icon work?

mashehu:
sorry to piggyback on this PR, but I had while reviewing it, I found the darkmode icon's were note muted enough if missing.
before:
Screenshot 2020-06-17 10 38 45
after:
Screenshot 2020-06-17 10 38 24
hover:
Screenshot 2020-06-17 10 56 44

mashehu:
I agree that the icons icon is misleading for our use case. what about `question-square`? I think it is not so easily confused with the `question-circle` which often is used for help/tooltips.

Speaking of help: I am not so satisfied with the way the help text is handled (and vertically centered between the `<hr>`s 😉). How about replacing the ambiguous "click here to read more" with "quick help" with a triangle indicating collapsed text?

ewels:
Aww, I was quite happy about the icons icon.. 😞 

I'm not sure about using `question-*` icons. We have the circular one currently and I think it's too ambiguous - specifically too close to "help" as you say (I saw some people interpret the previous one as exactly that). I'm not sure that the square version is a lot better tbh..

> We don't have music objects, for example ;)

Not, _yet_..!

> How about replacing the ambiguous "click here to read more" with "quick help" with a triangle indicating collapsed text?

Yeah the same thing crossed my mind but I couldn't be bothered. Feel free to add to this PR / open a new one :) Could even move it into a modal or something.

ewels:
So.. any suggestions for better alternatives for the icons icon?

mashehu:
how about making it more imperative and using `plus-square` https://fontawesome.com/icons/plus-square?style=regular ?

mashehu:
oh, we are already using that for "add parameter"... 😞

ewels:
I wondered the same, but to me that means add row or something. I expect an action on the entire row.

What about `paint-brush`  or `palette`  or something?

jfy133:
Any reason why not the file-image/file-polaroid? Or are you using that elsewhere already?

ewels:
I worry that people might use file icons a lot in their schema is one reason. Also again, to me it doesn't say "add icon".

ewels:
Maybe `image-polaroid` is ok actually..

mashehu:
`paint-brush` and `palette` to me say "choose brush/ polish item" and "choose color". If I see the image/polaroid icons I would expect an image upload dialogue, when clicking on it. maybe the icons-icon is actually the best solution...

ewels:
There's also `icons-alt`, which I only just noticed:

ewels:
Also, if you want more work to do @mashehu - the icon picker only shows the free icons 😉 I think that the GitHub repo we use to get the metadata only lists the free icons, so I'm not sure if we can do anything about this currently.

I'm also not sure if we _should_ do anything about it, sticking to only free icons may be better in terms of accessibility / use on other platforms / if we lost the pro subscription in the future.

mashehu:
The problem with icons-alt for me is that it is showing symbols not icons, like the normal version does, so I prefer the non-alt version.  

Concerning an icon picker with pro icons, I remember somebody linking to a list of pro icons, but I think we should stick to regular ones, so the schema can be better reused and spread 🙂

jfy133:
Yeah, I agree with @mashehu if there isn't any better alternative the first `icon` is slightly clearer.

ewels:
Ok, I'm merging before anyone can change their mind 😆

ewels:
Thanks both! If nothing else I hope that we can agree that this is an improvement 😄

PR #386, opened 2020-06-17
Remove old parameter-schema files

ewels:
With the new JSON schema formats being used by nf-core/tools, we no longer need a nf-core specific schema-schema. It's time for these files to be removed.

Hopefully this won't break anything for anyone...

workflow


ewels:
💥

PR #388, opened 2020-06-22
JSON schema builder: Add form to load previous build ID.

ewels:
See https://github.com/nf-core/tools/issues/631

When loading the schema builder URL without an ID specified, an extra form is shown where you can enter a previous ID to continue working on:

image

This is to tie in with better logging from `nf-core tools` to expose that ID clearly if the editor is exited manually or times out etc.

### DIFF
--- a/public_html/json_schema_build.php
+++ b/public_html/json_schema_build.php
@@ -198,6 +198,18 @@ This page helps pipeline authors to build their pipeline schema file by using a 
 
 <?php if(!$schema_cache){ ?>
 
+<h3>Load Schema</h3>
+
+<p>If you previously ran <code>nf-core schema build</code> and forgot to save, you can resume editing by entering the build ID below:</p>
+<form method="get" action="" class="form-inline mb-2">
+    <input type="text" class="form-control mr-2" name="id" placeholder="Build ID">
+    <button type="submit" class="btn btn-primary">Load</button>
+</form>
+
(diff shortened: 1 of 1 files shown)


ewels:
Also fixed the help text: _Use `Enter` and `Tab` + `Enter` to go up and down._ (should be `Shift` + `Enter`).

Thanks @MaxUlysse!

public_html/json_schema_build.php
```diff
@@ -179,7 +179,7 @@ function validate_cache_id($cache_id){
         <li>The <i class="fas fa-book help_text_icon"></i> icon shows whether help text has been written. To add, click on it.</li>
         <li>Be a power user with keyboard shortcuts!
             <ul class="small">
-                <li>Use <code class="border shadow-sm">Enter</code> and <code class="border shadow-sm">Tab</code>+<code class="border shadow-sm">Enter</code> to go up and down.</li>
+                <li>Use <code class="border shadow-sm">Enter</code> and <code class="border shadow-sm">Shift</code>+<code class="border shadow-sm">Enter</code> to go up and down.</li>
```
MaxUlysse, line 182: Also, you say `Enter` and `Shift` + `Enter`, to go up and down, but it's actually to go down and up.
 But I agree it sounds better that way.

ewels: Hah, good point! But yeah, that sounds kind of odd..

PR #395, opened 2020-06-24
Schema builder - group min height

ewels:
Set a minimum height for the content area of a group. This makes it easier to drop params into empty groups.

Very minor change, but makes it quite a bit easier to drop into new groups in my testing.

Before:

image

After:

image

Suggested by @ggabernet in nf-core/tools#626

### DIFF
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -929,6 +929,12 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 .schema_group .card-body {
   padding: 1rem 1rem 0.5rem;
 }
+.group-drag-drop-help {
+    font-size: 80%;
+    color: #6c757d;
+    font-style: italic;
+    margin-bottom: 0;
+}
 .fa_icon_picker.popover {
   max-width: 20rem;
 }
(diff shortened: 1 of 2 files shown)


public_html/assets/css/nf-core.css
```diff
@@ -928,6 +928,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
 }
 .schema_group .card-body {
   padding: 1rem 1rem 0.5rem;
+  min-height: 50px;
```
mashehu, line 931:
```suggestion
  min-height: 1 rem;
```

ewels:
```suggestion
  min-height: 3rem;
```

mashehu:
I am still not 100% happy how it looks like. How about some muted help text in there a la "drag and drop a parameter here"?

ewels:
Ok, I got it to work with the helper text instead. I think it's nicer 👍 

Jun-24-2020 14-41-14

ggabernet:
Even nicer like this!

ewels:
@mashehu - happy with this now?

mashehu:
yes! I would prefer if the text would disappear already while hovering with a grabbed parameter over the group, but one can't have everything 🙂

ewels:
Yeah me too, but that was going to be more difficult 😅 Also having the text there actually makes it easier to drop..

PR #396, opened 2020-06-24
Schema builder improvements

ewels:
I have a nasty feeling that I am being hypocritical and now implementing some advice that I previously rejected. But here goes anyway.

Changes made after doing some editing work myself. Process is quite a bit smoother with these changes in place.

* Moved group description back to be consistent order with params
* Display group type for consistency on big screens (not editable)
* Added Hide checkbox to groups, toggles all child params
* Renamed `Hidden` to `Hide` on the checkbox labels

Example before:

image

Example after:

Jun-24-2020 10-52-19

### DIFF
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -927,9 +927,8 @@ Based on https://codepen.io/wintr/pen/beBJBb */
   margin: 1rem 0;
 }
 .schema_group .card-body {
-  padding: 1rem 1rem 0.5rem;
+    padding: 1rem 2.3rem 0.5rem;
 }
-
 .fa_icon_picker.popover {
   max-width: 20rem;
 }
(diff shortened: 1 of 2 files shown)


mashehu:
- When I add a new group it is hidden by default, is that on purpose?
- When I drop a not hidden paramter in a hidden group, should it's status not also change to hidden
- How about adding more padding on the right for parameters in a group, e.g. 
```{css}
.schema_group .card-body {
    padding: 1rem 2.3rem 0.5rem;
}
```
so that the `Hide` check boxes are aligned:
Screenshot 2020-06-24 11 04 39
- How about centering the labels for checkboxes:
Screenshot 2020-06-24 11 26 38

I will push a commit with this changes to this PR.

ewels:
> When I add a new group it is hidden by default, is that on purpose?

No - it's hidden until there are some children which are visible. I'll add a check so that it's unhidden if there are zero children.

> When I drop a not hidden paramter in a hidden group, should it's status not also change to hidden

Good spot, I need to add checks to the drag + drop behaviour.

> How about adding more padding on the right for parameters in a group

I like it! But the extra padding on the left looks a bit odd to me. I'll revert that but keep the right alignment and centred text 👍

ewels:
Ok, updated - let me know what you think.

Issue #390, opened 2020-06-22
Schema builder: icon picker broken for groups


ewels:
The UI doesn't set the icon properly for groups, works fine for params.

ewels:
Turns out that this wasn't group specific, but rather any element that was added to the page after page load (which is mostly groups usually).

PR #397, opened 2020-06-24
Schema builder tweaks

ewels:
* New toolbar buttons to collapse and expand groups
* Restyle the 'back to top' button to make it obviously a button
* Add a new button on the Finished page to explicity go back to editing
* Properly disable and re-enable all buttons when Finished
* Box-shadows for groups to give them some visual separation
* Darker box-shadow for the toolbar plus more spacing
* Warning alert when returning to the editor that the console will have exited

### DIFF
--- a/public_html/assets/css/nf-core-dark.css
+++ b/public_html/assets/css/nf-core-dark.css
@@ -79,6 +79,10 @@ blockquote {
   border-color: #36646e;
   color: #ade4f1;
 }
+.alert-secondary{
+  background-color: #616161;
+  border-color: #616161;
+}
 .badge-light {
   color: rgb(218, 215, 210);
   background-color: rgb(26, 28, 29);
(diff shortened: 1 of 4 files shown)


ewels:
Now adding a new "Preview docs" feature, will append in a moment..

ewels:
New "Preview Docs" feature added:

Jun-24-2020 14-43-31

mashehu:
- I would prefer an even stronger visually differentiator between groups and parameters. When collapsed they still look a bit too similar for my taste:
Screenshot 2020-06-24 15 33 42
- Should there maybe be a visual differentiator between the settings and the collapse button?
- The tool bar is overcrowded/confusing for me. Should we make the two add buttons and the two collapse / expand buttons btn-groups?

ewels:
Button groups is a nice idea - looks better, pushed.

I have played with the visual differentiation quite a bit now. I haven't found anything yet that works without making the page too busy. In practice I don't think it's a big problem though - basically all schema should have all parameters within groups, so the scenario of having top-level params mixed with groups is likely to be rare.

mashehu:
btw. the iconpicker is currently broken for me. No, error message, so I guess the event is never recognized?

ewels:
Weird - icon picker is working fine for me? I did change some minor behaviour with it the other day - see #390, but that was fixing it not breaking it (hopefully!).

mashehu:
sorry, false alarm. the icon picker just doesn't work on my local version. works fine on the website.

PR #401, opened 2020-06-27
Improve iiitb logo

MaxUlysse:
Current logo that we have for `International Institute of Information Technology - Bangalore` seems to be just a part of the actual logo that was manually vectorized.
Here is a full vector version.


MaxUlysse:
current color
new color

current white
new white

PR #400, opened 2020-06-26
nf-core Launch GUI

ewels:
Web implementation of the graphical user interface for launch Nextflow workflows using `nf-core launch`.

Functionality for running this from a https://nf-co.re pipeline page is planned and basically built in and ready to go, but I need to go to bed so that will come in a future PR or be tacked on to this one later next week.

See Slack for screen capture demos: https://nfcore.slack.com/archives/CUC8PPDL2/p1593212486143900

PR #402, opened 2020-06-29
Launch improvements

ewels:
* Renamed URL `/json_schema_launch` to just `/launch`
* Added functionality to pick a pipeline and release, if loaded with no inputs
* Added functionality to load a schema and launcher through GitHub API

Ties in with updated https://github.com/nf-core/tools/pull/638


ewels:
Happy to merge this again, like #400 so that it can be tested on the live site.

Only changes to the new launch page, so shouldn't break anything. I think.

PR #403, opened 2020-06-29
better logos

MaxUlysse:


mashehu:
Nice! The project logos should also be white in dark mode, I think:
Screenshot 2020-06-29 16 34 40

MaxUlysse:
> Nice! The project logos should also be white in dark mode, I think:
> Screenshot 2020-06-29 16 34 40

That's a good point, we should have white version of them at some point.

But my PR was mainly about having actual SVG logos instead of a vectorization ;-)

PR #405, opened 2020-06-30
Launch - cache, CTA button, header info

ewels:
A few minor improvements around JSON schema launch:

* Pipeline page: check if a schema exists - if it does, show a _Launch_ button. If not, show the previous _Get release_ button.
* Launch: Cache schema downloads
* Launch: If we have a pipeline name and revision, print it in the header. More prominent cache ID in header.

### DIFF
--- a/public_html/launch.php
+++ b/public_html/launch.php
@@ -60,15 +60,37 @@ function launch_pipeline_web($pipeline, $release){
         return ["Error - Pipeline name <code>$pipeline</code> not recognised"];
     }
     // Try to fetch the nextflow_schema.json file
-    $api_opts = stream_context_create([ 'http' => [ 'method' => 'GET', 'header' => [ 'User-Agent: PHP' ] ] ]);
-    $gh_launch_schema_url = "https://api.github.com/repos/nf-core/{$pipeline}/contents/nextflow_schema.json?ref={$release}";
-    $gh_launch_schema_json = file_get_contents($gh_launch_schema_url, false, $api_opts);
-    if(!in_array("HTTP/1.1 200 OK", $http_response_header)){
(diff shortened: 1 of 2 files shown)


ewels:
image

With launch button (as on current live site):
image


If no schema available:

image

mashehu:
Maybe add a `<i class="fad fa-download"></i>` in front of the the 'Get version ' button?

ewels:
haha, I _knew_ you'd say that! 😆

I just copied what was already there previously - as I did a quick test I thought _"@mashehu is totally going to call me up on the lack of an icon here_"

ewels:
Not sure about the duotone look:


Inverted is maybe better, but still not great?


I think solid looks best:


Sound good? I'll put in a PR now.

ewels:
I considered the Tag icons too, but they didn't look right either.

mashehu:
yes, I prefer also the solid version. And a download icon on a link to a github release page is also a bit misleading, especially that's usually not how you would get the pipeline to use it. How about rephrasing it to "See release notes"?

ewels:
Hmm, not super catchy.. 😕 

How about this?

image

Still not as nice a CTA as the _"Get version"_ text.. But hopefully these will all disappear soon in favour of launch buttons anyway..

ewels:
Struggling to find a good one for _"See development code"_.. Maybe `sparkles`??

image

Any suggestions?

mashehu:
How about an homage to the old internet (which was admittedly before my time): `<i class="fad fa-construction"></i>` or go for `hammer` or `wrench`. Also "development code" sounds to me like the code used to develop the pipeline and not like code currently under development.

PR #415, opened 2020-07-02
Join: Remove Google Groups. Spruce up a bit.

ewels:
Removed Google Groups from the website completely, ahead of archiving that forum.

Spruced up the _Join_ page a little whilst I was there:

* Clearer alert box about the main Nextflow Gitter channel
* New consistent call-to-action button to join each channel section (used twitter and YouTube embeds)
* Minor rewording, mention of Slack group for GitHub invitations.

localhost_8888_join


ewels:
I just deleted the google group, so let's merge this ASAP if that's ok :)

mashehu:
On thing with respect to wording: "Follow this link to get an invite to join the nf-core Slack organisation." when refering to a button seems confusing to me. 
Pedantic side note: the twitter logo should be the bird without a box around it, which in my opinion would also fit better in with the other logos.

ewels:
Yup, I agree with both points. The Slack text was the end result of moving around and fiddling repeatedly, which is why it probably sounds a bit odd. Any suggestions for better alternatives? Better still, want to make a PR? 😉

PR #416, opened 2020-07-02
change wording on join page and switch twitter logo

mashehu:
As requested in #415 🙂

The new slack button text is a bit wordy. I am also not happy that the same button has two different texts on the page, but I don't like a wordy button in the button row. Open for other ideas.
Screenshot 2020-07-02 15 37 35

New twitter logo:
Screenshot 2020-07-02 15 17 10
Screenshot 2020-07-02 15 17 03


ewels:
Thanks!

Issue #422, opened 2020-07-06
Multi-selection for schema builder


apeltzer:
It would be speeding things up quite a lot to have some sort of multi-selection available, either via ticking certain parameters and then grouping them at once into a new group. 

Or by being able to select multiple entries and then shifting them together :-)

ewels:
Moved to website repo.

If I remember correctly, multi-selection was tricky during building. Could maybe get around this by using keyboard shortcuts to multi-select (eg. `shift` + `↑`). Add a selected class if so then make the dragging code look for this. Needs some thought either way.

ewels:
Maybe group selection is still going to be a bit insufficient. Might be better to have a dedicated interface (modal?) showing just parameter names which can then be bulk-added to groups.

Maybe show ungrouped top-level params only? Would simplify the new interface a lot..

mashehu:
yes, without adding another js library, the smoothest solution would probbly be a "Add to this group" button on the group header (or the toolbar if a group is selected), with a dropdown/modal with a list of ungrouped params (extra points for filtering the list by search).

ewels:
Yes, nice idea 👍

PR #423, opened 2020-07-06
add multi param grouping button

mashehu:
As discussed in #422 

First version:
Jul-06-2020 17-30-23

To-do:

- [x] write help text
- [x] filter option in modal?
- [x] Shift+click / cmd+click to bulk select parameters in the table
- [x] Display some explanation text in the table if there are no top-level params left
- [x] add "Select  all" button
- [x] add tooltips

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -930,8 +930,8 @@ $(function () {
     // multi-select modal
     //
     
-    $('#multi_select_modal').on('change', '.move_param', function(){
-        var num_selected = $('#multi_select_modal').find('.move_param:checked').length;
+    $('#multi_select_modal').on('change', '.select_param', function(){
+        var num_selected = $('#multi_select_modal').find('.select_param:checked').length;
         if(num_selected>0){
             $('#multi_select_modal #move_params').removeClass("disabled");
             if(num_selected === 1){
@@ -968,7 +968,7 @@ $(function () {
             params += `
                 <tr>
                     <td>
-                        <input type="checkbox" aria-label="Move this parameter" class="move_param" data-id=`+k+` id="group-move-`+k+`">
+                        <input type="checkbox" aria-label="Move this parameter" class="select_param" data-id=`+k+` id="group-move-`+k+`">
                     </td>
                     <td>
                         <label for="group-move-`+k+`" class="text-monospace">`+k+`</label>
@@ -980,16 +980,20 @@ $(function () {
                 `
 
         }
-        $('#multi_select_modal tbody').html(params)
-        
-
+        if (params===''){
+            params = '<div class="alert alert-info">No ungrouped parameters available.</div>'
+            $('#multi_select_modal .table').remove()
+            $('#multi_select_modal .modal-body').html(params)
+        } else {
+            $('#multi_select_modal tbody').html(params)
+        }
         $('#multi_select_modal').modal('show');
         
     }
     $("#move_params").click(function(){
         var id = $('#multi_select_modal .modal-header h4 span').text();
         var group_el = $('.schema_group[data-id="' + id + '"] .card-body');
-        var selected_params = $('#multi_select_modal').find('.move_param:checked')
+        var selected_params = $('#multi_select_modal').find('.select_param:checked')
         for (let i = 0; i < selected_params.length; i++) {
             var p_id = $(selected_params[i]).data('id')
             var row_el = $('.schema_row[data-id="' + p_id + '"]');
@@ -997,6 +1001,10 @@ $(function () {
         }
         schema_order_change();
     });
+
+    $("#select_all_params").click(function(){
+        $('.select_param').prop('checked', true)
+    })
     //
     // Collapse group button
     //
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -952,9 +952,10 @@ $(function () {
         launch_multi_select_modal(id);
     });
     function launch_multi_select_modal(id) {
-        // Reset button to initial state
+        // Reset button and search to initial state
         $('#multi_select_modal #move_params').addClass("disabled");
         $('#multi_select_modal #move_params').html("Move parameters");
+        $("#search_parameters").val("");
         // Build modal
         $('#multi_select_modal .modal-header h4 span').html(id)
         var params = '';
@@ -965,7 +966,7 @@ $(function () {
             }
             // Add to the preview
             params += `
-                <tr>
+                <tr data-id=`+k+`>
                     <td>
                         <input type="checkbox" aria-label="Move this parameter" class="select_param" data-id=`+k+` id="group-move-`+k+`">
                     </td>
@@ -979,14 +980,15 @@ $(function () {
                 `
         }
         if (params===''){
+            // add placeholder text if no top-level parameters are available
             params = '<div class="alert alert-info">No ungrouped parameters available.</div>';
+            $('#multi_select_modal .table').remove();
             $('#multi_select_modal .table').remove();
             $('#multi_select_modal .modal-body').html(params);
         } else {
             $('#multi_select_modal tbody').html(params);
         }
-        $('#multi_select_modal').modal('show');
-        
+        $('#multi_select_modal').modal('show');   
     }
     $('#move_params').click(function(){
         var id = $('#multi_select_modal .modal-header h4 span').text();
@@ -1001,13 +1003,13 @@ $(function () {
     });
     // select all parameter checkboxes via button
     $('#select_all_params').click(function(){
-        $('.select_param').prop('checked', true);
+        $('.select_param:visible').prop('checked', true);
         $('.select_param').trigger("change");
 
     });
     // select all parameter checkboxes via button
     $('#deselect_all_params').click(function () {
-        $('.select_param').prop('checked', false);
+        $('.select_param:visible').prop('checked', false);
         $('.select_param').trigger("change");
 
     });
@@ -1025,6 +1027,36 @@ $(function () {
         }
         last_checked_box = this;
     });
+    // filter parameters table from https://www.w3schools.com/howto/howto_js_filter_table.asp
+    $('#search_parameters').on('input propertychange',function(){
+        var q = $("#search_parameters").val();
+        if(q===""){
+            $("#params_table tr").show();
+        }
+        $("#params_table tr").filter(function () {
+            if($(this).data("id")){
+                return !$(this).data("id").includes(q);
+            }
+        }).hide();
+        // input = document.getElementById('search_parameters');
+        // filter = input.value.toUpperCase();
+        // table = document.getElementById('params_table');
+        // tr = table.getElementsByTagName('tr');
+
+        // // Loop through all table rows, and hide those who don't match the search query
+        // for (i = 0; i < tr.length; i++) {
+        //     td = tr[i].getElementsByTagName("td")[1];
+        //     if (td) {
+        //         txtValue = td.textContent || td.innerText;
+        //         if (txtValue.toUpperCase().contains(filter) > -1) {
+        //             tr[i].style.display = "";
+        //         } else {
+        //             tr[i].style.display = "none";
+        //         }
+        //     }
+        // }
+    })
+
     //
     // Collapse group button
     //
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1005,6 +1005,12 @@ $(function () {
         $('.select_param').trigger("change");
 
     });
+    // select all parameter checkboxes via button
+    $('#deselect_all_params').click(function () {
+        $('.select_param').prop('checked', false);
+        $('.select_param').trigger("change");
+
+    });
     // hold shift for selecting a range of checkboxes
     $('#multi_select_modal').on('click','.select_param',function(e) {
         var checkboxes = $('.select_param');   
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -990,7 +990,7 @@ $(function () {
         $('#multi_select_modal').modal('show');
         
     }
-    $("#move_params").click(function(){
+    $('#move_params').click(function(){
         var id = $('#multi_select_modal .modal-header h4 span').text();
         var group_el = $('.schema_group[data-id="' + id + '"] .card-body');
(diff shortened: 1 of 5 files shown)


public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
             </div>
         </div>
     </div>
-
+    <!-- Moving multiple params into group modal -->
+    <div class="modal fade" id="multi_select_modal" tabindex="-1" role="dialog" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-scrollable modal-xl" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h4>Move parameters into this <span></span></h4>
```
ewels, line 312:
```suggestion
                    <h4>Move parameters into <span></span></h4>
```

public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
...
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                      <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <div class="modal-body">
+                    <table class="table table-sm">
+                        <thead>
+                            <tr>
+                            <th scope="col">Move</th>
```
ewels, line 321:
```suggestion
                            <th scope="col" style="width:1%"></th>
```

public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
             </div>
         </div>
     </div>
-
+    <!-- Moving multiple params into group modal -->
+    <div class="modal fade" id="multi_select_modal" tabindex="-1" role="dialog" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-scrollable modal-xl" role="document">
```
ewels, line 309:
```suggestion
        <div class="modal-dialog modal-dialog-scrollable modal-lg" role="document">
```

public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
...
+                        <thead>
+                            <tr>
+                            <th scope="col">Move</th>
+                            <th scope="col">Parameter</th>
+                            <th scope="col">Description</th>
+                            </tr>
+                        </thead>
+                    <tbody>
+                    </tbody>
+                </table>
```
ewels, line 328:
```suggestion
                        <tbody>
                        </tbody>
                    </table>
```

public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
...
+                </div>
+                <div class="modal-footer">
+                    <div class="col text-right pr-0">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+                    <button type="button" class="btn btn-primary disabled" data-dismiss="modal" id="move_params">
+                        Move parameters
```
ewels, line 334:
```suggestion
                        Move into <span></span>
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +926,78 @@ $(function () {
...
+        var group_el = $('.schema_group[data-id="' + id + '"] .card-body');
+        var selected_params = $('#multi_select_modal').find('.move_param:checked')
+        for (let i = 0; i < selected_params.length; i++) {
+            var p_id = $(selected_params[i]).data('id')
+            var row_el = $('.schema_row[data-id="' + p_id + '"]');
+            group_el.append(row_el)
+        }
+        schema_order_change();
+    }
+    )
```
ewels, line 1000:
```suggestion
    });
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +926,78 @@ $(function () {
...
+            if (schema['properties'][k].hasOwnProperty('properties')) {
+                continue
+            }
+            
 
+            // Add to the preview
+            params += `
+                <tr>
+                    <td>
+                        <input type="checkbox" aria-label="Move this parameter" class="move_param" data-id=`+k+`>
```
ewels, line 971:
```suggestion
                        <input type="checkbox" aria-label="Move this parameter" class="move_param" data-id=`+k+` id="group-move-`+k+`">
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +926,78 @@ $(function () {
...
+                    </td>
+                    <td>
+                        `+k+`
```
ewels, line 974:
```suggestion
                        <label for="group-move-`+k+`" class="text-monospace">`+k+`</label>
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +926,78 @@ $(function () {
...
+                    </td>
+                    <td>
+                        `+ schema['properties'][k].description +`
```
ewels, line 977:
```suggestion
                        <label for="group-move-`+k+`" class="small">`+ schema['properties'][k].description +`</label>
```

public_html/json_schema_build.php
```diff
@@ -304,7 +304,40 @@
...
+                            <th scope="col">Description</th>
+                            </tr>
+                        </thead>
+                    <tbody>
+                    </tbody>
+                </table>
+                </div>
+                <div class="modal-footer">
+                    <div class="col text-right pr-0">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
```
ewels, line 332:
```suggestion
                    <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -1189,6 +1260,9 @@ function generate_group_row(id, param, child_params){
                 <div class="col-auto align-self-center schema_row_config border-left">
                     <i class="fas fa-cog"></i>
                 </div>
+                <div class="col-auto align-self-center schema_group_move_params" alt-text="Move paramter(s) into this group">
```
ewels, line 1263:
```suggestion
                <div class="col-auto align-self-center schema_group_move_params" title="Move parameter(s) into this group">
```

mashehu:
Yes, I thought we had tooltips once 🤔 definitely should have them for all buttons without text.
I updated the to-dos with your suggestions.

mashehu:
I might have overdone it with the tooltips for the icons...

I am currently failing to get the shift/cmd multi-select working following this https://stackoverflow.com/a/659571/1696643, but I will keep trying 🙂

mashehu:
Of course it worked after a coffee break. 🙃 
Anyway, I don't think the filter option is really needed for now, so I will remove the draft and WIP labels.

public_html/json_schema_build.php
```diff
@@ -304,7 +305,45 @@
...
+                        </thead>
+                        <tbody>
+                        </tbody>
+                    </table>
+                    
+                </div>
+                <div class="modal-footer">
+                    <div class="btn-group" role="group">
+                        <button type="button" class="btn btn-light" id="select_all_params">Select all</button>
+                        <button type="button" class="btn btn-light" id="deselect_all_params">Deselect all</button>
```
ewels, line 335: I actually didn't notice these until I read the code 🙄 

```suggestion
                        <button type="button" class="btn btn-outline-secondary" id="select_all_params">Select all</button>
                        <button type="button" class="btn btn-outline-secondary" id="deselect_all_params">Deselect all</button>
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +927,104 @@ $(function () {
...
+    //
+    
+    $('#multi_select_modal').on('change', '.select_param', function(){
+        var num_selected = $('#multi_select_modal').find('.select_param:checked').length;
+        if(num_selected>0){
+            $('#multi_select_modal #move_params').removeClass("disabled");
+            if(num_selected === 1){
+                $('#multi_select_modal #move_params').html("Move 1 parameter");
+            } else {
+            $('#multi_select_modal #move_params').html("Move "+num_selected+" parameters");
```
ewels, line 941:
```suggestion
                $('#multi_select_modal #move_params').html("Move "+num_selected+" parameters");
```

public_html/json_schema_build.php
```diff
@@ -304,7 +305,45 @@
...
+                    </div>
+                    <div class="col text-right pr-0">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+                    <button type="button" class="btn btn-primary disabled" id="move_params">
+                        Move into <span></span>
```
ewels, line 340: My intention with this new `<span>` was to have the group name in the button, where you actually click. This makes the intention a bit more explicit.

However, thinking about it now the group names can be quite long so I disagree with my past self. So should probably revert this..

```suggestion
                        Move parameters
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +927,104 @@ $(function () {
...
+                <tr>
+                    <td>
+                        <input type="checkbox" aria-label="Move this parameter" class="select_param" data-id=`+k+` id="group-move-`+k+`">
+                    </td>
+                    <td>
+                        <label for="group-move-`+k+`" class="text-monospace">`+k+`</label>
+                    </td>
+                    <td>
+                        <label for="group-move-`+k+`" class="small">`+ schema['properties'][k].description +`</label>
+                    </td>
```
ewels, line 977: Minor, I just find this a little easier to read..

```suggestion
                    <td><input type="checkbox" aria-label="Move this parameter" class="select_param" data-id=`+k+` id="group-move-`+k+`"></td>
                    <td><label for="group-move-`+k+`" class="text-monospace">`+k+`</label></td>
                    <td><label for="group-move-`+k+`" class="small">`+ schema['properties'][k].description +`</label></td>
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,7 +927,104 @@ $(function () {
...
+                </tr>
+                `
+        }
+        if (params===''){
+            params = '<div class="alert alert-info">No ungrouped parameters available.</div>';
+            $('#multi_select_modal .table').remove();
```
ewels, line 983: `remove()` breaks this if the user moves any params back to the top level. It shows the "no ungrouped" message forever more...

Better to have the alert in the main HTML but hidden by default, then just toggle their visibilities..

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -1068,7 +1166,7 @@ function generate_param_row(id, param){
         <div class="col-auto align-self-center schema_row_grabber border-right">
             <i class="fas fa-grip-vertical"></i>
         </div>
-        <button class="col-auto align-self-center param_fa_icon ">`+fa_icon+`</button>
+        <button class="col-auto align-self-center param_fa_icon " title="Select icon" data-toggle="tooltip">`+fa_icon+`</button>
```
ewels, line 1169: We need to initialise these tooltips when new rows are added to the schema. I thought this was broken at first as I was starting with an empty schema and adding stuff.

Just needs a `row.find('[data-toggle="tooltip"]').tooltip()` somewhere.. (Let me know if you can't find it right away and I can take a look).

ewels:
It upsets me slightly that the shift-clicking on the labels doesn't work for me (checkboxes are fine). After a very quick play I think that this might be down to a browser bug though rather than anything else though, so not worth messing with.

Could be worth adding a `class="small text-muted"` bit of help text above or below the table saying that you can shift-click the checkboxes though.. 👍

public_html/assets/js/nf-core-schema-builder.js
mashehu: Is it okay if I put that into the `init_group_sortable()` function? Would be called at the right time, but also a bit of a function overload...

ewels: Hmm, I think that will not work for when new rows are added though (only groups). I would probably add it to the row / group generation code, ie. [here](https://github.com/nf-core/nf-co.re/blob/7c782c12fd6f8fa231970b30e5c5c3b9c2c84b48/public_html/assets/js/nf-core-schema-builder.js#L169) and [here](https://github.com/nf-core/nf-co.re/blob/7c782c12fd6f8fa231970b30e5c5c3b9c2c84b48/public_html/assets/js/nf-core-schema-builder.js#L193).

I guess it should just be something like this?

```javascript
param_row.prependTo('#schema-builder').find('.param_id').select();
param_row.tooltip();
```

public_html/assets/js/nf-core-schema-builder.js
mashehu: fixed in 36e9458

public_html/assets/js/nf-core-schema-builder.js
mashehu: done 0ccb5b9

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,6 +929,124 @@ $(function () {
...
+            if (schema['properties'][k].hasOwnProperty('properties')) {
+                continue
+            }
+            // create row for the table
+            params += `
+				<tr data-id=`+ k + `>
+                    <td><input type="checkbox" aria-label="Move this parameter" class="select_param" data-id=`+ k + ` id="group-move-` + k + `"></td>
+                    <td><label for="group-move-`+ k + `" class="text-monospace">` + k + `</label></td>
+                    <td><label for="group-move-`+ k + `" class="small">` + schema['properties'][k].description +`</label></td>
+                </tr>
```
ewels, line 1002: Missing the end of this escaped block, causing a JS error..

```suggestion
                </tr>`;
```

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -926,6 +929,124 @@ $(function () {
...
+        $("#search_parameters").val("");
+        var params = '';
+        for (k in schema['properties']) {
+            // skip over Groups
+            if (schema['properties'][k].hasOwnProperty('properties')) {
+                continue
+            }
+            // create row for the table
+            params += `
+				<tr data-id=`+ k + `>
```
ewels, line 998:
```suggestion
            params += `<tr data-id=`+ k + `>
```

ewels:
Made a few minor tweaks in b91e4c5e49cda19f1abf5728c93660e431da9503:

* My `param_row.tooltip();` guess didn't work. Changed how the tooltips are initialised for the whole website instead, now works as expected.
* Changed the table filter a little to use the `keyup` instead of `input propertychange`. Seems to work better for me now.
* Small general code cleanup (simplified one or two small bits, couple of new comments, moved a chunk of code to follow a more logical order)

Thanks for this, it's a great addition!

PR #428, opened 2020-07-07
Pipeline page: add Slack and GitHub icons to buttons

ewels:
Added a couple of icons to the pipeline page:


mashehu:
Any specific reason slack got the `info` color? `light` would be closer to the current icon imo (after the thankfully moved away again from their purple).

PR #434, opened 2020-07-13
Add tutorial slides

FriederikeHanssen:


ewels:
How about embedding the presentation instead of linking to it? https://help.edublogs.org/google-slides/

PR #441, opened 2020-07-16
Virtual social event and slides Elin Talk

ggabernet:


ggabernet:
Ah yes, sorry, just fixed that. I hadn't understood the folders follow the same structure as the website pages :)

PR #443, opened 2020-07-16
hackathon 2020

ggabernet:
virtual event fix


ggabernet:
sorry about that @ewels , now it should be really fine

PR #447, opened 2020-07-21
extend markdown rendering functions in docs preview

mashehu:
- fix `outline-primary` button in dark mode
- make help-text preview modal wider, but still not sure what to do with wide tables (see screenshot)

Screenshot 2020-07-21 16 04 39


ewels:
Can wrap tables in a responsive wrapper: https://getbootstrap.com/docs/4.5/content/tables/#responsive-tables

I already do some manual find-and-replace in the PHP to add the `table` class, so should be a fairly easy addition on the server-side rendering at least.

PR #451, opened 2020-07-23
make tables in js-rendered markdown responsive, darken whites a tiny bit

mashehu:
as mentioned in https://github.com/nf-core/nf-co.re/pull/447#issuecomment-662895335.


mashehu:
It actually works also like that:
Screenshot 2020-07-23 at 11 46 01
But I will wrap it in the element to make sure, that nothing is broken anytime.

ewels:
Huh, cool. I wonder if it's the `tbody` element that then scrolls or something? But yeah, safer to stick to the Bootstrap docs way of doing things if possible I think..

PR #456, opened 2020-07-23
Auto docs

ewels:
Continuation of https://github.com/nf-core/nf-co.re/pull/440 on a branch so that @mashehu and I can both easily work on it.


mashehu:
Since I (accidentally) merged your PR into this, one  i will add comments about changes here:
I am not a fan of the vertical bars and large padding between parameters, I think they add too much visual noise, especially around the group names. For me they are already separated enough with the different headings and paragraphs and it has a better visual rhythm.
To clarify this is what I am talking about:
Screenshot 2020-07-24 at 15 07 04
without borders and padding:
Screenshot 2020-07-24 at 15 09 53

ewels:
The way I got here was that I felt that the help buttons were floating out in space too much. It made it tricky to line up which was which on wider screens, especially when there are lots of hidden params / buttons together.

The horizontal borders help a lot to keep things organised, but the negative margin stuff on the headers meant that the borders were a bit crazy and overlapped the previous elements. So I added in extra padding to make it fit properly. I was fairly happy with how it looked, so I left it like that.

ewels:
Note that I half wondered about taking a completely different approach and rendering the params docs as a (pseudo-?) table. Having the description alongside the parameter id would cut down on the whitespace and make it a lot easier to scan, and help text could be expanded as a full-width row under the relevant row. It would also make sense to downgrade parameters from headings then and make everything a lot smaller, so more stuff would fit on the screen.

I didn't take the time to prototype it though, it was just an idea in the shower. And to be honest I wasn't super keen to rewrite the output again (though actually it shouldn't be so bad now that it's in a function call).

mashehu:
Ah, okay. What if we put the help button (should it maybe be "Show help" help to fit with the hidden-button text) beneath the definition, above the card containing the text?

Yes, I was also thinking if we maybe should have it more like the the old bootstrap3 `.dl-horizontal`, to give it more of a lexicon style (rendering as tables sound like a very 00s way to layout text 😉). I think with our current setup of inputing descriptions (as a quite tiny input field), they will be a lot shorter than the help-text.

ewels:
> What if we put the help button (should it maybe be "Show help" help to fit with the hidden-button text) beneath the definition, above the card containing the text?

I think that'll be worse, as it'll then be closer to the following param ID than the one it applies to..

> rendering as tables sound like a very 00s way to layout text

I said _(pseudo-?)_! 😅 Yes that's kind of what I meant. Can just be `.row`s I guess.

Shall I try to whip up a quick proof-of-concept? Or do you want to go at it?

mashehu:
I'll give it a go, so you can concentrate on the tools.

ewels:
Oops, sorry I couldn't resist seeing what it might look like 🙄 

image

ewels:
I think it's much better tbh.. Reflows nicely on small screens too:

image

mashehu:
Hmm, now it looks too rigid to me... I kind of liked the flow of the param definitions in the `usage.md`. It is a lot easier to parse that way, but not really beginner friendly I guess. So we need to make sure to have a good introduction to the pipeline in the `usage.md`.

ewels:
Yeah, I think that's kind of how I imagined it being used to be honest. General introduction stuff written as prose, then more in-depth docs in the structured format..

PR #457, opened 2020-07-23
Auto-docs: Phil

ewels:
My continuation of https://github.com/nf-core/nf-co.re/pull/440, to add to the main PR: https://github.com/nf-core/nf-co.re/pull/456

- [x] Rename `details` to `help`
- [x] More visual hierarchy - help text inside a `.panel` and smaller font size
- [x] More visual hierarchy in ToC - smaller font size for submenus
- [x] Slightly less visual noise - lighter icons in ToC, non-green icons for param headings
- [x] Re-style the hidden items a little
- [x] Made hidden params + label clickable to show just that param
- [x] Check for missing icons (some `fa-icons fa-fw ml-0 text-muted` elements in eager which I guess is where the icon isn't set)


includes/functions.php
```diff
@@ -77,6 +77,12 @@ function round_nicely($num){
   return $num;
 }
 
+function endswith($haystack, $needle){
```
mashehu, line 80: isn't the order in php usually `($needle, $heystack)`, at least in the `preg_*` functions?

public_html/pipeline.php
```diff
@@ -5,97 +5,78 @@
...
+else if(endswith($_GET['path'], '/releases')){
+  $pagetab = 'releases';
+  require_once('../includes/pipeline_page/releases.php');
 }
 # Some other URL pattern that we don't recognise - 404
-else {
-    header('HTTP/1.1 404 Not Found');
-    include('404.php');
-    die();
+else if($_GET['path'] != $pipeline->name && $_GET['path'] != $pipeline->name.'/'.$release){
```
mashehu, line 66: why do we need the `if`-clause here, wouldn't we catch more cases if it was just `else`?

public_html/assets/js/nf-core.js
```diff
@@ -188,25 +188,10 @@ $(function () {
...
-            var selected = $('option:selected',this).text();
-            var pipeline = $('#version_select').data('pipeline');
-            var url = document.location.href.split(pipeline);
-            url[1] = url[1].replace(/^\/\d+\.\d+\.\d+|^\/\d+\.\d+|^\/dev/,"/"); // remove previously added version in url
-            document.location.href = url[0] + pipeline + "/" + selected + url[1]; // add selected version into url after pipeline name and before the rest
-        })
-    }
+    // Pipeline page version number dropdown
+    $('#version_select').on('change', function(){
+        document.location.href = $(this).val();
```
mashehu, line 193: Sorry, I am still more used to doing stuff in js than php 🙂

public_html/pipeline.php
```diff
@@ -166,81 +153,88 @@
...
-                </div>';
-      $toc .= '<nav class="toc nav flex-column">'.generate_toc($content).'</nav>';
-        echo $toc;
-    } else if($pagetab == 'output'){
-      $toc = '<nav class="toc nav flex-column">'.generate_toc($content).'</nav>';
-        echo $toc;
-    } else {
-        echo $pipeline_stats_sidebar;
+      $toc .= '<div class="btn-group mt-1" role="group">
+                <button class="btn btn-outline-secondary collapse-groups-btn" id="toggle_details" data-toggle="collapse" data-target=".schema-docs-help-text" aria-expanded="false"><i class="fas fa-question-circle mr-1"></i> Toggle help</button>
```
mashehu, line 230: Not sure of the wording in the two toggle buttons. Either it should be "toggle" for both or "show" for both not mixed.

includes/pipeline_page/usage.php
```diff
@@ -112,9 +112,9 @@
                 $help_text = parse_md($vv['help_text']);
                 if(strlen($help_text)>150){ // collapse long help texts
                   $schema_content.='<div class="col-2">';
-                  $schema_content.='<button class="btn btn-outline-secondary" data-toggle="collapse" href="#'.preg_replace("/\/|\s/","-",$kk).'-help" aria-expanded="false"><i class="fa"></i> Details</button>';
+                  $schema_content.='<button class="btn btn-sm btn-outline-secondary" data-toggle="collapse" href="#'.preg_replace("/\/|\s/","-",$kk).'-help" aria-expanded="false"><i class="fas fa-question-circle"></i> Help</button>';
```
mashehu, line 115: maybe add a `pull-right` class, to the button, so it aligns with the "required" tag.

ewels: Rewriting all of this code now 👍🏻

public_html/assets/js/nf-core.js
ewels: No problem! Unsurprisingly the site is built using the web stack that I'm used to and have been building for years 👴🏻 

I think a lot of people would argue that we _should_ be doing this stuff in JS instead of PHP 😅

includes/functions.php
ewels: Ah maybe, this was a copypasta from stack overflow I think. Will switch 👍🏻

ewels: Just started switching but actually I prefer it this way around. The way you call the function feels more logical to me:

```php
if(endswith("/some/long/url/my/page", "page"))
```

instead of:

```php
if(endswith("page", "/some/long/url/my/page"))
```

ewels:
Sorry, messy diff as I moved some files around and renamed them etc. But I've now removed a tonne of code duplication and refactored stuff quite a lot. Hopefully the PHP is now quite a bit easier to navigate and understand.

I've broken the ToC for the auto-generated params though, will take a look at that next.

includes/functions.php
```diff
@@ -159,12 +161,19 @@ function _h($level, $html){
...
+  # Class for hiding in ToC
+  $toc_hidden_class = $toc_hidden ? 'toc-hidden' : '';
   return '
-    <h'.$level.' id="'.$hid.'">
+    <h'.$level.' id="'.$hid.'" class="'.$toc_hidden_class.'">
       <a href="#'.$hid.'" class="header-link"><span class="fas fa-link"></span></a>
       '.$html.'
     </h'.$level.'>';
 };
+function _h1($html){ return _h(1, $html); }
```
mashehu, line 172: why do we need this nested functions? I think they just add an unneeded complexity to the code because `_h()` is already used in different parts, so one is familiar with it and `_h1()` is not much less to write than `_h(1,)`...

public_html/assets/css/nf-core.css
```diff
@@ -781,7 +784,7 @@ Based on https://codepen.io/wintr/pen/beBJBb */
   margin-top: .25rem;
 }
 
-.side-sub-subnav {
+.pipeline-page-content, .side-sub-subnav {
     padding-top: 50px;
```
mashehu, line 788: I don't like the empty space it creates when scrolled to the top. 
Screenshot 2020-07-24 at 14 12 59

Can we switch this to a scrollpoint based behavior?

public_html/pipeline.php
ewels: This `if` clause basically catches the homepage, just in reverse. We could make it just an `else`, but then we'd need another similar-looking `if` clause above to allow the homepage.

public_html/pipeline.php
ewels: Done 👍🏻

includes/functions.php
ewels: We don't need it, I just thought that `_h1` was nicer to write - it's more obvious what it does when you're reading the PHP code.

It's only used in one place so far but I was thinking of trying to use them more through the rest of the code base as I think that the anchors are written manually in a bunch of places.

public_html/assets/css/nf-core.css
ewels: What do you mean sorry?

mashehu: Can we add the padding above `.pipeline-page-content, .side-sub-subnav` only when it is actually needed, that is when the navbar would overlap them.

ewels: aha, yes - sorry was probably not looking at this page when I added that. I think that the Usage docs needed padding (can go elsewhere though).

PR #459, opened 2020-07-24
auto-docs matthias

mashehu:


mashehu:
sidebar clean up.
before:
Screenshot 2020-07-24 at 14 51 46
after:
Screenshot 2020-07-24 at 14 51 11

includes/pipeline_page/sidebar.php
```diff
@@ -46,9 +52,13 @@
 ?>
 
 <div class="pipeline-sidebar">
-  <h6><i class="fas fa-terminal fa-xs"></i> command</h6>
-  <div class="border pipeline-run-cmd p-1">
-    <code class="small"> nextflow run <?php echo $pipeline->full_name; echo $release_cmd; ?> -profile test</code>
+  <div class="row border-bottom pb-2">
+    <div class="col-12">
```
ewels, line 56: What's the advantage of adding these divs?

mashehu: row elements are somehow wider than non row elements (see bottom border of collaborators field in the old version)

ewels: Ah right, yeah this is a mistake that I make a lot. I think it's better to move the `bottom-border` class off the `.row` element rather than making everything rows. The `.row` has a negative margin to cope with the padding on `.col` elements.

mashehu: But I like the overhanging borders and wouldn't like to sacrifice it for padding inwards (because that would probably be again to narrow for the command etc.)

ewels: hah, ok ¯\_(ツ)_/¯

mashehu:
Sorry, I probably put too much of a mixture of changes in functionality and styling in the last commit. But now everything should work almost as before (I decided to apply a wikipedia-like approach to only link via the toc to the parameters, to circumvent the link offset problem). I think the added ToC-expansion on hover can be very useful, but might also be a bit to finicky, so I am open to remove it again. I changed the order of labels and help button, because I am imagining one would parse more for the label than the help button, so it is easier if they are always the rightmost.

The updated look:
Screenshot 2020-07-27 at 14 07 09

ewels:
Nice! Working on tools now but will take a proper look when I get a chance

In dark mode the hidden fields look kind of highlighted rather than hidden 😕  - could we change the styling to leave the background colour unmodified but drop the text colour or change the opacity or something?

mashehu:
Yes, fixed it. (One of the weird things with dark mode is that you have two directions for highlighting something, either with a darker or brighter color vs light mode where you usually can only go for darker 🤔)

mashehu:
Screenshot 2020-07-27 at 14 51 28

mashehu:
which php script renders https://nf-co.re/usage/introduction? I guess this was done with `usage.php`, but we dropped this specific case of no URL in the pipeline and now it is kind of broken (but surprisingly not competely). Same for all the other pages rendered from markdown 🙃

ewels:
Everything under `/usage` and `/developers` should be handled by `public_html/docs.php` I think. Off the top of my head I think that it tries to match the supplied query to a markdown file in `markdown/usage/xxx.php` etc and if that fails it then tries to match it to a pipeline name.

mashehu:
okay, found what I broke and it should be fixed now.

ewels:
..and actually I think `1400px` is better than `1800px` too - that's roughly the normal content width + the width of the sidebar and it feels less jarring to me.

ewels:
> I think the added ToC-expansion on hover can be very useful, but might also be a bit to finicky, so I am open to remove it again.

Woah! ok this is cool, but sorry I think it has to go 😞 I found it (a) very surprising and (b) kind of annoying at first - if you move your mouse in from the wrong angle to click on a heading then it jumps around all over the place. Kind of fun and useful once you get the hang of it, but I don't think that we can ask people to do that - they'll just get annoyed and leave.

includes/pipeline_page/_index.php
```diff
@@ -249,7 +249,7 @@
 ########
 if($pagetab !== 'stats'){
   echo '</div>'; # end of the content div
-  echo '<div class="col-auto d-none d-lg-block pl-2"><div class="side-sub-subnav sticky-top">';
+  echo '<div class="col-lg-3 pl-2"><div class="side-sub-subnav sticky-top">';
```
ewels, line 252:
```suggestion
  echo '<div class="col-12 col-lg-3 pl-2"><div class="side-sub-subnav sticky-top">';
```

includes/pipeline_page/docs_schema.php
```diff
@@ -47,12 +47,12 @@ function print_param($is_group, $param_id, $param, $is_required=false){
...
       $help_text_btn = '
-        <button class="btn btn-sm btn-outline-info float-right ml-2" data-toggle="collapse" href="#'.$param_id.'-help" aria-expanded="false">
+        <button class="btn btn-sm btn-outline-info ml-2 mb-1 mt-1" data-toggle="collapse" href="#'.$param_id.'-help" aria-expanded="false">
           <i class="fas fa-question-circle"></i> Help
         </button>';
       $help_text = '
-        <div class="collapse card col-12 schema-docs-help-text" id="'.$param_id.'-help">
-          <div class="card-body small text-muted">'.parse_md($param['help_text']).'</div>
+        <div class="collapse card col-12 schema-docs-help-text mb-2" id="'.$param_id.'-help">
+          <div class="card-body small text-muted p-2">'.parse_md($param['help_text']).'</div>
```
ewels, line 55:
```suggestion
          <div class="card-body small text-muted py-2 px-0">'.parse_md($param['help_text']).'</div>
```

includes/pipeline_page/docs_schema.php
```diff
@@ -73,20 +73,22 @@ function print_param($is_group, $param_id, $param, $is_required=false){
...
     # Extra group classes
-    $mt = '';
-    $id_cols = 'col-10 col-md-5 col-lg-4 col-xl-3 ';
+    $row_class = 'align-items-center';
+    $id_cols = 'col-10 col-md-5 col-lg-4 col-xl-3 small-h';
+    $h_level = 'h3';
     if($is_group){
-      $mt = 'mt-5';
-      $id_cols = 'col h2';
+      // $row_class = 'mt-3';
```
ewels, line 80:
```suggestion
```

includes/parse_md.php
```diff
@@ -101,7 +104,11 @@ function parse_md($markdown){
...
   // Find and replace emojis names with images
-  $content = preg_replace('/:([\S]+?):/','<img class="emoji" alt="${1}" height="20" width="20" src="https://github.githubassets.com/images/icons/emoji/${1}.png">',$content);
+  $content = preg_replace('/:(?!\/)([\S]+?):/','<img class="emoji" alt="${1}" height="20" width="20" src="https://github.githubassets.com/images/icons/emoji/${1}.png">',$content);
 
-  return $content;
+  $output['content'] = $content;
+  $output["meta"] = $meta;
+  $output["title"] = $title;
+  $output["subtitle"] = $subtitle;
+  return $output;
```
ewels, line 113: NB: This breaks a couple of other pages - namely `docs_schema.php` and `launch.php` (easy fix).

includes/pipeline_page/docs_schema.php
```diff
@@ -73,20 +73,22 @@ function print_param($is_group, $param_id, $param, $is_required=false){
...
     }
 
     # Build row
     return '
-    <div class="row '.$hidden_class.' params-docs-row border-bottom pt-2 '.$mt.'">
-      <div class="'.$id_cols.'">'.$fa_icon.$h_text.'</div>
-      <div class="col">'.$param_body.'</div>
+    <div class="row '.$hidden_class.' params-docs-row border-bottom d-flex '.$row_class.'">
+      <div class="'.$id_cols.'">'.add_ids_to_headers('<'.$h_level.'>'.$fa_icon.$h_text.'</'.$h_level.'>').'</div>
+      <div class="col">'.$param_body.$help_text.'</div>
```
ewels, line 90: I'm not sure about this change. Help text is likely to be quite long and restricting it to half of the content width seems a bit unnecessary. What was the problem with having it full width?

includes/pipeline_page/sidebar.php
ewels: NB: I never intended the side bar to be wide enough to accommodate the command. What I meant to do (and never got around to) was to add a copy-button like the one that GitHub has for cloning repos:

includes/pipeline_page/sidebar.php
```diff
@@ -93,16 +103,17 @@
...
+  <div class="row border-bottom">
+    <div class="col-12">
+      <h6>collaborators</h6>
+      <p class="contrib-avatars"><?php echo implode(array_keys($contrib_avatars)); ?></p>
+    </div>
+  </div>
+  <div>
+    <h6>get in touch</h6>
+      <p><a class="btn btn-sm btn-outline-info" href="https://nfcore.slack.com/channels/<?php echo $pipeline->name; ?>"><i class="fab fa-slack mr-1"></i> Ask a question on Slack</a></p>
+      <p><a class="btn btn-sm btn-outline-secondary" href="<?php echo $pipeline->html_url; ?>/issues"><i class="fab fa-github mr-1"></i> Open an issue on GitHub</a></p>
```
ewels, line 115:
```suggestion
    <p><a class="btn btn-sm btn-outline-info" href="https://nfcore.slack.com/channels/<?php echo $pipeline->name; ?>"><i class="fab fa-slack mr-1"></i> Ask a question on Slack</a></p>
    <p><a class="btn btn-sm btn-outline-secondary" href="<?php echo $pipeline->html_url; ?>/issues"><i class="fab fa-github mr-1"></i> Open an issue on GitHub</a></p>
```

Need a code style linter... Been using Black for Python a lot lately and I have fallen in love with it - so nice not having to put any brain power towards this stuff!

public_html/assets/css/nf-core.css
```diff
@@ -1119,10 +1147,15 @@ Based on https://codepen.io/wintr/pen/beBJBb */
   margin-right: 0.5rem;
   opacity: 0.8;
 }
-
+.param-docs-body p{
```
ewels, line 1150:
```suggestion
.param-docs-body p:last-child {
```

public_html/assets/css/nf-core.css
```diff
@@ -378,7 +378,20 @@ footer, footer a, footer a:hover, footer a:focus, footer a:active {
 .main-content .h4, .main-content .h5, .main-content .h6 {
   font-family: "Open Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
   font-weight: 500;
-  padding-top: 60px;
+}
+h2[id]::before, .bd-content > h3[id]::before, .bd-content > h4[id]::before {
```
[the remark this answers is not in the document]
mashehu: just makes sure that there's an `id` tag, so it isn't applied to all headings, just the ones with a anchor link.

public_html/assets/css/nf-core.css
```diff
@@ -378,7 +378,20 @@ footer, footer a, footer a:hover, footer a:focus, footer a:active {
...
+}
+.main-content h1::before, .main-content h2::before, .main-content h3::before,
+.main-content h4::before, .main-content h5::before, .main-content h6::before,
+.main-content .h1::before, .main-content .h2::before, .main-content .h3::before,
+.main-content .h4::before, .main-content .h5::before, .main-content .h6::before {
+  display: block;
+  content: "";
+  padding-top: 50px;
   margin-top: -50px;
 }
```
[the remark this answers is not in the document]
mashehu: Oops, didn't commit the most recent version and the above to commnted pieces are a copy-paste artifact. Fixed in 
54e55c7.

includes/parse_md.php
mashehu: Fix for docs_schema.php in dcdfddb.  `launch.php` has it's own parse_md(). Should we replace it with the global function?

includes/pipeline_page/docs_schema.php
mashehu: it looked out of context, when expanded the full row width. I like this solution more, because the help-text is an extension of the description section, so it should also be rendered in the same location. One compromise could be to make the help-text the length of the description+labels/help-button columns in a new row (would also prevent the parameter name to move too far away if a long help text is expanded.

includes/pipeline_page/sidebar.php
mashehu: Good, because I also want to have this copy button. I'll add it then. I still like the full width command because it makes it clearer what one is copying and get familiar with the nextflow command syntax.

includes/pipeline_page/docs_schema.php
ewels: What about if we use custom styles instead of a card? eg. before:

image

---

After:

image

---

I think that it's realistic to assume that some pipeline authors are going to write a _lot_ of content in here. There has been discussions on Slack about having tables, images, all sorts in this help text. I just think that longer help texts will look odd if kept narrow and be more difficult to read.

includes/parse_md.php
ewels: Ah, yeah that sounds sensible. Sorry, reinventing the wheel again there. I just found it on a project text search, didn't realise that it was different but named identically 🤦🏻

includes/pipeline_page/docs_schema.php
mashehu: well, I am not a fan of long text lines in general, because my eyes jump around too much. It doesn't have to be Bringhurst-short, but for example my intuition would be to cut the above at around "qualtity cutoff". I don't see why height of these objects would create a problem, since they were probably already that long in the old style (which had a similar width), especially if it is easily collapsible. Or what is the UX-problem you see there?

ewels: _"Bringhurst-short"_? 👀 

I guess it's a personal preference thing - I'm the opposite in that I find the second screenshot easier to read. Although no schema have it yet, I'm also worried that tables and other rich content will be wide and simply not fit in the narrower box. Making it full-width gives us the best chance of accommodating whatever people throw at us.

I also like that the UI doesn't change for the parameter name, as there is no vertical change. This can also be achieved this without being full-width, but we'll have a big chunk of empty whitespace on the left then.

mashehu: Bringhurst-short = 45 to 75 characters https://smad.jmu.edu/shen/webtype/linelength.html

Okay, we can go back to full width help-text. I prefer the light background option to the card one.

ewels: Nice! Learn something new every day 😅 

We can always meet in the middle if you want - stop the vertical height changing on the param id and make it a little wider on the right so that it flows underneath the Help button. I don't have very strong feelings on the subject.

Light background option = the light grey background?

mashehu:
I added the copy-command button and switched to fill-width help-text with a lightgray background. While testing I saw that the "Show hidden params" button doesn't seem to have any effect anymore. Did I remove an element with `.param-docs-body-hidden` by accident?

ewels:
Looking brilliant - awesome job!

Some minor tweaks that I will have a quick stab at myself now in a separate PR:

* Command font size looks a bit massive now, will drop to be more consistent with the rest of the sidebar
* Copy button should trigger a toast with a success message
* Most controversial - I'm going back to thinking that 'hidden' parameters should really be hidden
    * Hidden from the ToC (including group if all hidden) - I think that I got this to work a little way back, but seems to not work any more
    * I think that hidden parameters should be fully removed from the page by default. The argument previously was that anyone using ctrl-F to look for a parameter then wouldn't find it, so my proposal is to have some small text at the bottom saying _"32 params hidden: xxx, xxx, xxx. Click here to show hidden params"._ (or whatever). Then doing a ctrl-F will jump to this content so it'll be obvious.
     * The rest of the page will then be a lot shorter and easier to navigate, which is the point of the hidden flag.
* I'd like to look into making a `.htaccess` rule that makes existing URLs redirect to their new location
    * eg. `pipeline/docs/usage` >  `pipeline/usage`
    * _Maybe_ look into rendering arbitrary Markdown still as we do currently? Then at least links will work and people can always keep these files with internal URLs in their `usage.md` docs. This gives people a workaround if they really _really_ hate our strict 2-page docs, and again means that previous URLs may still function.

Nice work @mashehu! 🎉

PR #462, opened 2020-07-24
Tables not tables

ewels:
See https://github.com/nf-core/nf-co.re/pull/456#issuecomment-663571102


ewels:
This breaks the ToCs and all of the IDs because we're not using headers any more, but gives a proof of concept for the alternative layout.

I prefer it I think. What do you think @mashehu?

ewels:
Merging so that @mashehu can continue on it..

PR #463, opened 2020-07-28
New schema structure - Schema Builder

ewels:
Updating the schema builder for the new style JSON schema, using `definitions` for groups.

* Renamed schema builder URL (x-ref tools https://github.com/nf-core/tools/pull/695)
* Rewrote everything to use `schema['definitions']` for groups

I've done a bit of manual testing, but I imagine that there could still be some other bugs lurking..


ewels:
x-ref https://github.com/nf-core/nf-co.re/issues/449

Issue #449, opened 2020-07-22
JSON Schema - update to new structure


mashehu:
As described in https://github.com/nf-core/tools/issues/687, the json-schema will look a bit different compared to now.

Files which need to be checked and updated:
- [x] json_schema.php
- [x] ~~json_schema_build.php~~ `pipeline_schema_builder.php`
- [x] launch.php 
- [x] ~~usage.php~~ `docs_schema.php`
- [x] nf-core-schema-builder.js
- [x] nf-core-schema-launcher.js

ewels:
* `json_schema.php` - don't think that there's anything specific here that needs updating
* `pipeline_schema_builder.php` + `nf-core-schema-builder.js` - https://github.com/nf-core/nf-co.re/pull/463

ewels:
* `launch.php` + `nf-core-schema-launcher.js` - https://github.com/nf-core/nf-co.re/pull/466

PR #464, opened 2020-07-28
New schema structure - Auto docs

mashehu:
as mentioned in #449. Forked from #459, so should be merged after it.

Tested with eager.


ewels:
Merging, will continue work with the new schema structure stuff 👍🏻

PR #471, opened 2020-07-29
Fix schema builder ungrouped params sorting

ewels:
* When loading new schema, put ungrouped params at the bottom (same as docs and other page)
* Force ungrouped top-level params to be below all definitions groups when sorting
* Write some more help text for the schema builder
* Restyle help-text to have a help button and be more consistent with the rest of the site
* Rename classes so that `container-fluid` isn't clobbered and can still be used (eg. _Pipeline Health_ page)
  * Used `container-xl` so that it's consistent for if/when we get around to upgrading bootstrap

### DIFF
--- a/includes/pipeline_page/_index.php
+++ b/includes/pipeline_page/_index.php
@@ -197,7 +197,7 @@ if($pipeline->archived){
 </div>
 <div class="triangle subheader-triangle-down"></div>
 
-<div class="container-fluid main-content">
+<div class="container container-xl main-content">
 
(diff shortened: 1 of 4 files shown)


ewels:
New style parameter schema builder help:

---

image

---

image

ewels:
Also, killed the _"Preview Help"_ modal functionality. Now that we can view the docs for the `dev` branch, I think that this is a lot less helpful and it's just another (third) place to maintain. The fact that I forgot that this function existed and didn't test it when making the schema change is case in point.

PR #472, opened 2020-07-29
Tidy up toasts

ewels:
The new toast I added for the copy-command functionality on the pipeline page messed up the toast that we already had for the launch pipeline validation. It moved it to the top of the page, where it overlapped the "Launch" button even when not showing, meaning that you couldn't click the "Launch" button 🤦🏻 

Refactored the launch code and CSS so that it's less bad.

### DIFF
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -103,8 +103,8 @@ img.emoji{
 .toast {
   background-color: rgba(255,255,255,.90);
   position: fixed;
-  bottom: 3rem;
-  right: 3rem;
+  bottom: 0.5rem;
+  right: 1rem;
   z-index: 5000;
 }
 .toast-header img {
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -47,12 +47,6 @@ table img {
 .card p:last-child {
   margin-bottom: 0;
 }
-.toast {
-  position: fixed;
-  top: 5rem;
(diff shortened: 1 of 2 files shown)


mashehu:
Do we really need to fix the position of the toast element. It looks like bootstrap should already take care of the overlap otherwise: https://getbootstrap.com/docs/4.5/components/toasts/#stacking

ewels:
I have a feeling that they end up somewhere random if you don't specify a position for them. For linting I think it works best if they're in the bottom right of the screen - this is generally where I prefer them for what we're using them for.

ewels:
Yeah, just tried removing the positional CSS and it just gets dumped at the end of the HTML:

image

So yes we need to. The overlay stuff is for the bounding box that the toasts are within I think.

public_html/assets/css/nf-core.css
```diff
@@ -106,13 +100,12 @@ img.emoji{
...
-    position: fixed;
-    bottom: 5px;
-    right: 5px;
-}
 .toast {
-  background-color: rgba(255,255,255,.90)
+  background-color: rgba(255,255,255,.90);
+  position: fixed;
+  bottom: 3rem;
+  right: 3rem;
```
mashehu, line 107:
```suggestion
  right: 1rem;
```

mashehu: Putting it a bit lower into the corner
Screenshot 2020-07-30 at 10 57 26

PR #479, opened 2020-07-30
fix ToC generator

mashehu:
I think I cracked this now:
`while($curr_level > 1)` has to be `while($curr_level > 0)` otherwise even a simple case like `$html_string=<h1></h1>;` results in 
```{php}
$toc = <div class="list-group>'
          <a class="list-group-item list-group-item-action '.$toc_hidden.'"href="#'.$id.'">'.$name.'</a>';
```
with while($curr_level > 0):
```{php}
$toc = <div class="list-group>'
          <a class="list-group-item list-group-item-action '.$toc_hidden.'"href="#'.$id.'">'.$name.'</a>'
        </div>;
```
The other problem occurs when `$html_string=<h1></h1><h3></h3>;` (=no `h2` element between `h1` and `h3`). with `while($curr_level > 0):` this would result in:
```{php}
$toc = <div class="list-group>'
          <a class="list-group-item list-group-item-action '.$toc_hidden.'"href="#'.$id.'">'.$name.'</a>'
          <div class="list-group>'
            <a class="list-group-item list-group-item-action '.$toc_hidden.'"href="#'.$id.'">'.$name.'</a>'
          </div>
        </div>
        </div>;
```
because there is no `h2` list-group. Adding an empty list-group for the missing `h2` breakes the css toc expansion. So my solution is to basically add a counter for how many open `<div>`s there are and rely on that instead of just counting down from the current level.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -105,6 +105,7 @@ function get_self_url($strip_query=true){
 function generate_toc($html_string){
   $toc = '';
   $curr_level = 0;
+  $counter = 0;
   $id_regex = "~<h([1-3])([^>]*)id\s*=\s*['\"]([^'\"]*)['\"]([^>]*)>(.*)</h[1-3]>~Uis";
   preg_match_all($id_regex, $html_string, $matches, PREG_SET_ORDER);
   if($matches){
@@ -118,12 +119,14 @@ function generate_toc($html_string){
       $name = trim(str_replace('&nbsp;','', htmlentities(strip_tags($h_content) )));
       if($level > $curr_level){
         $toc .= "\n".'<div class="list-group">'."\n";
+        $counter +=1;
       } else if($level == $curr_level) {
         $toc .= "\n";
       } else {
         while($level < $curr_level){
           $toc .= "\n</div>\n\n";
           $curr_level -= 1;
+          $counter -=1;
         }
       }
       $curr_level = $level;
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -118,12 +119,14 @@ function generate_toc($html_string){
       $name = trim(str_replace('&nbsp;','', htmlentities(strip_tags($h_content) )));
       if($level > $curr_level){
         $toc .= "\n".'<div class="list-group">'."\n";
+        $counter +=1;
```
ewels, line 122: I wonder if this also needs to have a `while` loop to mirror the one for closing divs?
```suggestion
        while($level < $curr_level){
          $toc .= "\n".'<div class="list-group">'."\n";
          $curr_level += 1;
          $counter +=1;
        }
```

includes/functions.php
```diff
@@ -118,12 +119,14 @@ function generate_toc($html_string){
...
       } else if($level == $curr_level) {
         $toc .= "\n";
       } else {
         while($level < $curr_level){
           $toc .= "\n</div>\n\n";
           $curr_level -= 1;
+          $counter -=1;
         }
       }
       $curr_level = $level;
```
ewels, line 132: Presumably wouldn't need this if we are changing `$curr_level` in the new `while` loop in the first `if` block..

```suggestion
```

includes/functions.php
mashehu: No. This is what I try to say with the empty list-group above. This will not have an `<a>`-tag, which would be able to have an `active` class. So we could not expand the corresponding list-group sibling in the css.

ewels: ok, so the ToC hierarchy would look like h1 h2 even though it's h1 h3 I guess?

In fairness this shouldn't really happen. I think that one of the markdownlint looks for non-consecutive headings - not sure if we have turned that off or not though.

mashehu: yes, but I don't think it will have an effect on the rendering of the ToC. 
Top-level params would be h3 without a parent h2 element.

ewels: ok even better!

PR #480, opened 2020-07-30
switch to dark scroll bars in dark mode

mashehu:
While trying to fix the ToC, I got too annoyed by the white scroll bars 🤷‍♂️
Firefox:
Screenshot 2020-07-30 at 14 16 09
Webkit:
Screenshot 2020-07-30 at 14 16 35

### DIFF
--- a/public_html/assets/css/nf-core-dark.css
+++ b/public_html/assets/css/nf-core-dark.css
@@ -490,3 +490,22 @@ hr {
   color: #999;
   background-color: #181a1b;
 }
+
+/* dark mode scroll bar */
+*::-webkit-scrollbar {
+    max-width: 10px !important;
+    max-height: 10px !important;
+    background: #1d1d1d !important;
+}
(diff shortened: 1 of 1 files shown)


ewels:
This makes the scroll bars more obvious and annoying for me (Chrome, OSX).

Before the change the scrollbars are darker and automatically disappear when not being scrolled:

before

After the change they are there all the time and are more obvious:

after

Similar behaviour in Firefox and Safari. I guess this is an operating-system thing?

mashehu:
Ah, yes, I didn't test for cases when the OS hides scrollbars. Scrap this PR then. But interesting that your chrome bars are automatically dark. Which OS version and are you using a chrome theme?

ewels:
* macOS Catalina 10.15.6 (`19G73`)
* Chrome 84.0.4147.89

No theme..

mashehu:
okay, maybe I should finally update to Catalina (or Big Sur soon) then 🙂

PR #483, opened 2020-07-30
Make pipeline_names.json

ewels:
I'm playing with making GitHub Actions parallelise jobs across all pipelines using a matrix operator. This requires a JSON list of pipeline names. To make things simple, I think it's easier to have a basic JSON file on the website. It might come in handy for other stuff too...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 
 config.ini
 public_html/pipelines.json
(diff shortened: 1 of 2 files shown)


ewels:
I've manually put the file up already for testing, so you can see it here: https://nf-co.re/pipeline_names.json

PR #486, opened 2020-08-07
fix bug for updating the name of empty groups and adding the first new top-level parameter

mashehu:
Encountered a bug, where the names of newly created (empty) groups weren't updated in the schema, because one of the validation steps fails. 
While debugging, I ran into a second bug, where you couldn't add new top-level parameters, if the schema didn't have some already (basically the schema cleanup step was a bit too aggressive). 
Fixed both and it shouldn't interfere with other steps 🤞.


ewels:
Great, thanks! Yes there was a lot of stuff like this with the structure change. Well found 👍🏻

PR #490, opened 2020-08-18
Add autosave

mashehu:
Added a post request with the current schema to all instances, where the json field on the website is updated, which should cover all big change events.

I hope you meant the `#schema-send-status`-text as the one difference between autosave and the final save. Otherwise, I just copy-pasted it into a new function. Should I add a return value and also use it when the finish button is clicked?

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1616,15 +1616,6 @@ function update_schema_html(schema){
 }
 function autosave_schema(schema){
     // autosave schema file
-    schema = clean_empty_schema_keys(schema);
-    if(schema.hasOwnProperty('definitions') && Object.keys(schema['definitions']).length == 0){
-        delete schema['definitions'];
-    }
-    if(schema.hasOwnProperty('definitions')){
-        for(k in schema['definitions']){
-            schema['definitions'][k] = clean_empty_schema_keys(schema['definitions'][k]);
-        }
-    }
     
     post_data = {
         post_content: "json_schema",
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -106,6 +106,8 @@ $(function () {
             // Update printed schema in page
             update_schema_html(schema);
 
+            autosave_schema(schema);
+
             // Update form
             $('.schema_row[data-id="'+id+'"] .param_fa_icon i').removeClass().addClass(class_name+' fa-fw');
             $('.param_fa_icon').popover('hide');
@@ -152,6 +154,8 @@ $(function () {
             "properties": { }
         };
         update_schema_html(schema);
+
+        autosave_schema(schema);
     }
 
     // Add parameter button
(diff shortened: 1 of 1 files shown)


ewels:
Yup, see: https://github.com/nf-core/tools/blob/f14c7a5692c96896c2ed1b730bc355093dca2247/nf_core/schema.py#L497

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -248,7 +254,7 @@ $(function () {
             post_data = {
                 'post_content': 'json_schema',
                 'version': 'web_builder',
-                'status': 'web_builder_edited',
+                'status': 'web_builder_finished',
```
ewels, line 257: Whilst I agree that this is more intuitive - the change will break things as the string is expected by `tools`: https://github.com/nf-core/tools/blob/3b62ee18ea8606a59e655b8eb210652726a3566b/nf_core/schema.py#L530

mashehu: You were a bit too fast (or I was too slow with adding the following PR) 😄 https://github.com/nf-core/tools/pull/731

ewels: Ah nice! Ok but then we should split this out, as we can't merge the website change until we do a release of nf-core/tools..

ewels: ..whilst this is nice, I'm not sure that it's worth the trouble? Can we not just use the existing status names? As far as I can tell it's only cosmetic for those reading the code.

mashehu: not sure what you mean with split this out. Can we have the autosave without changing nf-core/tools? 

Yes, the naming change might be just cosmetic, but it was the reason why I thought I could reuse the status 🙂

ewels: > Can we have the autosave without changing nf-core/tools?

Yes - if you revert the change in this line and use the string `waiting_for_user` on L1623 then it should work fine I think.

mashehu: ah, okay, so to basically autosave the server cache and not autosave the json file. I will prepare the PRs accordingly.

mashehu: See #491

ewels: Ah I had never even considered auto-saving all the way back to the local copy! Yes this would only be for auto-saving the cached copy on the server.

Do we need auto-saving back to the user like that? I’m not really sure what problem that solves. I think I’d almost prefer not to have it as for me it would break expectations as a user..

mashehu: I see the web-interface just as a layer on top of the file, so whatever changes in the interface should be reflected in the file, similar to a word processor (where I am following the macos idea of the hidden autosave, which scars me whenever I have to use word and forget about the missing autosave 🙃 ). Therefore I was also a bit confused about the "finished" button at the beginning 🙂. 
But from a UX-perspective the server-based approach is probably enough, because there is only a small amount of scenarios where you would need to recover the state from a file, because the cache is somehow not reachable anymore.

PR #492, opened 2020-08-20
fix unescaped quotes in input values on a rendering level

### LINKED ISSUE #489: Use of double-quotes within Schema description field

mashehu:
fixes #489

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1154,7 +1154,7 @@ function generate_param_row(id, param){
             attrs += ' max="'+param['maximum']+'"';
         }
         if (param.hasOwnProperty('default')){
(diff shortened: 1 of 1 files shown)


ewels (approved):
Nice! It's only the rendering that's an issue and not the rest of it? Happy days! 👍🏻

ewels:
While we're here, maybe worth adding a couple more and wrapping into a function?

https://stackoverflow.com/a/48226843/713980

(with added backtick:)

```javascript
function sanitize(string) {
  const map = {
      '&': '&amp;',
      '<': '&lt;',
      '>': '&gt;',
      '"': '&quot;',
      "'": '&#x27;',
      "`": '&grave;',
      "/": '&#x2F;',
  };
  const reg = /[&<>"'`/]/ig;
  return string.replace(reg, (match)=>(map[match]));
}
```

PR #494, opened 2020-08-20
Launch - check release name, use latest if not specified

ewels:
My version of https://github.com/nf-core/nf-co.re/pull/493

* Check that the release actually exists (not done previously I think?) and give nice error message if not
* Use the latest release if none specified in the URL

### DIFF
--- a/public_html/launch.php
+++ b/public_html/launch.php
@@ -51,8 +51,9 @@ $nxf_flag_schema = array(
         ]
     ]
 );
-if(isset($_GET['pipeline']) && isset($_GET['release'])){
-    $error_msgs = launch_pipeline_web($_GET['pipeline'], $_GET['release']);
+$release = isset($_GET['release']) ? $_GET['release'] : false;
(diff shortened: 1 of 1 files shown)


mashehu (approved):

mashehu:
Looks good to me. One thing we could maybe change on the way: in the /launch page, we call it revisions for the dropdown, while we call it releases everywhere else.

ewels:
Ah yes, we call it `-revision` because that's the nextflow flag. Strictly speaking we should call it revision everywhere.

The fact that the launch UI form is sort of meant to represent a command-line command is not very obvious though, and doesn't work well on mobile. I'll have a stab at rewriting it quickly now.

Phil

ewels:
Ok, simplified / redesigned a little:

image

---

Mobile:

mashehu:
Not quite there yet for long ids (on iphone 11):
IMG_2934

ewels:
Yeah, can take that in a separate PR though, laptop is closed again now ;)

PR #495, opened 2020-08-24
fix overflowing form-selects on small devices 

mashehu:


public_html/launch.php
```diff
@@ -324,8 +324,8 @@ function build_form_param($param_id, $param, $is_required){
     // Boolean input
     if($param['type'] == 'boolean'){
         $input_el = '
-        <div class="form-control pl-4">
-            <div class="form-check form-check-inline mr-4">
+        <div class="form-control pl-md-4">
+            <div class="form-check form-check-inline  mr-md-4">
```
ewels, line 328: Does this work with the version of bootstrap that we’re using?

ewels (approved):
Haven’t tested, but LGTM

public_html/launch.php
mashehu: oops, no, it doesn't. But I then actually don't think the padding/margin is necessarily needed. It mostly takes space away for narrow screens, while not really adding much visual balance in wider screens, imo.

On a related note: the radio buttons look in general a bit off to me in the current layout. What do you think about using a switch element instead?

ewels: I wondered about a switch when I was writing it, but I thought that it was slightly less explicit. I quite liked having the values that will be set written out.

When I had seen this overflow before I had assumed that I would move things onto two lines for mobile. Would that be better than trying to squish everything into a tiny space? Might be more future-proofed too, as we don’t know how long all parameters will be.

I also have a vague memory of giving the parameter names scrolling x-overflow so that they don’t overlap the form controls. Is this not the case?

ewels:
When I had seen this overflow before I had assumed that I would move things onto two lines for mobile. Would that be better than trying to squish everything into a tiny space? Might be more future-proofed too, as we don’t know how long all parameters will be.

I also have a vague memory of giving the parameter names scrolling x-overflow so that they don’t overlap the form controls. Is this not the case?

public_html/launch.php
mashehu: For the parameters I saw, "On/Off" would make equal sense instead of "True/False", that's why I think a switch would translate quite well and even be a bit more intuitive/easier to parse.

mashehu:
moving it onto two lines is what I want to achieve with this PR. I think the x-overflow didn't work or was not aggressive enough, but I can't test it now.

public_html/launch.php
ewels: Yes maybe, can we test it and see how it looks?

ewels (changes requested):
Just tested this locally and I'm not totally happy with the changes yet.

See for example a launch page, before the changes:




and after the changes:



Anything with a help button gets kicked onto a second line, and many form groups are broken.

Could this CSS change be refined a little? And to be honest I can't actually remember what the original problem was that this was fixing..

Phil

mashehu:
the PR goes back to [this comment](https://github.com/nf-core/nf-co.re/pull/494#issuecomment-677856933). 
I will have a look but maybe after #513 has been merged, which has minor layout changes for that page.

maxulysse:

ewels:
Agree, let's close. I think the launch page is due to be rewritten and could probably do with a larger scale redesign anyway.

For example, I quite like what Oxford Nanopore are doing with their layout in EPI2ME [here](https://labs.epi2me.io/quickstart/). So maybe we could do something similar..

image

PR #502, opened 2020-09-17
Dates and times correction

JoseEspinosa:
Dates were corrected since they were not correctly read.
Times also corrected to show time zone.

### DIFF
--- a/markdown/events/2020/BovReg-CRG-workshop-2020.md
+++ b/markdown/events/2020/BovReg-CRG-workshop-2020.md
@@ -3,9 +3,9 @@ title: BovReg Workshop and hackathon
 subtitle: Reproducible genomics workflows using Nextflow and nf_core organized by The Center for Genomic Regulation (CRG), Barcelona
 type: workshop + hackathon
 start_date: "2020-11-17"
-start_time: "09:30" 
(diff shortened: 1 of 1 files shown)


ewels (approved):

ewels:
Yup, as you saw - the date strings are parsed as date objects, so stuff like `(TBC)` breaks the rendering :)

ewels:
Another thing is that the event types use a controlled vocabulary - currently `hackathon`, `talk`, `poster` or `tutorial`. As a result, `workshop + hackathon` isn't styled properly. We should pick one of the four presets, or if we want we can add another (code [here](https://github.com/nf-core/nf-co.re/blob/436cd1707944cfeaaeed15f0881a5fb95654d830/public_html/events.php#L4-L15)).

Finally, it should be `nf-core` and not `nf_core` in the top page description 😉

JoseEspinosa:
Couldn't have made more typos in a shorter text 😓

ewels:
Hah, don't worry about it! We should probably add CI and stuff for this, there is no documentation or any way to know about these things without diving into the code and poking around. We haven't had so many contributors so far though so it hasn't really been an issue. But definitely not your fault 😉

Issue #499, opened 2020-09-10
json schema escaped `\`


ggabernet:
When editing the parameter json schema with the web builder, the `/` characters are escaped to `&#x2F;`. 

For example, the igenomes_base parameter default is set to `s3://ngi-igenomes`

<img width="1146" alt="Screenshot 2020-09-10 at 19 40 58" src="https://user-images.githubusercontent.com/3973503/92802030-7160fb00-f3b6-11ea-8988-bb3c3c4046f3.png">

But when opening the schema with the web builder and save it again, it is stored in the json file as: `s3:&#x2F;&#x2F;ngi-igenomes`

<img width="521" alt="Screenshot 2020-09-10 at 19 41 39" src="https://user-images.githubusercontent.com/3973503/92802163-8e95c980-f3b6-11ea-8b86-00ebd4d4846f.png">

And also displayed as: `s3:\/\/ngi-igenomes` in the json web viewer:

<img width="613" alt="Screenshot 2020-09-10 at 20 00 05" src="https://user-images.githubusercontent.com/3973503/92802270-a8cfa780-f3b6-11ea-8be9-4c584b0d74a3.png">

mashehu:
Looked quickly into this and I think it works as intended in chrome, but not on firefox 😞 

@ggabernet: which browser were you using?

ggabernet:
I'm using chrome usually. I can test again sometime this evening but 99% sure I did that with chrome.

ggabernet:
I've just tested it again and it happened again with chrome: 
`nf-core version 1.10.2` 
`google chrome version 85.0.4183.83` for MacOS

ggabernet:
@Leon-Bichmann had the same issue. Which browser are you using? And version?

mashehu:
okay, if you are also using chrome, my theory is not valid and I have to look deeper into it (my version is 85.0.4183.102, but that shouldn't make a big difference).  

I could now partially reproduce the bug. For some reason it works for me in the case of `igenomes_base`, which is why I thought it works in chrome. But I now found several other instances of printed html character codes in the json file.

mashehu:
Made a hot-fix (#503) for the issue, but it is by far not an optimal solution.

The basic problem is: We added a sanitize function for strings, to escape special characters in #492 to fix #489. Unfortunately, the html entities are directly written into the schema object and never decoded in a later step (I assumed `JSON.parse()` would do this).

ggabernet:
awesome, thanks!

[...]

PR #503, opened 2020-09-21
fix html encoding artefacts in schema

mashehu:
by decoding the affected strings before writing the object

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1576,7 +1576,8 @@ function find_param_group(id){
 function update_param_in_schema(id, new_param){
     // Given an ID, find the param schema even if it's in a group
     // Assumes max one level of nesting and unique IDs everywhere
-
+    new_param.description = reverse_sanitize_html(new_param.description)
+    new_param.default = reverse_sanitize_html(new_param.default);
     // Simple case - not in a group
     if(schema.hasOwnProperty('properties') && schema['properties'].hasOwnProperty(id)){
         schema['properties'][id] = new_param;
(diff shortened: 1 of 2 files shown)


mashehu:
This would be one quick fix for #499. But it is by far not optimal, since already created schemas values would have to manually be touched once, to be re-encoded. Should we maybe fix this on the tools side in the build function, @ewels ?

ewels (approved):
Ok nice, so the whole issue is due to the `sanitize_html()` function? Then yes I think that this looks like a good fix. Haven't tested it, but assuming that you have I'm happy for this to be merged to at least avoid any future damage.

I think other fixes in tools etc can be in addition to this fix? Thoughts?

mashehu:
Having thought more about it, maybe the tools part is not needed. We don't have so many schemas out in the wild yet, so a more manual fix should be doable. Either I go manually through them and look for incorrect encodings (`/` being the main culprit) and/or write a linter function for it.

But I will test this draft a bit more (especially the json preview) and will let you now, when I need a review.

ewels:
Sounds good! 🚀

ewels:
The issue reported two different kinds of encoding weirdness - the HTML entities, but also the `\/` escaping.. Is the latter a read herring, and just part of regular JSON?

mashehu:
Yes, the second one is what I mentioned with the json preview. I figured it out: php's `json_encode()` is a bit too clever, but thankfully has an `JSON_UNESCAPED_SLASHES` option.

ewels:
Nice! 🚀

mashehu:
Okay, I had a deeper look into the problem: I was over-sanitizing the strings and without realizing that javascript+jQuery would already take care of the special characters in places, where we don't write direct html.

So, this PR should actually fix both problems mentioned in #499 and should automatically fix the incorrect json schemas if loaded once into the browser (because we write it into the text area in the php part => HTML entities are converted into the symbols => the javascript reads the text and doesn't reintroduce the html entities => win 🙂). But this part I can't competly test, unfortunately.

ewels (approved):
LGTM! 🚀

PR #506, opened 2020-09-23
add additional check between param ids and group ids

### LINKED ISSUE #500: Issues with same IDs between parameters and parameter groups in JSON schema
Hi!

As discussed on [Slack](https://nfcore.slack.com/archives/CUC8PPDL2/p1597768672037000) if you use the same ID for a group and a single param, you get issues like not being able to configure advanced options of the param, or the param taking over the description of the group.

This should probably be checked during linting or somewhere else in advance.

mashehu:
fixes #500


ewels (approved):
LGTM, can merge after fixing the conflict 👍🏻

PR #513, opened 2020-10-01
launch page fixes/improvements

mashehu:
- Fix incorrect pipeline command, see [slack](https://nfcore.slack.com/archives/CE7DN1U7M/p1601294871003100) for issue report.
- Move ToC from dropdown menu to the side (unfortunately I couldn't reuse `generate_toc()` due to the differences in html structure, but I simple reused the code from the dropdown menu (so only group names appear there, which is enough for me)).
- fix incorrect id for definition group fieldsets.

While fixing this I tried the using switches instead of the radio buttons for booleans, as mentioned in #495. It looks nicer imo:
Screenshot 2020-10-01 at 15 26 06
The problem: A switch only returns a value if it is set to "on", so if we have some parameter true by default and a user switches it off, we wouldn't be parsing that. So, I had to move back to the old radio buttons 😞.

### DIFF
--- a/public_html/assets/css/nf-core.css
+++ b/public_html/assets/css/nf-core.css
@@ -1097,7 +1097,9 @@ Based on https://codepen.io/wintr/pen/beBJBb */
     max-height: calc(100vh - 9.1rem);
   }
 }
-
+.text-smaller{
+  font-size: 90%;
+}
 .toc > .list-group > .list-group-item {
   font-weight: bold;
 }
--- a/public_html/launch.php
+++ b/public_html/launch.php
@@ -540,66 +540,67 @@ else if($cache['status'] == 'launch_params_complete') {
         }
     }
     ?>
+<div class="container main-content pt-5">
+    <h1>Launch parameters saved</h1>
 
-<h1>Launch parameters saved</h1>
+    <?php if(isset($cache['cli_launch']) && $cache['cli_launch']): ?>
+    <p>The <code>nf-core launch</code> command running in your terminal should have automatically detected your settings.
+        Follow the prompts in your command line to launch the pipeline.</p>
+    <p>If the launch command has stopped running for any reason, you can still launch by following the instructions below:</p>
+    <?php else: ?>
+    <p>Your workflow parameters are ready to go! Follow the instructions below for instructions on how to launch your pipeline:</p>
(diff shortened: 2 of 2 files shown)


mashehu:
The diff is very confused because I moved things some things around. My new code/changes are basically on line 129, 725, 767, adding the sidebar layouting in line 636-637 and adding the toc on line 741, and 776-782 and replacing (the appending the former echo command to `$toc_list` in line 767 and 771)

ewels (changes requested):
Ok I don't really understand the diff even with your line numbers 😅 (links are better), but tested and it's looking great!

A few requests:

* `container-xl` class for the body to make use of some additional width now that we have an extra page element.
* The usage docs ToC has a CSS rule under `.toc .list-group .list-group` for 90% font-size, which is missing here and is needed.
* The launch page was still missing the `nf-core/` prefix when I tested with `hic`: 
* Also the same for the `nf-core launch` command: 
image
* Minor: the "back to top" link does a hard jump instead of scrolling like the other ToC links. Could be nice if it shared the same behaviour...

Looking good though generally! The sidebar ToC is a major improvement 👍🏻

mashehu:
added the `container-xl`, but I had to do it in a bit clumsy way, because for some reason `$mainpage_container` (which adds a `container main-content` element at the beginning of the page) isn't validated when switching between pipeline selection form and the form with the launch parameters. So I manually add the `$mainpage_container` element to the page before and after the launch form. There has to be a better way for this...

Are you sure you used the correct website when taking the screenshot (and it is the same string used on the page and the terminal so if one is incorrect the other one has to be as well):
Screenshot 2020-10-07 at 11 21 23

I added the smooth scrolling and it scrolls now back to the top of the form instead of the top of the page. Should we add the smooth scrolling also to the usage and output page ToCs?

ewels:
> Should we add the smooth scrolling also to the usage and output page ToCs?

Yes probably, but I'm happy with this PR so will merge now 😉 

Launch command looking fine now, so probably yes it was me messing up which PR I had active when I tested that.

Thanks! 🚀

ewels (approved):

mashehu:
the smooth scrolling was already added to docs with ToCs with #531 🙂

PR #514, opened 2020-10-02
AWS megatests outputs page

### LINKED ISSUE #435: Show #awsmegatests results on pipeline page
See https://github.com/nf-core/awsmegatests/issues/15

We will soon start having the results from the full-size test datasets stored on s3. It would be great to make these visible through the nf-core website as easily as possible.

Need a few things here:

* [x] Bucket setup / permissions for access, preferably without any auth at all
* [x] Standardised file structure on s3 so that we can predict / find results
* [ ] Server side / client side code to get file listing
* [ ] Server side / client side code to fetch files and render in the browser?

Note that we could potentially dodge the last two by just cheating and using an iframe to the s3 browser? Not sure how this would work.

mashehu:
Basically copy-pasted from [aws-js-s3-explorer](https://github.com/awslabs/aws-js-s3-explorer). Files are currently downloaded when clicking their name in the table and not yet rendered on page, as was suggested. 
I am actually not 100% sure, how to handle that. Should we basically have a download for non-html files and html files will be rendered on that page (with the file table above or below)?

We also need to figure out a way to handle the dev branch, since it doesn't appear in `pipeline.json` yet, so we don't have the commit-hash for it (which is needed for the path in the aws bucket). Similarly, for versions without a aws results, the "example output" tab should be hidden.

Closes #435.


ewels:
Absolutely love it 🚀 

For dev, could maybe show a list of unrecognised commits? Or even just query the GitHub API to find the commit for the latest dev? Not sure what would be best.

Rendering HTML / txt files etc would definitely be nice. May be tricky as I guess they are downloading due to the header that is being set from amazon though. Not sure how we can change / bypass that?

Minor stuff:

* Missing path separators in the top path: image  (needs some CSS `:after` magic for the breadcrumbs I guess)
* Full-width page might be nicer?
* When I load a new pipeline the breadcrumbs show that I'm in the folder for the results, but the browser shows that top-level directory that I have to click to see inside
* Would be nice to hide the top-level directories above that release perhaps, so that you can't (easily) browse off to other pipelines etc or lose your place if you click too high in the breadcrumbs

ewels:
Also, I have a feeling that in some cases files are being shown even when they should be hidden within a folder? eg. digging around various `viralrecon` runs.

There are also a crap-tonne of browser console logs (expected if still dev code I guess), plus a loading spinner would be quite nice when it's having a proper thing about stuff 👀 

Still totally love how well this works - is pretty much exactly what I was hoping for 😍

ewels:
Added a couple of very minor tweaks just so that I can take some screengrabs for a talk 😉

ggabernet:
Great work, the viewer looks amazing! 😍 🚀 

I just have some minor things that I saw when testing it with the dark website mode:
- when using the dark view mode of the website the html background is also dark which makes it difficult to read.
- In the dark theme the AWS full path at the top is difficult to read because of the light colour.

It might be nice to have the possibility of downloading the whole results folder in addition to individual files, if that is possible.

Alright, now just the example outputs need to be in the bucket! I'll try to go over the most recent releases of the pipelines and execute the runs in AWS with the release commit so that at least the "small" tests results are there as a showcase.

public_html/assets/css/nf-core-dark.css
```diff
@@ -498,3 +498,17 @@ hr {
...
+}
+.breadcrumb-item.active{
+  color: #e5e6e7;
+}
+.breadcrumb-item+.breadcrumb-item::before{
+  color: #e5e6e7;
+}
+#file-preview iframe{
+  background-color: #e5e6e7;
+}
```
ewels, line 514: I think maybe giving this a max-height with scroll is a good idea - I tried it on a CSV file and it was mega long, so getting to the page footer was going to take forever. Scrolling within the page feels a bit more manageable..

```suggestion
}
#file-preview .card-body {
  max-height: 30rem;
  overflow-y: auto;
}
```

public_html/assets/js/aws-s3-explorer.js
```diff
@@ -0,0 +1,664 @@
...
+                    if(extension !== '.html'){
+                        data = '<pre><code>'+data+'</code></pre>';
+                    }
+                    else {
+                        data = '<iframe srcdoc="'+sanitize_html(data)+'" style="border:none; width:100%; height:1000px;"></iframe>';
+                    }
+                    var btn_group = `<div class="btn-group" role="group">
+                      <button type="button" class="btn btn-outline-secondary copy-url" data-target=${url}>Copy URL</button>
+                      <button type="button" class="btn btn-outline-secondary copy-url" data-target=${s3_url}>Copy S3 URL</button>
+                    </div>`;
```
ewels, line 132: Additional button to just open in a new tab would be nice:

```suggestion
                    var btn_group = `<div class="btn-group" role="group">
                      <a class="btn btn-outline-secondary" href="${url}" target="_blank">Open in new tab</a>
                      <button type="button" class="btn btn-outline-secondary copy-url" data-target=${url}>Copy URL</button>
                      <button type="button" class="btn btn-outline-secondary copy-url" data-target=${s3_url}>Copy S3 URL</button>
                    </div>`;
```

ewels (commented):
Just had another look at this and I absolutely love it 😍 The new file icons are an absolute winner..

The one thing I find a bit unintuitive still is that clicking a file opens it in a new tab to download. I expect it to open folders, but the download surprises me every time.

What do you think about a file click opening the preview box underneath for everything? Then we could remove the "Preview file" buttons and just have the box preview if it meets criteria. If it doesn't, we can get the preview box to say "Binary file, 20kb. Click here to download" or whatever. I'm thinking kind of like how GitHub does it:

image

This then requires two clicks to download the file, which I think is appropriate given that some of the files are huge and that most people won't actually want to download files most of the time.

includes/header.php
```diff
@@ -76,6 +76,16 @@
     <script src="/assets/js/showdown.min.js"></script>
     <script src="/assets/js/nf-core-schema-launcher.js?c=<?php echo $git_sha; ?>"></script>
     <?php endif;
+    if(isset($aws) && $aws): ?>
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.12.0/css/all.css">
```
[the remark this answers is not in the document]
mashehu: oops, sorry copy-paste mistake 🙃

includes/header.php
```diff
@@ -76,6 +76,16 @@
...
+    <link rel="stylesheet" href="https://cdn.datatables.net/1.10.22/css/dataTables.bootstrap4.min.css">
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/bootbox.js/5.4.0/bootbox.min.js"></script>
+    <script src="https://sdk.amazonaws.com/js/aws-sdk-2.765.0.min.js"></script>
+    <script src="https://cdn.datatables.net/1.10.22/js/jquery.dataTables.min.js"></script>
+    <script src="https://cdn.datatables.net/1.10.22/js/dataTables.bootstrap4.min.js"></script>
+    <script src="/assets/js/aws-s3-explorer.js?c=<?php echo $git_sha; ?>"></script>
```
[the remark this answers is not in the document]
mashehu: I just used the CDN's out of laziness/habit. I will switch to local deps for consistency.

Yes, package managers can be nice, but can also turn into a big time sink. Do you already have something in mind, like composer for php stuff and npm for the frontend? Maybe we could combine that switch with some code preprocessing (sass 😍)?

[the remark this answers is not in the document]
ewels: Yes, I think those two probably make sense. I'm not sure that we have enough PHP dependencies to make composer worthwhile? But if you think that would be worth it then I'm fine with that. I think npm would be a smart move. And sass / scss could presumably be bundled into that switch.

Should probably move this into a new issue ➡️

includes/header.php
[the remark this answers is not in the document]
ewels:
```suggestion
```

mashehu:
After testing it, I am not so sure about the "Open in new tab" button, because it directly starts a download for html and image files (on firefox) when clicking on it and only opens text files in a new tab.
And since we have now a "download file" button in every file row, maybe we should can remove it from there and rename the "Open in new tab" button into download file button. 

I couldn't figure out the URL to download a whole directory, so I just added a copy button for the s3 url of it (the button label "Copy Bucket S3 URL" needs some rephrasing I think 🙂)

ewels:
Ok nice - yes I agree, switch the text for the new tab button for a download button (as that's what it does most of the time) and remove the download button from the table 👍🏻 

Can we also add some additional checks to prevent previewing the file in the browser in some cases? eg. very large files and certain file extensions (`.xml`). On the mhcquant page I nearly crashed my browser by clicking on a 6MB XML file that was loaded into the DOM (didn't show anything except a very very long empty `<pre>` element with lots of invisible content).

Checked with `.bam` and `.bai` files and it showed _No preview available for binary or compressed files._ very nicely 🙌🏻  

Only other request that I can think of is making the fontawesome file icon part of the link. I keep clicking on them and nothing happens 😉

ewels:
Something a bit odd happens with the iframe in some cases too - eg. for Sarek `nf-core-awsmegatests/sarek/results-bce378e09de25bb26c388b917f93f84806d3ba27/Reports/1234N_1234N_M1/bamQC/1234N_1234N_M1/qualimapReport.html`

This seems to use `localhost` as the relative link for the CSS and images etc, instead of the s3 URL? It works fine in its own tab: http://nf-core-awsmegatests.s3-eu-west-1.amazonaws.com/sarek/results-bce378e09de25bb26c388b917f93f84806d3ba27/Reports/1234N_1234N_M1/bamQC/1234N_1234N_M1/qualimapReport.html

I don't really understand this, if the iframes are simple in the way that I've used them before..

ewels:
Just poking a bit and seems to be using `<iframe srcdoc="">` with inline file contents. Any reason to do this instead of just `<iframe src="">` with the URL?

mashehu:
I used `<iframe srcdoc="">` because some files for example the `execution_*.html` inside the `pipeline_info` directory, don't have the correct content-type, when requested. So a `<iframe src="">` just shows a white page and starts a file download instead.

ewels:
Right, gotcha - yes and it works beautifully to get around the HTML headers. So we're stuck between a rock and a hard place - either download some HTML files or break the rendering of others.

I can see the difference between these two HTML files in their metadata on s3:

<table>
<thead><tr><th>

`execution_report.html` (triggers download)

</th><th>

`qualimapReport.html` (renders in browser)

</th>
</tr>
</thead>
<tbody>
<tr>
<td>

image

</td><td>

image

</td>
</tr>
</tbody>
</table>

@pditommaso - do you have any idea why files with the same filetype would be uploaded to s3 by Nextflow with different `Content-type` headers?

As the in-browser rendering is working great for all other filetypes, we could simply have a cron-job that loops over all HTML files in the s3 bucket and fixes the content-type headers. Then we can use regular iframes and have the reports load properly. What do you think @mashehu?

ewels:
Confirmed - if you edit the content-header to `text/html`, the execution HTML reports view in browser (assuming that your browser doesn't have a cache of them).

mashehu:
Yes, I think the majority should load fine by using `src=`. So whatever solution you find for the AWS side of this is fine for me. 🙂 

I was playing around on how to do the handle the `dev` version. I couldn't figure out a way to find the most recent `results-` folder in a bucket on AWS (it looks like only files have a date and the directory structures are a bit too complicated/inconsistent for me to sort them based on that). So now `dev` shows all available result directories for a pipeline. 

I fixed the URL based rendering, so now we can deep link into a specific AWS directory, e.g. http://localhost:8888/mhcquant/example_output#sarek/results-bce378e09de25bb26c388b917f93f84806d3ba27/Reports/1234N_1234N_M1/bamQC/1234N_1234N_M1/

Oh and the xml file rendering works now as well (the large xml files killed your browser because it was trying to render it as an html file 🙄 ). 
I set a file size limit of 10MB, because some multiqc_reports can be quite large and I think they are one of the more important things to have a preview for. But I am happy to lower the limit, if you think this is not necessary .

ewels:
The other thing that we could do to avoid messing around with AWS content headers is to tunnel HTML files through a PHP script on the server. I've done this in other cases to strip annoying `Access-Control-Allow-Origin` headers that prevent files being loaded into the DOM.

Basically, just have a PHP file with the following:

```php
<?php
header("Content-type: text/html; charset=utf-8");
echo file_get_contents($_GET['url']);
```

Then load that in the iframe instead. Might be easier than messing around with cron jobs and AWS authentication etc.

pditommaso:
>  do you have any idea why files with the same filetype would be uploaded to s3 by Nextflow with different Content-type headers?

Because NF is not aware of support for MIME provided by S3

mashehu:
> The other thing that we could do to avoid messing around with AWS content headers is to tunnel HTML files through a PHP script on the server. I've done this in other cases to strip annoying `Access-Control-Allow-Origin` headers that prevent files being loaded into the DOM.
> 
> Basically, just have a PHP file with the following:
> 
> ```
> <?php
> header("Content-type: text/html; charset=utf-8");
> echo file_get_contents($_GET['url']);
> ```
> 
> Then load that in the iframe instead. Might be easier than messing around with cron jobs and AWS authentication etc.

The problem with this solution is that we generate the iframe dynamically with javascript, so I am not sure how to do this then...

ewels:
> @pditommaso - Because NF is not aware of support for MIME provided by S3

So I guess there is some kind of AWS magic somewhere that is setting these automatically? They are there and they vary across different HTML files, so something is doing it 🤔 

> @mashehu - The problem with this solution is that we generate the iframe dynamically with javascript, so I am not sure how to do this then...

Just set the `iframe` `src` attribute (not `srcdoc`) to `'/s3/'+url` so that the URL ends up being something like:

```
https://nf-co.re/s3?url=http://nf-core-awsmegatests.s3-eu-west-1.amazonaws.com/sarek/results-bce378e09de25bb26c388b917f93f84806d3ba27/Reports/1234N_1234N_M1/bamQC/1234N_1234N_M1/qualimapReport.html
```

Can start being more clever with nicer URLs and url / file extension / file size validation etc, but that's the core concept. Should work fine with the iframe with dynamic content I think.

mashehu:
I think we reach the limits of my PHP-skills here. 🙂
Where should I add handlers in the php code for these special URLs? Can you maybe point me to examples of what you are thinking of.

ewels:
Ok I had a quick play with the redirection thing and whilst it works (headers are stripped and the files display in the browser properly), the web pages with relative links still break in the same way as they do with the iframe `srcdoc`. So let's just stick with that. I'm thinking that there shouldn't be so many HTML reports that do this anyway - hopefully just Qualimap. If it's a big issue then we can always come back to it.

ewels:
@mashehu - are you happy to revert the changes back to `srcdoc` then? Sorry about this. Then I think that we're good to merge 👍🏻

includes/pipeline_page/example_output.php
```diff
@@ -0,0 +1,95 @@
...
+                        </ul>
+                        <button type="button" class="btn btn-outline-secondary copy-url ml-auto" data-target="">Copy Bucket S3 URL</button>
+                    </div>
+                    <!-- Controls -->
+                    <div id="navbuttons" class="">
+                        <div>
+                            <!-- Dual purpose: progress spinner and refresh button, plus object count -->
+                            <div class="btn-group" id="refresh">
+                                <span id="bucket-loader" class="btn fa fa-refresh fa-2x " title="Refresh"></span>
+                                <!-- <span id="badgecount" class="badge pull-right" title="Object count">42</span> -->
```
ewels, line 34:
```suggestion
```

includes/pipeline_page/example_output.php
```diff
@@ -0,0 +1,95 @@
...
+  <div class="toast-header">
+    <img src="/assets/img/logo/nf-core-logo-square.png" class="rounded mr-2" alt="">
+    <strong class="mr-auto">URL copied to clipboard!</strong>
+    <button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+  <!-- <div class="toast-body">
+    Paste this command into your terminal to run the pipeline with a small test dataset.
+  </div> -->
```
ewels, line 70:
```suggestion
```

public_html/assets/css/nf-core-dark.css
```diff
@@ -498,3 +503,20 @@ hr {
   color: #999;
   background-color: #181a1b;
 }
+.breadcrumb {
+  background-color: #616161;
+  /* border-color: #616161; */
+  /* color: #ededed; */
```
ewels, line 509:
```suggestion
```

ewels (approved):
Just a couple of comments left which I think can be deleted, but then LGTM for merge - amazing work 🙌🏻 

Ongoing discussing on Slack `#aws-megatests` about fixing the content-type headers for all HTML files, so may revisit this in the future if we can get that to work. But I'd rather merge this now as it is basically working and come back to that if anything happens.

ewels:
Hah, found my first bug as soon as I merged: `Copy Bucket s3 URL` button doesn't seem to do anything:

image

Shows a toast saying it worked, but nothing changed in my clipboard..

ewels:
Other copy buttons for specific files are working..

apeltzer:
It also doesn't seem to show anything for e.g. atacseq, eager, ...

ggabernet:
@apeltzer, unfortunately the problem of the container not being built before the release AWS test is triggered is persisting and did not find a good solution for that.

However, I've manually re-triggered the workflow again for the latest release commit of the most recently released pipelines (until two months ago). The tests were triggered again for these. In the case of `eager`, it failed and I made a PR to hopefully fix that: https://github.com/nf-core/eager/pull/600.

`Proteomicslfq` and `epitopeprediction` also fail on AWS, so the output can't be showed.

I've triggered with Tower `chipseq` which was passing in the past and it's running at the moment, so hopefully this solves it for this pipeline as well.

ggabernet:
Hi @ewels @mashehu, I'll look into triggering a Lambda function now on all the `html` files. What exactly in the `html` needs to be converted to what?

This is one example of `html` file only displayed when clicking `download`.

Is it this line? And exactly which tag?
```
<html><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
```

mashehu:
Some have a content-type `octet-stream`, but they should all be `text/html`. See https://github.com/nf-core/nf-co.re/pull/514#issuecomment-718760044

ggabernet:
Hi, sorry I still don't get exactly the issue. For me, the `execution_report.html` of your example, which I guess would be [this one](https://nf-co.re/sarek/2.6/example_output#sarek/results-75fe254f0b5d93d1f3f0173f2163afa83686f6fd/pipeline_info/), renders fine in the browser, even though the metadata is `Content-type: application/octet-stream`.

ggabernet:
Ah gotcha, it triggers a download in a new window

ggabernet:
Unfortunately after reading about it, it does not seem to be so straight-forward:

```
Each Amazon S3 object has data, a key, and metadata. The object key (or key name) uniquely identifies the object in a bucket. Object metadata is a set of name-value pairs. You can set object metadata at the time you upload it. After you upload the object, you cannot modify object metadata. The only way to modify object metadata is to make a copy of the object and set the metadata.
```

Issue #512, opened 2020-09-29
Broken Links to design files


alaincoletta:
https://nf-co.re/atacseq/1.2.1/assets/design_pe.csv
and 
https://nf-co.re/atacseq/1.2.1/assets/design_se.csv
return a 404. They referenced at the end of this section: 
https://nf-co.re/atacseq/1.2.1/usage#--input

https://nf-co.re/chipseq/1.2.1/assets/design_pe.csv
and
https://nf-co.re/chipseq/1.2.1/assets/design_se.csv
return a 404. They referenced at the end of this section: 

referenced at the end of this section: 
https://nf-co.re/chipseq/1.2.1/usage#--input

apeltzer:
Probably it would be safest to use Fully Qualified URLs here @drpatelh - that should resolve these issues?

drpatelh:
Sorry, I missed this. Yup I suspect it's because of the recent changes to the website. Which branch would we point to in the full URLs? With relative URLs that doesn't matter. Or am I missing something? Ping @ewels @mashehu

apeltzer:
Relative has the benefit that it stays up2date no matter when there is a new release. If we remove links like this in an upcoming release (and thus subsequently from `master`) a previous link to master will be gone too - so maybe we should link to a specific tag instead? Not sure how to achieve this though...

mashehu:
We have a full link to the master branch for the igenome confs, as resolved here: https://github.com/nf-core/tools/pull/691/files. Since the docs on the website are available for different version, it might actually be good to reparse these kind of links and replace `blob/master` with `blob/$version`.

drpatelh:
Yes, I think we should somehow point/redirect to the version of a file stored under a particular tag. Maybe replace both `blob/master` and `blob/dev` with `blob/$version` for any release tagged docs on the website? There will be instances where certain files are added to `dev` first and so may not be on `master` until release - at which point the pipeline author may forget to update `dev` -> `master` in the links.

I guess the problem with relative URLs is that the file can be deleted in a new release but the link from older releases will still point to `master` where it no longer exists.

drpatelh:
Also, is this something we can lint for? I think it is something that can be easily missed and will only be updated on new release 🤔

mashehu:
you mean linting for relative urls or for occurences of `/blob/dev`?

ewels:
I think that it would be better to just make the relative URLs work. I'm going to have a quick play.

ewels:
[...]

PR #517, opened 2020-10-10
If a 404 GitHub suggestion URL exists, redirect there

ewels:
Since we rewrote how documentation works on the website, arbitrary relative links that work on GitHub no longer work on the website. We wrote some logic in to guess GitHub URLs and show these on the 404 error page, but it wasn't very sophisticated.

This PR adds to that logic a little by (a) checking whether the suggested replacement URLs exist and (b) if they _do_ exist, automatically redirecting there directly.

As such, it essentially fixes relative links in documentation, and should solve #512

### DIFF
--- a/includes/pipeline_page/_index.php
+++ b/includes/pipeline_page/_index.php
@@ -134,6 +134,7 @@ else if($_GET['path'] != $pipeline->name && $_GET['path'] != $pipeline->name.'/'
   header('HTTP/1.1 404 Not Found');
   $suggestion_404_urls = [
     $protocol.$_SERVER['HTTP_HOST'].'/'.$pipeline->name,
+    'https://github.com/nf-core/'.$pipeline->name.'/blob/'.$release.'/'.$url_string,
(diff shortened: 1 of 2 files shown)


drpatelh (approved):
Thanks @ewels Will merge and test properly.

MaxUlysse:
What happens if links suggested do not exist?
Should we suggest them to begin with?
Should we suggest the main readme or documentation?

ewels:
If the suggested link does not exist then it is not suggested - this was added in this PR: https://github.com/nf-core/nf-co.re/blob/4b4fbbc69749e9fb390eb3c091747e74ea1b06ec/public_html/404.php#L7-L8

The auto-redirection only happens if the suggested link is on GitHub. Others are just checked and shown.

404s from https://nf-co.re/<pipeline>/<anything> always suggest https://nf-co.re/<pipeline> - basically suggesting the main documentation 👍🏻 

You can try it out: https://nf-co.re/sarek/sdfjh

PR #525, opened 2020-10-14
change to new tower launch syntax

### LINKED ISSUE #524: Tower Launch - use new field names
Nextflow Tower has changed some of the field names, so the submission button needs updating - see https://github.com/seqeralabs/nf-tower-cloud/issues/314#issuecomment-708336778

A new `paramsText` field also means that we can submit JSON directly, so can simplify the code a bit as we already have JSON (can kill [this code](https://github.com/nf-core/nf-co.re/blob/1bd3302f3af2171ea7e0549d525d3fe3e4414bd5/public_html/launch.php#L530-L540)).

mashehu:
fixes #524 

I tested it as far as I could (I don't have a compute environment I can kick of from tower (yet)) = the launch pipeline page displays the "input_params" as a correct json in the "Pipelines parameters" field in my test cases.

### DIFF
--- a/public_html/launch.php
+++ b/public_html/launch.php
@@ -515,7 +515,7 @@ else if($cache['status'] == 'launch_params_complete') {
     $tower_fields = array(
         "pipeline" => "https://github.com/".$cache['pipeline'],
         "revision" => $cache['revision'],
-        "configParams" => "",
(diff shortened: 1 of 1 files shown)


pditommaso:
This looks fine to me

ewels (approved):
LGTM 👍🏻

PR #537, opened 2020-10-23
Show default values

### LINKED ISSUE #528: Render default values for parameters obtained via the JSON schema
Would be nice to see what the default values for each parameter wherever they are rendered on the website too. These values are already in the JSON file and are used with the launch functionality already.

mashehu:
Closes #528.


includes/pipeline_page/docs_schema.php
```diff
@@ -58,6 +58,13 @@ function print_param($is_group, $param_id, $param, $is_required=false){
           '.parse_md($param['help_text'])['content'].'</div>';
     }
 
+    # default value 
+    $default_val = '';
+    if(array_key_exists("default", $param) && strlen(trim($param['default'])) > 0){
+      $default_val = is_string($param['default'])? "'".$param['default']."'":$param['default'];
+      $default_val = '<code class=""><span class="text-muted">default:</span>'.$default_val.'</code>';
```
ewels, line 65: Do we need the span? Can the text-muted not go in the `<code>` tag?

mashehu: If I add `.text-muted` to the whole code tag, the default value itself will also be muted. I think it helps to parse the docs faster, if the default value is highlighted a bit.

ewels: Ah of course 👍🏻 

Ok, then why the empty `class=""` on the `<code>`?

ewels (approved):
Just checked locally and I think it's great 👍🏻 

Two minor requests:

* Drop the font-size a little, maybe `0.8rem`?
* Increase the colour contrast a little with the light theme (looks good in dark)

mashehu:
The contrast between which elements do you mean? `default:` vs. the default value?

ewels:
Yup 👍🏻  I just think the greys are a little close. The grey / orange in dark mode looks fab.. Maybe we could have grey / green or something for light mode?

ewels:
This is me being very pedantic of course - I almost certainly wouldn't have noticed / said anything if I hadn't noticed the dark mode formatting and thought it nice 😅 

Also maybe a space between the `default:` and the value? And I get why you did the quotes, but I'm not sure that they are really necessary.. I'm not sure that they really add much and they add to the visual noise.

mashehu:
How about if we switch the `code` color for both to green, which gives the dark mode the classic matrix color scheme 😎:
Screenshot 2020-10-30 at 12 16 21Screenshot 2020-10-30 at 12 16 43

I like the quote for strings, because it implicitly shows the type of the value, for example that the `2:3` above is a string. But maybe I am overthinking here and it actually doesn't need this kind of clarification.

jfy133:
> How about if we switch the `code` color for both to green, which gives the dark mode the classic matrix color scheme :
> Screenshot 2020-10-30 at 12 16 21Screenshot 2020-10-30 at 12 16 43
> 
> I like the quote for strings, because it implicitly shows the type of the value, for example that the `2:3` above is a string. But maybe I am overthinking here and it actually doesn't need this kind of clarification.

:100: for quotes for strings are important - because I still have people forgetting (when running on the CLI) to put quotes around paths which messes up a lot of stuff with bash expansion etc. I think it gives it a nice demarkation of: numbers fine without, anything with 'text' must be in quotes. 

Also I'm all over the matrix colouring ;)

ewels:
Love it 😍 

Happy to be outvoted on the quotes 👍🏻

PR #544, opened 2020-10-29
Param docs tweaks

ewels:
Formatting tweaks for the params docs.

* Dropped the `--` before parameter names and made them a normal font instead of monospace / code
* Made the parameter name bold and darker, made the description text lighter
* Moved labels (`required`, `hidden`) to be before the help buttons, so that they align
* Tweak the background colours for the help text

Before:

image

After:

image


mashehu:
I played around with it a bit:
- The parameter name font is a bit too heavy for my taste. Maybe change `font-size` instead?
- I liked that the parameter names had a different background so it was easier to recognize, when a parameter name was actually overflowing:
Screenshot 2020-10-30 at 12 42 31
Screenshot 2020-10-30 at 12 31 21

ewels:
Yes, I wasn't totally sure about the bold font either. I actually tried going the other way and having `font-weight: 100` and nearly went for that, with a bigger font size. Maybe see what you think of that instead.

Good point with the overflow. What do you think about using `text-overflow: ellipsis` and then making it show (with an opaque background colour) on text hover? Basically the same as this w3schools demo: https://www.w3schools.com/cssref/tryit.asp?filename=trycss3_text-overflow_hover Might be clearer and easier than scrolling to see the full value.

ewels:
NB: Mobile layouts need some love too:

image

I'm starting to wonder if I should be annoying and go totally back on my earlier self where I fought for a compact layout. I'm now wondering if we should space all of this our way more and make it less tabular. Show all help all the time, use parameter names as headings, etc. 🤔

mashehu:
The problem I see with spacing it out more is that description and help text are not so long for many parameters, so we create just a lot of empty space. But we can give it a try.

I thought a bit more about consistency and I think we should leave the parameter names inside a `<code>` tag, because this is also how they are often rendered in the output docs for example.

ewels:
Yeah, good point about the consistency. I'll have a play with this a bit more next week and see where I end up.

maxulysse:

FranBonath:
Just to be annoying here, I would leave the double dashes before. Without it, newbies may be confused on how to use them correctly.

includes/pipeline_page/docs_schema.php
```diff
@@ -124,7 +120,7 @@ function print_param($is_group, $param_id, $param, $is_required=false){
         if(array_key_exists("required", $group) && in_array($param_id, $group["required"])){
           $is_required = true;
         }
-        $schema_content .= print_param(false, '--'.$param_id, $param, $is_required);
+        $schema_content .= print_param(false, $param_id, $param, $is_required);
```
FranBonath, line 123: I would leave these

includes/pipeline_page/docs_schema.php
```diff
@@ -135,7 +131,7 @@ function print_param($is_group, $param_id, $param, $is_required=false){
       if(array_key_exists("required", $schema) && in_array($param_id, $schema["required"])){
         $is_required = true;
       }
-      $schema_content .= print_param(false, '--'.$param_id, $param, $is_required);
+      $schema_content .= print_param(false, $param_id, $param, $is_required);
```
FranBonath, line 134: and also these

FranBonath (commented):
except for that, I have no objections and I guess the double dashes is also a matter of taste

ewels:
Closing as code is outdated.

Some design points still valid, but I think the new website changes the design enough that this should be considered at a later date with a fresh take on any redesign.

PR #547, opened 2020-10-30
fix param docs ToC and launch page

### LINKED ISSUE #545: Parameter docs ToC - show hidden
1. When Parameter docs were moved into their own tabs, they left their special ToC buttons behind at `/usage`:
    
    These need to be moved over to `/parameters`.

2. Whilst we're there, it could be good to mimic this UI in the Launch page too, for consistency. This also makes the top bar even cleaner (assuming that we remove the "Show hidden params" button from there), which I think is no bad thing.

3. Final ToC point - at the moment the Usage and Parameters page has `h1` headings in the ToC, meaning that they load with just the one top level item:
    
    I'm not sure that we need this any more, so maybe we could just show `h2` and below (or if we want to be careful, only do this if there is a single `h1` found in the page..?).

mashehu:
fixes #545
fix border-color for valid radio buttons in dark mode

New launch page button layout:
Screenshot 2020-10-30 at 17 11 09

I decided to leave the h1 as part of the ToC and just change the visibility level of the lower levels. It's easier that way (no fiddling with the counter in `generate_toc()` needed) and should cover most use cases.

Coming from #544, should we also change how we render the parameter names in the launch page? I am more okay to have them rendered like command line parameters here. I guess because of the layout where you enter the values just next to it and it therefore looks more natural, compared to having the text next to it on the parameters page. But we could style them a bit more like they would be in the nextflow config, so just drop the `--` and maybe add a `=`.

### DIFF
--- a/includes/pipeline_page/_index.php
+++ b/includes/pipeline_page/_index.php
@@ -276,7 +276,7 @@ else {
   $toc = '<nav class="toc">';
   $toc .= generate_toc($content);
   # Add on the action buttons for the parameters docs
-  if($pagetab == 'usage'){
+  if($pagetab == 'parameters'){
     $toc .= '
     <div class="btn-group w-100 mt-2 mb-1" role="group">
       <button class="btn btn-sm btn-outline-secondary" data-toggle="collapse" data-target=".schema-docs-help-text" aria-expanded="false">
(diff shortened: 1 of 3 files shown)


ewels (approved):

ewels:
Nice 👍🏻 

I think that we should either leave the parameter styling on the launch page as they are or make them the same as the parameter docs. I agree that here the current rendering makes a bit more sense, especially as there is a mixture of core Nextflow options and params, so for more experienced users this is made clear by the `--`.

Let's leave this for now and play around with it in a separate PR.. (maybe in #544 for example).

PR #550, opened 2020-11-03
 fix bug, where directories only containing directories were not displayed

mashehu:


drpatelh (approved):
Lets see if it works thennnnn!!

PR #561, opened 2020-11-06
Nav icon tweaks for pipeline page

ewels:


includes/pipeline_page/_index.php
```diff
@@ -215,27 +215,27 @@
 
 <ul class="nav nav-fill nfcore-subnav">
   <li class="nav-item">
-    <a class="nav-link<?php if($pagetab==''){ echo ' active'; } ?>" href="<?php echo $url_base; ?>">Introduction</a>
+    <a class="nav-link<?php if($pagetab==''){ echo ' active'; } ?>" href="<?php echo $url_base; ?>"><i class="fas fa-sign-in mr-2"></i> Introduction</a>
```
mashehu, line 218:
```suggestion
    <a class="nav-link<?php if($pagetab==''){ echo ' active'; } ?>" href="<?php echo $url_base; ?>"><i class="fas fa-sign-in mr-1"></i> Introduction</a>
```

mashehu (approved):
One small change, to match the other margins

includes/pipeline_page/_index.php
ewels: This difference was deliberate, not an accident as I felt that the icon was close enough to the text that it made it a little trickier to read.

But yeah, I guess I probably should have done the same for all icons rather than just one 🤔  Not a big deal anyway.

PR #565, opened 2020-11-13
schema builder: fix type conversion error for numbers as enum values

mashehu:
As pointed out on [slack](https://nfcore.slack.com/archives/CE5LG7WMB/p1605255297369400).


ewels (approved):
Don't 100% understand the issue here, but I trust that you do :)

mashehu:
when you entered `2|4` it was parsed as `["2","4"]` and therefor the default value of `4` was not equal to any elements there. Basically coming down to: 4!="4" 
With the changes here, it checks if the parameter type is anything numeric, which will then be parsed as float, so the above results in `[2,4]`.

Issue #551, opened 2020-11-03
Minor bug: s3 results explorer


ewels:
Minor bug: if a directory does not contain any paths, the view does not update when clicking the directory (the path a the top docs).

See for example https://nf-co.re/atacseq/example_output#atacseq/results-1b3a832db5a53c92c2c89a4d6d79455e860461ad/

![Nov-03-2020 14-06-44](https://user-images.githubusercontent.com/465550/97988815-d7c63080-1ddd-11eb-984b-c0ed2f7c30a6.gif)

mashehu:
Sorry, I can't reproduce this, because the directory you link to is not empty any more and I can't find any similar ones at the moment...

PR #567, opened 2020-11-13
Fix soft links back button + empty directories

### LINKED ISSUE #555: Megatests file browser - softlink browser back / forward buttons
It's fantastic how the soft-linking works for the s3 explorer for AWS megatests. One small improvement would be to listen to the browser back / forward events and honour them in the front end. Currently clicking forward or back changes the URL soft link but has no effect in the page.

mashehu:
Fixes #555 and #551.


ewels (approved):
Thanks!

PR #572, opened 2020-11-28
fix implementation of range parameters

mashehu:
This PR fixes the website part of https://github.com/nf-core/tools/issues/738 and should be merged together with https://github.com/nf-core/tools/pull/801.


ewels (approved):
Looks great! No harm in merging this right away, right? Feel free to do so if you think it's safe 👍🏻

PR #573, opened 2020-11-30
IFC talk

ewels:
New event for my talk on Friday. Includes minor tweak for label on BovReg event and also includes the time zone on the printed time at the top of the single event page.


ewels:
Quick merge so that I can advertise the event. Tests passing ✅

PR #576, opened 2020-12-02
Clarify nf-core lint --release command

jfy133:


mashehu (approved):
Yes, this way it is a lot better, thanks!

jfy133:
> Yes, this way it is a lot better, thanks!

Indeed, I was so confused when the command wasn't working! It's a implicit python-coder thing (amonst others) I've noticed, to assume you must always specify the input without a flag ;).

PR #577, opened 2020-12-02
Fix slack users and num-issues on stats page

ewels:
Hopefully this should fix both the Slack user count metric, and also the issues (808 issues in total apparently?? should be ~2k)


ewels:
Quick merge, as I would like this to be working due to the Slack 1000 users announcement. Plus a talk I'm giving tomorrow.

PR #590, opened 2020-12-09
Take a stab at the sync TEMPLATE branch cleanup docs

ewels:
Wrote very quickly, probably needs a little review - especially the final couple of commands.

### DIFF
--- a/markdown/developers/sync.md
+++ b/markdown/developers/sync.md
@@ -13,6 +13,7 @@ There are three topics covered in this documentation page:
 1. [Merging automated PRs](#merging-automated-prs)
 2. [Manual synchronisation](#manual-synchronisation)
 3. [Setting up a pipeline for syncing retrospectively](#setting-up-a-pipeline-for-syncing-retrospectively)
+4. [Fixing a broken `TEMPLATE` branch](#fixing-a-broken-template-branch)
 
 ### How template synchronisation works
 
@@ -359,10 +360,10 @@ be created to merge updates in to your pipeline for you.
 
 That's it, you're done! **Congratulations!**
 
-# Fixing a broken TEMPLATE branch
+# Fixing a broken `TEMPLATE` branch
 
 If merge conflicts are resolved via the GitHub interface instead of after pulling changes to a fork as described above, the commit history from the `dev` branch will be merged into `TEMPLATE`.
-This leads to complex problems in later `TEMPLATE` merges as the later updated `TEMPLATE` branch removes all the pipeline-specific files that were accidentality included in problematic merge, resulting in many (in some cases >100!) of files to resolve conflicts in.
+This leads to complex problems in later `TEMPLATE` merges as the later updated `TEMPLATE` branch removes all the pipeline-specific files that were accidentally included in problematic merge, resulting in many (in some cases >100!) of files to resolve conflicts in.
 
 If during one of the automated syncs you see you have an usually high number of changed files you can check whether `dev` was accidentally merged into `TEMPLATE` by looking at how many commits the `TEMPLATE` branch has (should be in the range of 5-15ish, depending on how many template updates your pipeline has had). You can also look at your repository's GitHub Network Graph under the _"Insights"_ tab or even look through the `TEMPLATE` branch commit history to see if there is a commit described as 'Merge branch `dev` into `TEMPLATE`'.
 
--- a/markdown/developers/sync.md
+++ b/markdown/developers/sync.md
@@ -4,6 +4,7 @@ subtitle: Learn how nf-core pipelines are automatically kept up to date with com
 ---
 
 # Introduction
+
 To keep all the nf-core pipelines up-to-date with the latest version of the community standards, we have implemented a synchronisation tool.
 This ensures that updates to the community standards are propagated to all nf-core pipelines.
 
@@ -194,6 +195,7 @@ Remember to go back to your `dev` branch as above before running `nf-core sync`.
 Much of the merging process should then be the same as described above with the automated pull requests.
 
 # Setting up a pipeline for syncing retrospectively
+
 This section describes how to set up a correct TEMPLATE branch in the case your pipeline was not created with a TEMPLATE branch from the beginning. If you created a pipeline with the `nf-core create` command, you should be all ready to go and can skip this step. Otherwise proceed with caution. It is probably a good idea to make sure you have all your local changes pushed to github and you could even make a local backup clone of your repository before proceeding.
 
 You should also consider the option to restart your pipeline project by running the `nf-core create` command and simply copy in the modifications you need into the newly created pipeline.
@@ -306,6 +308,7 @@ git push --set-upstream origin TEMPLATE
 ```
 
 ### Merge TEMPLATE into main branches
+
 The only remaining step is unfortunately a rather tedious one.
 You have to merge the `TEMPLATE` branch into your main pipeline branches, manually resolving all merge conflicts.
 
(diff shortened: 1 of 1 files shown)


markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
+If during one of the automated syncs you see you have an usually high number of changed files you can check whether `dev` was accidentally merged into `TEMPLATE` by looking at how many commits the `TEMPLATE` branch has (should be in the range of 5-15ish, depending on how many template updates your pipeline has had). You can also look at your repository's GitHub Network Graph under the _"Insights"_ tab or even look through the `TEMPLATE` branch commit history to see if there is a commit described as 'Merge branch `dev` into `TEMPLATE`'.
 
 If so, the easiest solution is to start your `TEMPLATE` branch from scratch.
 
-To do this, clone your repository to your local machine, ensuring you have pulled both `TEMPLATE` and `dev` branches.
+* Clone the main nf-core pipeline repository to your local machine (not your development fork)
```
MaxUlysse, line 371: That's a good clarification

markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
 
-Next, retrieve the commit hash when the original cookie-cutter template was used to generate pipeline i.e. with `nf-core create`.
-Assuming you originally started with the nf-core template, you can simply look at your git log from within your repository:
+  ```bash
+  git clone <git URL>
```
MaxUlysse, line 374: why not including the link directly: `https://github.com/nf-core/<pipeline>.git`

MaxUlysse: and do we need to use https or ssh, or both are working?

MaxUlysse (approved):
LGTM.
only improvment would be `https://github.com/nf-core/<pipeline>.git` instead of <git URL>

markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
 
-```bash
-git log --reverse
-```
+* Next, retrieve the commit hash when the original nf-core template was used to generate pipeline i.e. with `nf-core create`.
+  * Assuming you originally started with the nf-core template, you can simply look at your git log from within your repository:
 
-The first commit will then typically represent the original template.
+    ```bash
+    git log --reverse
```
MaxUlysse, line 382: - On which branch should we do that? (I'm assuming `TEMPLATE`

markdown/developers/sync.md
ewels: That would be better!

markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
+    ```
 
-Next we can delete the `TEMPLATE` branch with:
+  * The first commit will then typically represent the original template.
```
MaxUlysse, line 385: Is there a commit message created when we use nf-core create, or do we actually suggest one like `Initial template commit`

markdown/developers/sync.md
ewels: Shouldn't matter here actually as all branches should have the same root commit. But it will matter when we reset in the later command!

markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
+* Reset the `TEMPLATE` branch back to this commit, discarding all changes after that
 
-```bash
-git branch -D TEMPLATE
-```
+  ```bash
+  git reset --hard <hash of first commit after nf-core create>
```
ewels, line 389:
```suggestion
    git checkout TEMPLATE
    git reset --hard <hash of first commit after nf-core create>
```

markdown/developers/sync.md
ewels:
```suggestion
    git checkout TEMPLATE
    git log --reverse
```

markdown/developers/sync.md
ewels, line 389:
```suggestion
  # Make sure you're definitely have TEMPLATE checked out!
  git reset --hard <hash of first commit after nf-core create>
```

markdown/developers/sync.md
ewels:
```suggestion
  git clone git@github.com:nf-core/<PIPELINE>.git
```

markdown/developers/sync.md
```diff
@@ -358,58 +361,58 @@ That's it, you're done! **Congratulations!**
...
-git push origin TEMPLATE --force
-```
+* You probably want to now delete your local copy of the pipeline that you checked out from the main nf-core repository.
+* On your personal fork of the pipeline you'll want to pull in this fresh template branch:
 
-This will then replace the broken `TEMPLATE` branch on GitHub with a nice clean one, which can be viewable by checking the commit history.
+  ```bash
+  cd <path/to/forked/pipeline>
+  git branch -D TEMPLATE # Delete the TEMPLATE branch in your fork if you have it
+  git remote add upstream <git URL of main nf-core repo> # You might already have this set up?
```
ewels, line 413:
```suggestion
  git remote add upstream git@github.com:nf-core/<PIPELINE>.git  # You might already have this set up?
```

markdown/developers/sync.md
ewels: > and do we need to use https or ssh, or both are working?

Both should work, though https may need password on the cli. I think ssh is probably "better"

markdown/developers/sync.md
ewels:
```suggestion
  * The first commit will then typically represent the original template, with a commit message like `initial template build from nf-core/tools, version 1.9`
```

ewels: `nf-core create` usually makes the first commit after version 1.9 or so..

markdown/developers/sync.md
```diff
@@ -371,21 +371,23 @@ If so, the easiest solution is to start your `TEMPLATE` branch from scratch.
 * Clone the main nf-core pipeline repository to your local machine (not your development fork)
 
   ```bash
-  git clone <git URL>
+  git clone git@github.com:nf-core/<PIPELINE>.git
```
jfy133, line 374: Doesn't this assume someone is using ssh keys though?

jfy133 (approved):
This section of the page should also be added to the contents in the page's intro.

Also, there is also a typo (I thought I fixed earlier :thinking: ): accidentality should be accidentally

markdown/developers/sync.md
Zethson:
```suggestion
  git clone https://github.com/nf-core/<PIPELINE>
```

ewels: Do people not use ssh keys?

ewels: @Zethson - missing the `.git`

```suggestion
  git clone https://github.com/nf-core/<PIPELINE>.git
```

ewels:
> accidentally

Fixed / added 👍🏻

ewels:
@jfy133 - Fixed / added 👍🏻

markdown/developers/sync.md
Zethson: @ewels very few do from my experience. The default is certainly not.

The `.git` is not required. you can clone without it.

jfy133: I've seen mixed, which is why originally I think I didn't specify the exact clone mechanism.

Zethson: I would say that the people that are somewhat new to git/Github will use https and not clone using their SSH keys. The people however that do use their SSH keys will know how to adapt the URL to their needs, but this does not hold the other way around (more or less).

PR #591, opened 2020-12-09
Config docs - how to tune a workflow

ewels:
A common question  that we get in Slack is how to change the resource requirements for steps in a pipeline, often combined with _"why is `--max_memory` not doing anything?!"_

In the distant past, we tried not to document this too much as we considered it to be "core Nextflow" and not really specific to nf-core. But now we have a lot of people using nf-core pipelines who are not necessarily well versed in Nextflow already and really need some pointers on this.

Here is a first stab at extending the _Configuration_ documentation on the nf-core website a bit to cover some of the common points that often come up in Slack. Feedback welcome!

### DIFF
--- a/markdown/usage/configuration.md
+++ b/markdown/usage/configuration.md
@@ -3,11 +3,24 @@ title: Pipeline configuration
 subtitle: How configure Nextflow to work on your system
 ---
 
+- [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


MaxUlysse (approved):
LGTM

PR #596, opened 2020-12-10
add ToC to publications, move loading of scripts into header

mashehu:
A less markdown based solution compared to #594 
Screenshot 2020-12-10 at 11 26 04

PR #585, opened 2020-12-08
Added more detailed description about pipeline structure

KevinMenden:
This just adds the markdown file with the new pipeline structure description docs.
Easier to work on it here - comments on stuff that could be changed or added are welcome :-) 

This PR only adds the markdown file - so will still have to be added to the actual site.

Issue #482, opened 2020-07-30
Better ToC everywhere


mashehu:
Now that we have a nice ToC generator, should we add it as a sidebar element to all pages with a handwritten ToC (e.g. /stats, /tools)?

ewels:
Sounds good! Note that I would quite like to reorganise the tools docs / all docs generally: https://github.com/nf-core/tools/issues/670 https://github.com/nf-core/nf-co.re/issues/352 so the pages may be affected by that work. But no reason not to have sidebar navs generally I guess..?

Could maybe have a test to check for the length of the page if markdown, plus a variable to enable manually. That's how auto-ToCs work with Hugo and it's quite nice.

PR #597, opened 2020-12-10
augment markdown parsing with automated ToC, make first ToC item always active on the initial load.

mashehu:
Had a quick lunch-sprint to solve #482, at least for pages rendered from markdown (after it came up again for #596 and #585).

General functionality: Every time a markdown file is rendered a ToC is included (can be supressed by setting `$no_auto_toc` to true (see `_index.php`)).

Things to make this even better: 

- Find better criteria when the ToC is generated, currently I just look if it has a header element. Maybe number of header elements? Or number of characters (~= page length in parsed`$content`?
- Find markdown ToCs and remove them in the parsed `$content`.
- also use it for the toc in the pipeline docs (currently not yet possible, because of the hide/show buttons at the bottom there.


ewels:
Ok, after feeling guilty about not getting to this forever I have finally taken a quick look.

Look amazing 🎉 Love it!

Only comment really is that some of the ToCs are super long - so long that they kind of stop being useful. Two example pages:

* http://localhost:8888/tools/
* http://localhost:8888/usage/tutorials/nf_core_tutorial#running-nf-core-pipelines

It looks like `<h2>` gets top priority in the ToC - I wonder if we switch this to `<h1>` if only for the auto-generated ToCs? What do you think?

ewels:
This also makes me think again that we should maybe try to make all of the page widths wider.. Maybe upgrade to Bootstrap 5 in the process? 👀

mashehu:
Fixed the collapse-level:
Screenshot 2021-05-24 at 15 51 19

Next step is to remove hard-coded ToC from all markdowns now 🙂

mashehu:
Okay, I removed the hard-coded ToCs, where I thought they were superfluous (but kept them for example for the troubleshooting page).
I don't like the bold text in some ToCs for longer top-level elements, but I don't know how to avoid these (except for adding a higher level heading above them): 
Screenshot 2021-05-24 at 17 13 20

ewels (approved):

mashehu:
The ToC prioritization unfortunately doesn't look if it is `h1` or `h2`, but only what is the highest heading on the page (which is in this cage h2). So, we would need to always start counting from an h1, but I don't want to touch the ToC generation counting again. Leaving this as a pointer if somebody want to give it a try. 🙂

ewels:
Sounds good 👍🏻

PR #598, opened 2020-12-10
Added 'Financial Support' section to about, darkmode for any image

ewels:
I realised that we don't describe our funding sources anywhere on the website, which is definitely something that we should celebrate. Here I added a new section on the "About" page describing the grants that specifically fund us.

I also wanted to have logos that work in dark-mode, so I added generic support for this with any image.

* In markdown, use an `` tag and make sure that it has the class `darkmode-image`. The code will do the rest.
* In PHP, build on the following minimal example:
  ```html
  <img src="/assets/img/contributors-<?php echo $theme == 'dark' ? 'white' : 'colour'; ?>/IMAGE.svg" class="darkmode-image" />
  ```

Would be great to also get dark-mode compatible images for DFG and BovReg, 'cc @ggabernet @JoseEspinosa.


ewels:
oh and I also added some clearfixes on the community page to fix the funky image floats for the _Projects we are involved with_ section.

apeltzer (approved):

apeltzer:
Adding images from institutions that work in dark mode would be great too, though that might be quite some work involved to get these for all images we have currently.

MaxUlysse:
> Adding images from institutions that work in dark mode would be great too, though that might be quite some work involved to get these for all images we have currently.

We have monochrome versions of all logos already, should be fairly easy to switch to these ones in dark mode

ewels:
Err, have you guys tried switching to dark mode? That's worked for about a year now already 😅

PR #599, opened 2020-12-11
Tweaks based on feedback from beta testing of branch fixing of ampliseq

jfy133:
- Replaces the clone command with HTTPS rather than SHH key (as HTTPS is default github clone mechanism)
- Adds missing command to fetch the upstream branch information into the local pipeline


ewels (approved):
Awesome, thanks! 🚀 

I had a feeling that I might have been missing a `git fetch`. Was writing those commands in the 4 minutes before my work christmas party started so was rushing a bit too much, sorry! 😓

PR #604, opened 2020-12-18
add white background to base image for logos

mashehu:
improves readability on github with darkmode (works for pipeline names up to ~43 characters, which should be enough, I guess 🙂)


ewels (approved):
Nice! Can we strip out the code about alpha support? I guess it doesn't matter so happy to merge with it still in (might even be nice to keep it in case we want something different in the future actually).

mashehu:
I don't think it hurts and for future compatibility I also left the base image with the transparent background.

ewels:
Cache has been wiped and regenerated for all nf-core pipelines 👍🏻

PR #607, opened 2021-01-04
Update release checklist

jfy133:
Minor tweak for clarificaiton, as AFAIK `nf-core` tools does not bump the CHANGELOG itself (at least doesn't for eager)


ewels (approved):
Nice! It doesn't say that `nf-core bump-version` does alter the changelog here does it? But additional detail is good anyway.

Maybe link to the release name generator website instead of a specific release of tools. eg. See the tools release checklist for a URL: https://github.com/nf-core/tools/blob/master/.github/RELEASE_CHECKLIST.md

Phil

jfy133:
No it doesn't say that, but as a new user I would maybe assume it does.

I only noticed last time that it wasn't updated when I double checked the changlog, and that made me realise if a user checked that first before doing the bump they might forget.

Ooh nice, haven't seen that before. Will update it tomorrow 👍

PR #611, opened 2021-01-14
Logo caching - include width in cache filename if given

ewels:
I broke stuff with my recent changes to logo generation where we cache the results. The website cache doesn't take into account that the script can specify a target width for the logo. When creating a new pipeline we get two copies of the logo - 400px wide for the email and 600px wide for the readme. The 400px one is fetched first. If it's a new pipeline, then that image is generated first and then returned for all future requests.

This also means that all pipelines will probably have pulled full size logos instead of smaller versions since I implemented the caching. Just whatever width was specified the first time.

Oops. Sorry.

### DIFF
--- a/public_html/logo.php
+++ b/public_html/logo.php
@@ -6,6 +6,14 @@ if(!isset($_GET['f'])){
     $textstring = strtolower($textstring);
     $textstring = preg_replace("/[^a-z]/", '', $textstring);
 }
+
+$new_width = false;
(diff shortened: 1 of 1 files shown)


apeltzer (approved):

ewels:
Tested, works, have removed all cached logos and regenerated 3 copies (full size, 600, 400) for all existing pipelines.

PR #616, opened 2021-01-29
Locale-aware event times

ewels:
Used [Moment Timezone](https://momentjs.com/timezone/) to guess the user's time zone and convert the event time.

Tested on these pages:

* https://nf-co.re/events/2019/hackathon-qbic-2019 (has no start time, only a date)
* https://nf-co.re/events/2021/bytesize-1-nf-core-into (for me now shows in CET correctly).

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -58,7 +58,11 @@ if(isset($subtitle) && strlen($subtitle) > 0){
     <script src="/assets/js/bootstrap.min.js"></script>
     <script src="/assets/js/highlight.pack.js"></script>
     <!-- Page-specific CSS and JS -->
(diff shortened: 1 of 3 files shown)


mashehu (approved):
Looks good to me.
Do you think it is possible to also parse the markdown file itself and replace times there?

ewels:
> Do you think it is possible to also parse the markdown file itself and replace times there?

Ugh, sounds difficult 😓 

Maybe if we use HTML elements in the markdown with extra data attributes? That's what I did for the dark mode image support:

https://github.com/nf-core/nf-co.re/blob/fd1bdd7a24bfe116e8f24e160642810ea723a88c/markdown/about.md#L55

PR #632, opened 2021-02-16
update bytesize #3 link

ggabernet:
- Update to edited video version


MaxUlysse (approved):
I think 10 sec can still be removed at the beginning.
But already a good improvement, so I approve anyway

PR #635, opened 2021-02-17
Homepage dev tips

ewels:
## Added new homepage section

Added new section above feature cards with three key points for potential developers:

image

My hope is that by making this more visible, we will catch developers earlier and reduce the number of people coming to us with fully formed pipelines that don't meet the guidelines (most common is not using `nf-core create` or having an existing pipeline in place).

## Guidelines

* Added table of contents
* Move 'ask the community' to the bottom and renamed header
* Add section saying that people must use the template (previously was not mentioned in the guidelines)

### DIFF
--- a/markdown/developers/guidelines.md
+++ b/markdown/developers/guidelines.md
@@ -12,22 +12,17 @@ pipelines must adhere to.
 > If you're thinking of adding a new pipeline to nf-core, please read the documentation
 > about [adding a new pipeline](adding_pipelines.md).
 
-## General rules
+* [Workflow specificity](#workflow-specificity)
+* [Workflow size](#workflow-size)
+* [Use the template](#use-the-template)
+* [Minimum requirements](#minimum-requirements)
+* [Recommended features](#recommended-features)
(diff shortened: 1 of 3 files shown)


ewels:
Screenshot of the full homepage, for context:

localhost_8888_

Dark mode: 

image

MaxUlysse (approved):
Love it

mashehu:
I like the idea. Coming to the page as a pipeline user and not developer (which is the majority I guess), I think this is a bit confusing. I would address developers more clearly and actually switch the order between the pipeline panel underneath and the new one.

ewels:
>  I would address developers more clearly and actually switch the order between the pipeline panel underneath and the new one.

Yes I wondered the same thing, but was so keen to make it as huge and flashing and attention grabbing as possible that I started by putting it there 😅  But looking at it fresh, I think you're right, it should go down one position.

Nice suggestion about making it more obviously targeted to developers 👍🏻

ewels:
Yeah, I think this looks better. The gradient of greys looks better like this too!

localhost_8888_ (1)

PR #641, opened 2021-02-17
Add troubleshooting step for memory weirdness

BrunoGrandePhD:
I ran into this issue as a newcomer to Nextflow, so I wanted to extend the troubleshooting docs to gratefully pay forward the assistance I got from the sarek team (nf-core/sarek#347), even through it wasn't an issue with their workflow; just my own Nextflow/Groovy ignorance.


FriederikeHanssen:
I don't know what the documentation wizards think, but I think it would be worthwhile to reference this https://nf-co.re/usage/configuration#tuning-workflow-resources as well, as it explains a bit more about resource allocation.

ewels:
The `--max_memory` will now be validated for this, but not configs that target the process specifically.

However, the pipeline should have logic in to sanity check this really. I think that I remember writing something along these lines years ago. With the new DSL2 modules it should be fairly easy to centralise so all pipelines can use it.

In short, thank you for the documentation but I think I'd rather fix this rather than document how to avoid it..

BrunoGrandePhD:
I'm happy with a preventative solution rather than relying on docs. I made a docs contribution only because I'm not yet proficient with Nextflow. Feel free to close this PR without merging!

PR #644, opened 2021-02-18
Fix pipeline health page, add template branch protection

### LINKED ISSUE #638: Add branch protection to TEMPLATE
I think it would be a good idea to limit the `TEMPLATE` branch for each pipeline to `nf-core-bot`

I actually implemented that in nf-core/sarek, but that's one pipeline, we can and should automatize that for all pipelines.

ewels:
* Refactor code to check HTTP API response codes
    * GitHub changed the exact string that they return in the HTTP header for successful responses (`HTTP/1.1 200` instead of `HTTP/1.1 200 OK`) which broke everything.
    * Refactored the code that checks these strings to work with the new + old format.
* Added new test for branch protection on `TEMPLATE`
  * Simpler than branch protection for `dev` and `master` - just checks that nothing is allowed apart from pushes from @nf-core-bot

Closes #638

### DIFF
--- a/public_html/pipeline_health.php
+++ b/public_html/pipeline_health.php
@@ -102,6 +102,7 @@ class RepoHealth {
   public $gh_branches;
   public $gh_branch_master;
   public $gh_branch_dev;
+  public $gh_branch_TEMPLATE;
   public $gh_webpage;
   public $gh_social_preview;
 
@@ -135,6 +136,7 @@ class RepoHealth {
   public $branch_dev_code_owner_reviews;
   public $branch_dev_required_num_reviews;
   public $branch_dev_enforce_admins;
+  public $branch_template_restrict_push;
 
   public function get_data(){
     $this->get_repo_data();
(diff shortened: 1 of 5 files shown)


public_html/pipeline_health.php
```diff
@@ -687,6 +734,7 @@ function get_gh_team_repos($team){
   'branch_dev_code_owner_reviews' => 'dev: code owner reviews',
   'branch_dev_required_num_reviews' => 'dev: 1 review',
   'branch_dev_enforce_admins' => 'dev: enforce admins',
+  'branch_template_restrict_push' => 'T push',
```
mashehu, line 737:
```suggestion
  'branch_template_restrict_push' => 'TEMPLATE: restrict push',
```
Any reason to keep it so short?

mashehu (approved):

public_html/pipeline_health.php
ewels: Yeah, I wrote that originally but it makes the column a lot wider, compressing all of the other columns a lot as a result.

The header isn't so important any more to be honest as we have so many repos that it usually scrolls out of view quite quickly. Hovering over any cells in the column shows the full description so it's easy to see what the failures are.

Issue #639, opened 2021-02-17
Check that master == latest release


ewels:
A new pipeline health check that could be useful is to check the commit on `master` against the commit hash on the latest release (if there is one).

This could help to alert us about pipelines that have changes merged to master but no release.

PR #645, opened 2021-02-18
Pipeline health - check that master is last release

ewels:
* Compare commit hash of the last release against the commit hash on master
* Restyle invalid tests to be green instead of grey, to improve contrast of failing tests

### DIFF
--- a/public_html/pipeline_health.php
+++ b/public_html/pipeline_health.php
@@ -529,13 +529,14 @@ class RepoHealth {
   public function print_table_cell($test_name){
     $test_url = $this->test_urls[$test_name];
     $test_url = str_replace('{repo}', $this->name, $test_url);
+    $test_url = str_replace('{latest-tag}', $this->last_release->tag_name, $test_url);
     if(is_null($this->$test_name)){
       echo '<td class="table-secondary text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
         <a href="'.$test_url.'" class="d-block" target="_blank"><i class="fas fa-question text-secondary"></i></a>
       </td>';
     } else if($this->$test_name === -1){
-      echo '<td class="table-secondary text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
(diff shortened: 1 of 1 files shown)


ewels:
Closes #639

public_html/pipeline_health.php
```diff
@@ -529,13 +529,14 @@ private function _save_cache_data($path, $data, $encode_json=true){
...
     $test_url = str_replace('{repo}', $this->name, $test_url);
+    $test_url = str_replace('{latest-tag}', $this->last_release->tag_name, $test_url);
     if(is_null($this->$test_name)){
       echo '<td class="table-secondary text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
         <a href="'.$test_url.'" class="d-block" target="_blank"><i class="fas fa-question text-secondary"></i></a>
       </td>';
     } else if($this->$test_name === -1){
-      echo '<td class="table-secondary text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
-        <a href="'.$test_url.'" class="d-block" target="_blank"><i class="fas fa-times text-secondary"></i></a>
+      echo '<td class="table-success text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
```
mashehu, line 538:
```suggestion
      echo '<td class="table-warning text-center" title="<strong>'.$this->name.':</strong> '.$this->test_descriptions[$test_name].'" data-toggle="tooltip" data-html="true">
```
green = good, yellow = 🤷‍♂️, imo.

mashehu (approved):

public_html/pipeline_health.php
ewels: Yeah that's what I originally had, but then I changed it to grey a while backto be more of an `N/A` than 🤷🏻‍♂️ 

But now even the grey I find a bit tricky as you have a wall of green then one column that is all mixed grey and red. We generally don't care about green or grey and only care about red, so switching it to green like this makes the reds pop out much more.

ewels: Take a look after merge - can always change it back if it bothers you.. 😄

mashehu: okay, makes sense to have a red and not-red status. 👍

PR #678, opened 2021-03-18
Strip out old tools lint error page, update URL to new sphinx docs

ewels:
To be merged after version 1.13 of nf-core/tools is released.


ewels:
To be fair, can merge straight away as the sphinx docs are based on `dev` anyway so the URLs work..

PR #680, opened 2021-03-18
Fix ToC for event pages

mashehu:
Fixes bug mentioned in https://github.com/nf-core/nf-co.re/pull/675#issuecomment-802033646. The problem was that there was no `h1` heading in these pages, so all counting was off by one. The code now looks for the highest heading and shift everything by the deviation from 1.

Removes code naïvely circumventing the table rendering bug.


ewels (approved):
Great! These auto-ToCs are a pain in the arse, but they are amazing when they work 🤩

PR #697, opened 2021-03-22
Adds newer avaliable container engines to Installation page

jfy133:
To close https://github.com/nf-core/nf-co.re/issues/696

Input from @phue would be helpful for the descriptions, I've not used any of the new ones and the descriptions on the respective websites weren't so clear on why they are different ;).

### DIFF
--- a/markdown/usage/installation.md
+++ b/markdown/usage/installation.md
@@ -57,7 +57,7 @@ nf-core pipelines utilise the built-in support for software packaging that Nextf
   * Often used as an alternative to Docker on multi-user systems such as HPC systems.
   * Uses Linux user namespaces to run containers with no privileged operations or daemons
 * [Shifter](https://www.nersc.gov/research-and-development/user-defined-images/)
-*  An experimental implementation of container system that can convert from a wide range of other images
+  * An experimental implementation of container system that can convert from a wide range of other images
 * [Conda](https://conda.io/)
   * Packaging system that manages environments instead of running analysis in containers.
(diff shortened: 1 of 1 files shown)


phue:
@jfy133 
Thanks for adding this. The descriptions look good and sufficient to me as they are, some points for additional context:
I think all of them are daemonless.
- `Podman` is cli compatible to docker and can be used as drop-in replacement. It also has a rootless mode (which still has some issues at the moment).
- `Charliecloud` is considered more secure because it works with user namespaces only (doesn't require the `setuid` bit as `Singularity` does for example)
-  `Shifter`: I never used it, so I don't really know the details either. We just added it because nextflow supports it out-of-the-box

markdown/usage/installation.md
```diff
@@ -50,6 +50,14 @@ nf-core pipelines utilise the built-in support for software packaging that Nextf
   * Often used as an alternative to Docker on multi-user systems such as HPC systems.
   * Also runs _containers_ and can create these from Docker images
   * Does not need root access or any daemon processes - images built from files
+* [Podman](https://podman.io/)
+  * Often used as an alternative to Docker on multi-user systems such as HPC systems.
+  * Is a daemonless container engine
```
phue, line 55:
```suggestion
  * Is a daemonless container engine that can serve as a drop-in replacement for Docker.
```

phue (approved):

jfy133:
Thanks for the clarifications @phue !

PR #724, opened 2021-04-13
Fix for showing the countdown for shorter events

mashehu:
Should we maybe extend the window to 24h instead of 1h?


ewels (approved):
Good spot 👍🏻 

Yes I think 24h would be better..

ewels:
Could do 7 days ahead for hackathons etc too..

PR #727, opened 2021-04-13
Tools directory

ewels:
Prep website for new multi-version tools documentation. Combines with https://github.com/nf-core/tools/pull/1004

If this works, will follow up with a PR to update the navigation and remove the old "lint errors" stuff / pulling tools repo branch etc.


ewels:
Self merging, so that I can do a test deployment

PR #735, opened 2021-04-16
Add bytesize 9

mashehu:
I also prefixed the running index of the bytesize files with a `0`, so the ordering will not be messed up in filesystems 🙂


ewels:
Ah should have done the `0` prefix from the start! This will break the existing URLs though which have already been quite widely shared, so I'm not sure if it's worth it?

ewels (approved):
Thanks!

PR #739, opened 2021-04-19
fix incorrect call of scroll_to function

mashehu:
Now scrolls nicely to the correct position 🙂


ewels (approved):
Would be nicer in a variable but otherwise good :)

PR #755, opened 2021-04-27
check if emoji file exist before replacing name between two `:` by an emoji image

### LINKED ISSUE #754: emoji parsing not specific enough
**Describe the bug**
Currently, every word between two `:` is parsed as an emoj, leading to some weird renderings.

**To Reproduce**
e.g. the :STAR_ALIGN: in https://nf-co.re/rnaseq/dev/usage#resource-requests

**Screenshots**
image (2)

**Additional context**
Probably best solution for this bug, as proposed by @ewels: Check if image exists before adding it in the `` in parse_md.

mashehu:
fixes #754

### DIFF
--- a/includes/parse_md.php
+++ b/includes/parse_md.php
@@ -114,8 +114,15 @@ function parse_md($markdown){
   }
 
   // Find and replace emojis names with images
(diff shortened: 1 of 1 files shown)


ewels (approved):
Very tidy! 🚀

PR #760, opened 2021-04-30
fix integer printing (int!= integer), keep 0 and false as defaults 

mashehu:
fireproof some value checks (by replacing == with ===), 
fix typos

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1675,7 +1675,7 @@ function clean_empty_param_keys(param){
     if(param.hasOwnProperty('description') && param['description'] === ''){
         delete param['description'];
     }
-    if(param.hasOwnProperty('default') && param['default'] === ''){
+    if(param.hasOwnProperty('default') && (param['default'] === '' || param['default'] === false)){
         delete param['default'];
     }
     if(param.hasOwnProperty('help_text') && param['help_text'] === ''){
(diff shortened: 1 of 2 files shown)


mashehu:
fixes 759

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -263,7 +263,7 @@ $(function () {
             };
             $.post( "pipeline_schema_builder", post_data).done(function( returned_data ) {
                 console.log("Sent schema to API. Response:", returned_data);
-                if(returned_data.status == 'recieved'){
+                if(returned_data.status == 'received'){
```
ewels, line 266: Good spot! 🤦🏻

ewels (approved):
I'm generally a bit terrified of using `===` as I've been bitten by it in the past, but I can see why it's needed to fix this issue - good find 👍🏻  If you're sure it's working in all of the other cases as well then I'm happy to merge :)

public_html/assets/js/nf-core-schema-builder.js
```diff
@@ -1672,16 +1672,16 @@ function clean_empty_schema_keys(subschema){
 function clean_empty_param_keys(param){
 
     // Clean up empty strings
-    if(param.hasOwnProperty('description') && param['description'] == ''){
+    if(param.hasOwnProperty('description') && param['description'] === ''){
         delete param['description'];
     }
-    if(param.hasOwnProperty('default') && param['default'] == ''){
+    if(param.hasOwnProperty('default') && param['default'] === ''){
```
mashehu, line 1678: This is the only place I am not 100% sure, because this changes some behavior to before, where everything evaluating to `false`, e.g. `0`, but also `false` was removed, @ewels

ewels: Hmm, yes this might need a little more code then. I think that we want to remove empty strings and boolean `false` but maybe not numeric `0`?

```suggestion
    if(param.hasOwnProperty('default') && (param['default'] === '' || param['default'] === false)){
```

Untested - what do you think?

PR #774, opened 2021-05-21
Make transcripts a bit more interactive

mashehu:
Main two features:
- timestamps in the transcripts move the embedded video to that timepoint (and not link to the page on youtube)
- the paragraph currently talked about in the video is highlighted (if it contains a timestamp)

https://user-images.githubusercontent.com/6169021/119160918-072a4b00-ba59-11eb-96de-6219c2cc248c.mp4

Open to a different background color to make the highlight more visible with the light theme (didn't want to add yet another color 🙂):
Screenshot 2021-05-21 at 17 13 47


ewels:
Nooo way, this is awesome! 🚀 

We should really work up two separate bootstrap themes for light and dark, then we can just use class names like `.bg-light` that would work in both. I can't see the highlight at all on the light mode one so I think it does need another CSS colour for now sorry.

Seeing as you've made it this far, would it be possible to make the transcripts div quite short but scrollable, then move the scroll alongside the highlight? So that the current text sticks below the video like a subtitle? Then could have a button to expand it out to the full height to be easier to read as prose.

ewels:
I hadn't cottoned on that you guys were manually adding the YouTube timestamps as you go along. That's insane! Massive kudos transcription team! 🙇🏻

ewels (approved):
Happy to merge as-is, comments are only for further additional features.

mashehu:
> Seeing as you've made it this far, would it be possible to make the transcripts div quite short but scrollable, then move the scroll alongside the highlight? So that the current text sticks below the video like a subtitle? Then could have a button to expand it out to the full height to be easier to read as prose.

Unfortunately, not every paragraph has a time stamp at the moment, so I don't think this will work. We should maybe make an issue to optimize the transcripts for that and then add this feature.

mashehu:
I added a more transparent version of `bg-success` to also nf-core theme it a bit 🙂 
Screenshot 2021-05-24 at 18 01 35
Screenshot 2021-05-24 at 18 01 52
The padding is quite tight...

ewels:
Great! Could we not do it as-is and just scroll as we highlight? As long as the height is long enough that you can read ahead a fair few paragraphs I think it should be fine? Just worried that it will be tricky to follow along as it is currently as you might have to be scrolled miles away from the video to see the current text..

mashehu:
Now with automatic scrolling text, @ewels 📜 
Jun-04-2021 11-44-40

ewels (approved):
amazing

PR #781, opened 2021-05-28
Added transcript for the talk

renbot-bio:


markdown/events/2021/bytesize-13-tuning-pipeline-performance.md
```diff
@@ -35,3 +35,48 @@ The talk will be presented on Zoom and live-streamed on YouTube:
...
+
+[9:28](https://youtu.be/Qw1gLpYtMec?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=568) It could be the default that is there for each of the processes, but these would likely be insufficient for all the processes in the pipeline. So it should be defined if some processes need more resources. For nf-core pipelines, we define this in the base.config that specifies different types of processes. So for example process_low that will need more resources as a default but still low resources or process_medium that will increase to a minimum resource definition here. This is done via labels and this code will also be a part of the nf-core template, so you don’t need to define this. You should just try to use those labels as they are defined in the base.config when defining your own processes. So for example, for the `FASTQC` process, this one contains a process_medium label already, meaning that for this process, these resource requests will be used. You can also define these labels for new processes as part of nf-core pipelines. Then the resource request definitions are already pretty fine.
```
jfy133, line 70:
```suggestion
[9:28](https://youtu.be/Qw1gLpYtMec?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=568) It could be that there is a default for each of the processes, but these would likely be insufficient for all the processes in the pipeline. So it should be defined if some processes need more resources. For nf-core pipelines, we define this in the base.config that specifies different types of processes. So for example  `process_low` that will need more resources as a default but still low resources or `process_medium` that will increase to a minimum resource definition here. This is done via labels and this code will also be a part of the nf-core template, so you don’t need to define this. You should just try to use those labels as they are defined in the base.config when defining your own processes. So for example, for the `FASTQC` process, this one contains a process_medium label already, meaning that for this process, these resource requests will be used. You can also define these labels for new processes as part of nf-core pipelines. Then the resource request definitions are already pretty fine.
```

markdown/events/2021/bytesize-13-tuning-pipeline-performance.md
```diff
@@ -35,3 +35,48 @@ The talk will be presented on Zoom and live-streamed on YouTube:
...
+
+[10:58](https://youtu.be/Qw1gLpYtMec?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=658) However for some of the processes, you might need to define some defaults that don’t match any of the labels here. In this case you can also define the base resources request for that process `withNAME` statement, and there you provide the name of the process directly instead of a label.
```
jfy133, line 72:
```suggestion
[10:58](https://youtu.be/Qw1gLpYtMec?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=658) However for some of the processes, you might need to define some defaults that don’t match any of the labels here. In this case you can also define the base resources request for that process `withName` statement, and there you provide the name of the process directly instead of a label.
```

jfy133 (approved):
Couple of very minor tweaks but looks good to me (just accept and merge directly :) )

renbot-bio:
Thank you @jfy133 :-)

PR #790, opened 2021-06-04
Make toc more mobile friendly

### LINKED ISSUE #747: ToC should come first on mobile
When viewing eg. parameter docs on mobile, the ToC comes last in the content so you have to scroll through the whole page to get to it.

Should probably come first. Maybe be collapsible etc. so it stays visible whilst scrolling? eg. off-canvas with button in header? Not sure.

mashehu:
Rather simple solution...which means that the sidebar on the pipeline pages is always above the content for smaller displays. I checked how they solved this on npmjs.com, but they also show their sidebar above the content on mobile.

I also centered the release dropdown in the subnav of the pipeline pages on mobile.

Fixes #747 ✈️


ewels (approved):
Looks good!

ewels:
I guess an off-canvas type thing would still be nice, as once you start scrolling the ToC is still gone. But this is definitely a big improvement 👌🏻

PR #796, opened 2021-06-09
Updated with transcript and links

renbot-bio:
Transcript + Bilbili and Figshare.


markdown/events/2021/bytesize-16-module-test-data.md
```diff
@@ -27,3 +28,40 @@ This week, Kevin Menden ([@KevinMenden](http://github.com/KevinMenden/)) will pr
...
+
+[4:54](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=294).. For this the most important file is the `test_data.config` file, which is stored in `modules/tests/config/test_data.config directory`. So this file is used to define all these different test data files that are available. It has a link to the test data directory, which is the nf-core test dataset repository, and specifically the modules branch of this repository where we store all the test data. There are links to specific files in this directory, and these links are again linked in this dictionary. This is just a small example of this whole test data parameter dictionary. The reason we do it like this is so that we are able to change these links at some point without changing the code in the `test_main.nf` files of the modules because you will only have these keys here and these entries are defined in your script, so you don’t really need to care about what’s happening on this side of the code. So this is where I can look for the keys and the data I need in the file.
```
jfy133, line 47:
```suggestion
[4:54](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=294).. For this the most important file is the `test_data.config` file, which is stored in `modules/tests/config/test_data.config directory`. So this file is used to define all these different test data files that are available. It has a link to the test data directory, which is the nf-core test dataset repository, and specifically the modules branch of this repository where we store all the test data. There are links to specific files in this directory, and these links are again linked in this dictionary. This is just a small example of this whole test data parameter dictionary. The reason we do it like this is so that we are able to change these links at some point without changing the code in the `test_main.nf` files of the modules because you will only have these keys here and these entries are defined in your script, so you don’t really need to care about what’s happening on this side of the code. So this is where I can look for the keys and the data I need in the file.
```

jfy133 (approved):
Not sure the comment I've is necessary, the GitHub interface seems to stop higlighting the inline code blocks halfway through :shrug:

markdown/events/2021/bytesize-16-module-test-data.md
```diff
@@ -27,3 +28,40 @@ This week, Kevin Menden ([@KevinMenden](http://github.com/KevinMenden/)) will pr
...
+
+[6:52](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=412) We currently have mainly two genomes available, one is the sarscov2 and the other is a subset of the human chromosome 22. The reason for this is that sarscov2 is really small, so we can really use the whole genome, and use all kinds of files for this genome without having large datasets. This is important for two reasons, first, we want to have small datasets to test, so that the test repository does not explode and the tests are also smaller for small datasets. For the human genome viz. part of chromosome 22, we again have a small dataset. Currently we only have genomics data available on the test dataset repository, and we also have some non-standard genomics data for specific tools that have been added by users.
```
KevinMenden, line 51:
```suggestion
[6:52](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=412) We currently have mainly two genomes available, one is the sarscov2 and the other is a subset of the human chromosome 22. The reason for this is that sarscov2 is really small, so we can really use the whole genome, and use all kinds of files for this genome without having large datasets. This is important for two reasons, first, we want to have small datasets to test, so that the test repository does not explode and the tests are also smaller for small datasets. For the human genome subset of chromosome 22, we again have a small dataset. Currently we only have genomics data available on the test dataset repository, and we also have some non-standard genomics data for specific tools that have been added by users.
```

markdown/events/2021/bytesize-16-module-test-data.md
```diff
@@ -27,3 +28,40 @@ This week, Kevin Menden ([@KevinMenden](http://github.com/KevinMenden/)) will pr
...
...[8:57](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=537) So first, it is important to read the instructions (in the `README` in the test dataset repository). We have a few lines about how to add new test data. Make sure to check this out and then think about a suitable minimal test dataset. The optimal thing here would be if you could find something that went with sarscov2 because that’s currently our favourite genome because it is small and there’s a lot of data available. So hopefully you can use this because then everything stays small and we don’t need to add a new genome. Then once you identify something, it is best to ask for opinions to gauge if it suitable for the test dataset repository in the `#modules` channel on Slack. Once you’ve established that it’s OK, you can create a pull request to the `modules` branch of the test dataset repository and ask for reviews `#modules` so that you can eventually have your pull request merged. Once that’s done and your data is available, you need to add an entry to the `test_data.config` file in the `modules` repository. That’s again a separate pull request to have an entry merge, but it shouldn’t be a problem.
```
KevinMenden, line 57:
```suggestion
[8:57](https://youtu.be/QXfAerydAT0?list=PL3xpfTVZLcNiSvvPWORbO32S1WDJqKp1e&t=537) So first, it is important to read the instructions (in the `README` in the test dataset repository). We have a few lines about how to add new test data. Make sure to check this out and then think about a suitable minimal test dataset. The optimal thing here would be if you could find something that went with sarscov2 because that’s currently our favourite genome because it is small and there’s a lot of data available. So hopefully you can use this because then everything stays small and we don’t need to add a new genome. Then once you identify something, it is best to ask for opinions to check if it's suitable for the test dataset repository in the `#modules` channel on Slack. Once you’ve established that it’s OK, you can create a pull request to the `modules` branch of the test dataset repository and ask for reviews `#modules` so that you can eventually have your pull request merged. Once that’s done and your data is available, you need to add an entry to the `test_data.config` file in the `modules` repository. That’s again a separate pull request to have an entry merge, but it shouldn’t be a problem.
```

KevinMenden (approved):
Awesome, thanks!
Just suggested some minimal changes, otherwise looks really good! :+1:

renbot-bio:
> Not sure the comment I've is necessary, the GitHub interface seems to stop higlighting the inline code blocks halfway through 🤷

Not sure what happened there... Thanks both @jfy133 and @KevinMenden for reviewing this so quickly.

PR #822, opened 2021-07-07
fix dates for upcoming bs talks

mashehu:


ggabernet (approved):

ggabernet:
Thanks for spotting the mistake!

Issue #541, opened 2020-10-28
Add package manager and task runner for dependencies


mashehu:
As brought up in https://github.com/nf-core/nf-co.re/pull/514#discussion_r512683214, since we are adding quite a few libraries now, some package management should help us keep it tidy.

Because most of our deps are frontend code, npm would be the easiest choice. If we want to make it extra consistent, we could also use composer for php related stuff.

I think it is easiest if we do all the compiling/moving of the deps inside npm and not add gulp or something similar.

Along with this step, we could actually switch to sass/scss to simplify all the styling (and make the dark mode more consistent) and add an auto-prefixer as well (something like [this](https://itnext.io/setting-up-a-sass-build-process-aa9fd92fa585)). 
The best would be to add a `npm watch` to the docker compose command, so the developer doesn't have to start too many things.  If we want  we could have an `npm build` github action to be sure that we always have the compiled versions in the github-repo.

ewels:
+1 for automating as much as possible so that we can effectively forget about the build steps 👍🏻

ewels:
I finally got around to playing with npm the other day. Got it running with Bootstrap (plus a nice bootstrap dark mode package) and a few other bits in a Docker compose script. Might be able to steal bits of that.

mashehu:
I have a branch lying around with npm+sass conversion. Unfortunately, how to put it into the docker compose was my problem. I want that it automatically transpiles the files to css on save, but haven't figured out how yet...

mashehu:
Update: I got it to run (it was looking from the wrong `working_dir`). Something I can't figure out in your code there: How/when do you trigger `npm run build` to compile autoprefixed and purged css files?

mashehu:
Ah, I was too fast. I found it now in the backend dockerfile. 🙂

PR #823, opened 2021-07-07
Add event download links + composer as dependency manager

mashehu:
Added a dropdown with calendar event links to non-past events, in the list view:
Screenshot 2021-07-07 at 17 24 36
and the details view: 
Screenshot 2021-07-07 at 17 24 18
Open for inputs about the wording and placement of these buttons.

Since I was adding another external library, I decided to switch all php dependencies to be managed by composer (partially addressing #541 with it). Therefore, I added another image to `docker-compose.yml` which runs `compose install` on every start, so one doesn't need to manually install the dependencies locally.
I tested that this doesn't break the implementation of any of the other libraries we use (twitteroauth, parsedown(-extra) and Spyc). But it would be good if somebody else can have another look. 

PS: If we really end up using this we owe the developers a postcard 🙂: https://github.com/spatie/calendar-links#postcardware


includes/parse_md.php
```diff
@@ -8,8 +8,9 @@
 require_once('functions.php');
 
 // Markdown parsing libraries
-require_once(dirname(__FILE__).'/libraries/parsedown/Parsedown.php');
-require_once(dirname(__FILE__).'/libraries/parsedown-extra/ParsedownExtra.php');
+require "../vendor/autoload.php";
```
ewels, line 11: Do you need this again as it's already been done in `functions.php`?

ewels (approved):
Awesome! Composer is another one of those things I've heard about but never actually sat down to figure out. This looks super nice, definitely a good improvement 👍🏻

Calendar additions look fab too. Maybe reword the download link to _Download calendar file_ or something? Don't love that, but to make it a bit more explicitly calendar related. _Download .ics file_?

mashehu:
How about this solution:
Screenshot 2021-07-07 at 21 10 20

includes/parse_md.php
mashehu: thanks, I fixed it.

public_html/events.php
```diff
@@ -19,10 +20,35 @@
...
+  $link = Link::create($event['title'], $start, $end)
+    ->description($event['subtitle'] ? $event['subtitle'] : "")
+    ->address($address);
+
+  $event_download_button =  '<div class="dropdown btn-group" role="group">
+          <button   type="button" class="btn '.$button_style.' dropdown-toggle" href="#" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+            <i class="far fa-calendar-plus mr-1"></i> Export event
+          </button>
+          <div class="dropdown-menu">
+            <a class="dropdown-item" href="' . $link->ics() . '" target="_blank"> Download as iCalendar</a>
```
mashehu, line 39:
```suggestion
            <a class="dropdown-item" href="' . $link->ics() . '" target="_blank"> Download iCal Event</a> 
```

PR #859, opened 2021-09-01
update czi and aws acknowledgments

mashehu:
I updated the about page a bit (details should be filled in by somebody else 🙂) and added them to the footer (as usual horizontal alignment of logos and text is not trivial 😠):
Screenshot 2021-09-01 at 16 03 22


apeltzer (approved):
Looks good when testing locally - can still adapt once merged 👍🏻

ewels:
On my phone so tricky to see but LGTM from here 👍🏻

Would also like a bigger section about this on the homepage under the contributor logos if possible, and yeah need to rework the text a bit in the about page (probably also the headings etc, maybe even a dedicated page..?) but that can be in a separate PR ✅

Great stuff, thanks!

Phil

PR #864, opened 2021-09-07
Use include paths relative to script that's doing the importing

ewels:
The tools documentation was giving a HTML 500 error. I think that this is because `include` and `require` statements use paths relative to the script being executed. So when the directory depth was different, the Composer autoload path broke.

I tried on the live server and this fixed stuff anyway, so hopefully should be happy now.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1,9 +1,10 @@
 <?php
 // Common PHP functions for the website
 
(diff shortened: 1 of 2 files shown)


ewels:
ok self merging as I actually already applied this change on the live server anyway 🤫

apeltzer:
🤠 😆

PR #868, opened 2021-09-09
Pipeline health page updates

ewels:
* Remove social preview check
* Avoid PHP warnings about missing releases.
* Check for JSON schema and DSL2 modules dir.
* Don't fetch & cache the repo webpage.


drpatelh (approved):
🚀

PR #870, opened 2021-09-15
Updated with bilibili and figshare links

renbot-bio:


ggabernet (approved):
Thanks for adding this!

PR #877, opened 2021-09-22
Update adding_modules.md reflecting the new `versions.yml` 

grst:
See also  nf-core/modules#665 and nf-core/tools#1247


drpatelh (approved):
👌🏽

PR #879, opened 2021-09-28
Add ALICE profile

bbartholdy:
Add ALICE cluster at Leiden University to the list of institutions


mashehu (approved):

bbartholdy:
Forgot to mention that the white logo also has some black on it (thought it should be ok depending on what is used for)

maxulysse (changes requested):
White logo should only be white.
It's not a black and white logo, but more a monochrome logo actually.

ewels:
Thanks for this @bbartholdy! As @maxulysse says, the white logo needs to be only white (it's used in dark mode when the website background is near-black, plus on dark images etc). The colour logo here is vector and monochrome already, so if you can just edit it in Inkscape / illustrator and make it all white then that would probably be perfect ✅

ewels:
..or if you're not comfortable with editors, I suspect that you can just duplicate your SVG here and change `.cls-2{fill:#001158;}` to `.cls-2{fill:#ffffff;}` and that'll probably be right..

PR #897, opened 2021-10-15
Info groups hackathon

ggabernet:
Hi,

here a first draft for adding more info for the hackathon event.

Mostly for @drpatelh , @d4straub, @maxulysse, @jfy133, @FriederikeHanssen to double check that the info about their leading groups is ok. I've put the groups into more thematic groups to include a couple or 3 pipelines that might be sharing modules together in each of the groups. 

Also the link to the HackMD slides needs to be updated once we have the real link.

The Schedule is still missing as I realized it was an html tab, but will find some time to update it soon as well.


ggabernet:
> Also @ggabernet the slides are here: https://hackmd.io/team/nf-core?nav=overview
> 
> But currently I structured as all pipelines in one slide deck (as per previous hackathons). D you want me to separate them?

Ah no, that's fine, I will just fix it and add this link. :)

PR #899, opened 2021-10-16
Adds files for bingo firebase webapp

jfy133:
Originally from https://github.com/cordeliadillon/bingo modified the react webapp here for nf-core hackathons.

Normally will be rendered in nfcore-bingo.web.app

Dumped in `external` for the time being as I don't know where it would go, so just let me know where you would prefer it to be.


jfy133:
Probably not worth the effort! Closing

PR #931, opened 2021-10-30
Update stats - bugfix Composer, remove core repo 404 cache.

ewels:
Critical fix for this script, which was silently dying when it got to the twitter section, meaning that the new stats results were never saved.

* Fix require statement when using Composer for Twitter
* Remove core repos from the cache if GitHub returns a 404
* Debug print statements cleanup


ewels:
Self merging as late. Have tested on the live server, will test once more once deployed.

PR #936, opened 2021-11-01
Fix laggy rendering of contributing institutes on the homepage

ewels:
It's been bugging me for a while that when scrolling to the end of the homepage the browser becomes almost unresponsive and lags badly. After a bit of investigation I think it was the fading in and our of the SVGs that was difficult for the browser. Plus _all_ contributor images were loaded at page load, just some were hidden.

To speed up rendering, this PR only loads 8 images at any given time and drops the fade in and out. Instead it animates margin (fast and easy) and side-scrolls images using CSS `overflow: hidden`. In my testing it is much smoother.

### DIFF
--- a/public_html/assets/js/nf-core.js
+++ b/public_html/assets/js/nf-core.js
@@ -94,34 +94,19 @@ $(function () {
     }, 2000);
   }
   function switch_contrib_img() {
-    // Reset if all images have been shown
-    if ($(".homepage_contrib_logos a:hidden:not(.contrib_shown)").length == 0) {
-      $(".homepage_contrib_logos a").removeClass("contrib_shown");
-      $(".homepage_contrib_logos a:visible").addClass("contrib_shown");
-    }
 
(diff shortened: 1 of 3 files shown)


mashehu (approved):
Looks good to me. But the animation feels a bit off to me, how about if we make it a non-stop slowly crawling band of logos?

ewels:
Yes that was my initial plan but I was rushing as dinner was nearly ready and this was easier 😆

ewels:
The difficulty is that the logos are all different widths. So knowing the speed / time to switch the images seemed non-trivial to me. This animation only depends on the one image being removed so was simpler to wrap my head around.

ewels:
Ok, I'm fairly happy with this as it is and it fixes quite an annoying bug, so I'm going to merge if that's ok as you've given it a ✅ 

We can take the continuous scrolling thing as a separate PR another day.

PR #937, opened 2021-11-02
Update homepage cards

ewels:
Updated the feature cards on the homepage to give less prominence to Docker / Singularity / Bioconda and add reference to portability and cloud instead.

image

### DIFF
--- a/public_html/index.php
+++ b/public_html/index.php
@@ -263,11 +263,12 @@ include('../includes/header.php');
       <div class="card h-100 shadow-sm">
         <div class="card-body d-flex align-items-center">
           <div>
-            <h5 class="card-title">Docker</h5>
-            <p class="card-text">Software dependencies are handled with docker containers
-              which Nextflow downloads for you, so no need for any software installations.</p>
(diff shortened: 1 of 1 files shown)


maxulysse (approved):

mashehu:
One thing which stings a bit in my eye... Can we make the "Stable Releases" check mark in nf-core green and maybe use a different color for the parcel in packaged software e.g. the github-blue we have in the CI Testing graph (i know it is the info color we use for the rest of the theme, but in this context it clashes a bit)?

ewels:
To be honest, I'm not loving the container box so have been half wondering if we should revert to the Docker whale anyway.. More immediately recognisable and prettier to boot.

Less fussed about the green, but happy to change as long as we're careful. I'd want to be careful that it doesn't become too dominant as it's quite a big block of colour and the nf-core green is more saturated / darker. But probably just requires a bit of trial and error.

mashehu:
How about something like this?

Screenshot 2021-11-10 at 13 45 00

PR #938, opened 2021-11-02
Start working on homepage demo tabs

ewels:
Started playing with using self-hosted [asciicinema](https://asciinema.org/) demos on the homepage instead of static code blocks. I think it looks quite nice!

Needs some styling fixes and for all of the demos to be recorded, but thought it was worth pushing a proof of concept for some feedback.

https://user-images.githubusercontent.com/465550/139961184-4c391f25-e8c4-4008-9f98-0ad29d0135f5.mov


mashehu:
I like the idea. But can't we can instead simulate this in Javascript, so the text is still copy-paste-able?

maxulysse:
Love it, but we need to keep this up to date.
Can we automatize that with every release?

ewels:
> I like the idea. ~But can't we can instead simulate this in Javascript, so the text is still copy-paste-able?~ just saw that it is actually like that already. Nevermind 🙃

Yeah it's already selectable, pretty nice. Can also mess with the styling and stuff which is how I got it to blend in.

I originally wanted to do an interactive demo, as with https://clusterflow.io, but that's a lot more work so thought this might be more realistic instead.

> Can we automatize that with every release?

Short answer: No. Longer answer: Maybe but probably still no.

I think that it's too complex to do any serious update. I'd rather keep the demo(s) as simple as possible so that it's easy to manually repeat them if they get too out of date. It's very easy to do the recordings.

mashehu:
I think I found a way to automate it quite nicely: asciinema + [demo-magic](https://github.com/paxtonhare/demo-magic). Calling `asciinema rec -t "test" -c "./test_script.sh -n"` with the following test_script.sh:
```
#!/bin/bash

########################
# include the magic
########################
. demo-magic/demo-magic.sh

# hide the evidence
clear

# Put your stuff here
pe "nextflow run nf-core/fetchngs -profile docker,test"
```
generated: 
[asciicast](https://asciinema.org/a/448005)

mashehu:
The GH action for it could look similar to this https://github.com/benthayer/git-gud/blob/main/.github/workflows/generate-preview.yml Triggered for example with every new tools release or of any other pipeline we use in the demo.

ewels:
Ooooh 🤩  I stand corrected... 🙇🏻 

I confess that I did manually edit some of the recordings in this PR to remove ugly warnings from the test profile and stuff. But in return for automation I'd be happy to leave them in.

We could do similar stuff all over - mostly thinking of the tools docs (which also has copy+pasted excerpts that get outdated fast).

mashehu:
Demo-magic let's you quite easily fake in- and output, so we don't have to run the actual commands of they give weird warnings. 

Which commands did you have in mind for the tools docs? Because in my opinion people reading it should a bit more computational experience, so should be able to handle text only docs, no?

ewels:
Yeah true. A few bits have copy pasted terminal examples but you're right, this is probably overkill.

maxulysse:

ewels:
I think we should do these manually using [Screen Studio](https://www.screen.studio/) so they're all fancy, then make this one of the primary elements right at the top of the page.

Leaving this PR open as a placeholder for design concepts, will come back to it and close / reimplement at a later date.

mashehu:
Ah, I thought you were talking about having this videos somewhere in the docs. Doesn't need to be automated for the mainpage imo (but maybe we hide the  version number, so it is less obvious?)

ewels:
ok let's move this to https://github.com/nf-core/nf-co.re/issues/1422 instead and close here.

PR #939, opened 2021-11-05
move icons of the anchor links to the right

mashehu:
As discussed on slack:

Screenshot 2021-11-05 at 12 14 51

(as shown the vertical centering isn't the best for some 😞)

I tried to replace as many headings as possible with the php `_h()` function, without changing too many links. Only exception (because I think there should be rather minor effects): 
- #{current,past}_events -> #{current,past}-events
- #future_events -> #upcoming-events

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -331,8 +331,8 @@ function _h($level, $html, $toc_hidden = false)
   $toc_hidden_class = $toc_hidden ? 'toc-hidden' : '';
   return '
     <h' . $level . ' id="' . $hid . '" class="' . $toc_hidden_class . '">
-      <a href="#' . $hid . '" class="header-link"><span class="fas fa-link"></span></a>
       ' . $html . '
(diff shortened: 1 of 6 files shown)


ewels (approved):
Nice 👍🏻 

I guess some of the diffs here are from an auto-formatter that you're using? What is it? Should we implement it globally and add to the CI?

ewels:
Ack, merge conflict because of the masonry card fix..

mashehu:
> Nice 👍🏻
> 
> I guess some of the diffs here are from an auto-formatter that you're using? What is it? Should we implement it globally and add to the CI?

Argh, sorry for that, this one slipped through. I use [intelephense](https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client), but I found now a setting to just apply the formatting on modifications, so it should not happen anymore 🤞🏻

PR #940, opened 2021-11-08
Update module docs: efficiently handle compressed files

grst:
I've seen all sorts of custom groovy code dealing with compressed files in modules. Most of them could be handled by a simple `zcat -f` with the additional benefit of never writing the uncompressed files to disk. 

Even the slightly more complicated cases can be handled by named unix pipes most of the time.

### DIFF
--- a/markdown/developers/adding_modules.md
+++ b/markdown/developers/adding_modules.md
@@ -305,6 +305,28 @@ using a combination of `bwa` and `samtools` to output a BAM file instead of a SA
   - `*.fastq.gz` and NOT `*.fastq`
   - `*.bam` and NOT `*.sam`
 
+  If a tool does not support compressed input natively, we RECOMMEND passing the
+  uncompressed data via unix pipes, such that it never gets written to disk, .e.g
+
+  ```bash
+  zcat -f $input | tool > $output
+  ```
+
+  The `-f` option makes `zcat` auto-detect if the input is compressed or not.
+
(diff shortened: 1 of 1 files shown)


drpatelh:
Nice! Have we got a feel for which modules are rubbing your pet peeve up the wrong way?

drpatelh (approved):

rpetit3:
Many of my recent ones have used groovy logic. I too would prefer these suggestions and adding this to the docs will be super helpful for guidance

@grst  I'll start reviewing them and see if `zcat -f` or named pipes will work.

charles-plessy:
Nice! Note that in order for it to work on Mac OS, it has to be `zcat -f < $input`.  This is because `zcat` on Macs still expects the old `.Z` file suffix that was used at the time of the `compress` command.

grst:
@charles-plessy, good catch! 

Would `gzip -cdf` behave more consistently on mac? On linux it's the exact equivalent of zcat.

charles-plessy:
> Would `gzip -cdf` behave more consistently on mac? On linux it's the exact equivalent of zcat.

Nice! It works well on my Mac and the intent is much clearer than with `zcat -f <`.

JoseEspinosa (approved):
Beautiful! :rocket: 
And with an extra tip of how to make it run in mac, did not know that trick :nerd_face:

PR #942, opened 2021-11-09
fix twitter testimonials, change scrollspy initialisation

mashehu:


ewels (approved):
Removed orphaned `.masonry` CSS, then LGTM ✅

PR #949, opened 2021-11-10
fix highlightjs warning about html tags inside code blocks.

mashehu:
Not the most straightforward solution maybe, but it seems to do the job 😉

### DIFF
--- a/includes/parse_md.php
+++ b/includes/parse_md.php
@@ -103,8 +103,13 @@ function parse_md($markdown){
 
 
   // Prohibit breaking up `--` in option names
-  $content = preg_replace_callback('/([^\w])(--\w+\S*)/i', function($match){
+  $content = preg_replace_callback('/([^\w])(--\w+\S*)|<pre>(?:(?!<pre\s?>).)*/i', function($match){
+    if($match[1]!=">"){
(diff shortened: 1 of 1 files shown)


ewels (commented):
This regex is making my head 🤯 a little. Doesn't the `|` mean OR, so the new section is essentially totally separate from the first? How does this work?

mashehu:
I tried to find a regex to not catch things inside a `<pre>` tag, so I explicitly catch these elements, but then ignore them with the if clause. Is there a more php-typical way to do this?

ewels:
So the regex doesn't look inside the `<pre>` tag because it already matched the entire thing? Smart!

Not sure off the top of my head how I would do it. Probably in a messier way than this. If it works then I'm happy 👍🏻 (maybe add a bit more verbose comment explaining how it works for when I forget this conversation).

ewels (approved):

mashehu:
I tested it a bit more and it actually doesn't work like expected in all cases. I simplified it now, so it uses the old regex, but keeps checking if it is immediately inside a tag, which excludes them inside the code blocks (if they are not a string) or also inside headers, which is currently breaking some pages. To fix these I will merge this now, but maybe come back with a better solution 🤞🏻

ewels (approved):
LGTM 👍🏻

PR #966, opened 2021-11-19
Update tools cmds in synchronization" and "create modules" pages

JoseEspinosa:
Update some nf-core commands to the last version in "synchronization" and "create modules" pages. (`--tools` option in modules has been deprecated + the directory has to be provided with the `--dir` flag.


JoseEspinosa:
After your comment, I checked the commands again and I realized that `--dir` is actually not needed, so I deleted it from the commands. Actually, I also detected that I did not copy the whole command for modules `nf-core modules fastqc` instead of `nf-core modules create fastqc` 😅

PR #972, opened 2021-11-30
hotfix: redirect lint errors to correct page (always linking to the latest version)

mashehu:


ewels (approved):
Good spot 👍🏻 Need to finish this stuff up at some point... (probably with a significant rewrite in tools itself)

PR #976, opened 2021-12-02
adding function to generate logos for dark or light backgrounds using `?theme`

mashehu:
Related: https://github.com/nf-core/logos/pull/23

Giving no `?theme` parameter assumes `theme='light'` and generates `nfcore-pipelinename_logo_light.png`. Should we instead handle an empty  parameter as empty and generate `nfcore-pipelinename_logo.png` (with light mode values) to have some backwards compatibility?

I also softened the white and the black a tiny bit (as well as the stem of the apple for the dark mode).

After we agree that this works, the following things need to be taken care of:

- [x] document the `?theme` option in the https://github.com/nf-core/logos readme and maybe also in the graphical guidelines on the website.
- [x] update the pipline readme in the template


public_html/logo.php
```diff
@@ -14,7 +14,12 @@
     $new_width = 400;
 }
 
-$filename = 'nfcore-'.preg_replace("/[^a-z]/", '', $textstring).'_logo.png';
+$theme = 'light';
+if (isset($_GET['theme'])) {
+    $theme = $_GET['theme'];
+};
```
ewels, line 20:
```suggestion
if (isset($_GET['theme']) && $_GET['theme'] == 'dark') {
    $theme = 'dark';
};
```

ewels (approved):
Nice, this is less code than I imagined 👍🏻

I wouldn't be sad if we keep backwards comparability with the filenames, but it does add bloat. If we drop it then we need to try to time with a tools update / release.

Maybe maintain the filename for now with a comment that it's deprecated and can be removed at a later date?

PR #978, opened 2021-12-06
Fix background colour for images in join page

ewels:
Currently:

image

image

Fix untested, just took the variable from the [`body`](https://github.com/nf-core/nf-co.re/blob/16ad29476c2d07bfc333c3fa04b1c50aa3efce2d/public_html/assets/scss/_nf-core.scss#L16) style.


mashehu (approved):
tested it locally. looks good.

PR #989, opened 2022-01-06
make `background-color` of .bg-light  different from .bg-body

mashehu:
Did it mainly so that `.bagde.bg-light` is actually useful also in the light theme when used outside of cards (and to get one more background color to use 😉).
Might need to think about, if some of the places we use it should actually be `.bg-body`.
Screenshot 2022-01-06 at 13 06 12
Screenshot 2022-01-06 at 14 13 10

### DIFF
--- a/includes/pipeline_page/pipeline_releases_stats.php
+++ b/includes/pipeline_page/pipeline_releases_stats.php
@@ -69,28 +69,28 @@ $(function(){
 
 <h1 id="stats" class="mt-5"><a href="#stats" class="header-link"><span class="fas fa-link" aria-hidden="true"></span></a> Pipeline Statistics</h1>
 <div class="card-group text-center stats_keynumbers mt-5 pb-5">
-  <div class="card bg-light">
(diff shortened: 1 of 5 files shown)


ewels:
What are the screenshots showing sorry? The first looks the same as the current site to me.. For the second screenshot I'm not so keen on the darker backgrounds. I wonder if we should just remove the `bg-light` class from these elements? I guess I added it when the page background was white to give a little contrast. Now that the page background is not white reverting to their defaults (white) gives equivalent contrast and looks good I think.

mashehu:
Sorry, should have added a before after, for this tiny change in the badges (they are now a bit darker).
Before: 
Screenshot 2022-01-12 at 11 44 32
After:
Screenshot 2022-01-06 at 13 06 12

I agree with your second point, and changed them now.

ewels (commented):
Super nice 🚀 

Shaun The Sheep

ewels (approved):

PR #1001, opened 2022-01-21
fix printing layout for pipeline pages

mashehu:
Kicked off by the comment by @pcantalupo on Slack, I had a quick go on the print layout for the pipeline pages:

- adds a dependency-library, which provides some basic improvements
- hides the more interactive parts (buttons, ToCs)
- uncollapses everything by default (help texts and hidden params)

Looks already very nice, but I am not sure how we should handle `.table-responive` elements, because they are too wide for a page (see for example https://nf-co.re/eager/usage#tsv-input-method).


ewels (approved):
Great stuff! Not too fussed about exceptional super wide tables, as long as we cater for most (eg. parameters etc)

Happy to merge as-is, haven't tested but code looks good and non-destructive.

PR #1003, opened 2022-01-24
Update bytesize-32-nf-core-dualrnaseq.md

renbot-bio:


mashehu (changes requested):
You also need to change the file name

mashehu (approved):

PR #1010, opened 2022-02-02
Add gitpod setup

mashehu:


mashehu:
Setup should be quite simple, but for some reason Apache is throwing a file permission error: https://mashehu-nfcore-ymt9am34bef.ws-eu30.gitpod.io

mashehu:
Probably good to squash the commits on merge, because it is was a bit of trial and error.

ewels (approved):
Awesome work! 👏🏻 

Do we want Prettier installed in vscode?

Also, do we want to run the commands that update the pipeline list etc? Might be faster to make these cache files publicly available on the main website, then we could just `curl` them..

mashehu:
agree on prettier.
Not sure about the json files. Making them public on the webite+curl is probably the easiest. We could also run it in the `init` step to do this as a prebuild, but this might be a bit complicated with the github credentials... 
Can we do it in a way which future proofs this, when we move more and more things into a database?

mashehu:
I had to add an additional condition to the https-redirect, so that the gitpod preview doesn't get redirected to the actual url. Can you please have a look @ewels that we don't loose some edge cases with the given url pattern.

ewels:
Redirect looks good, but might be sensible to be at the keyboard when we merge in case we need a quick revert if it breaks.

.gitpod.yml
```diff
@@ -0,0 +1,33 @@
+---
+# based on https://github.com/gitpod-io/template-docker-compose
+
+tasks:
+  - init: |
+      docker-compose pull
+  - command: |
+      chmod a+x /workspace/nf-co.re && docker-compose up
```
ewels, line 8:
```suggestion
  - init: docker-compose pull
  - command: chmod a+x /workspace/nf-co.re && docker-compose up
```

ewels (approved):
Minor nitpick

PR #1012, opened 2022-02-04
Update module `when:` guidelines

mahesh-panchal:
Updates module guidelines for using `when:` after addition of https://github.com/nf-core/tools/pull/1393


maxulysse (approved):
<3

PR #1027, opened 2022-02-15
add social event

mashehu:
@ggabernet could you please double check if all the times and items are correct? It should render nicely in gitpod.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1027"></a>


ggabernet (approved):
Yes, perfect, the events and the times are right!

I love the GitPod preview!

PR #1030, opened 2022-02-23
Some fun side activity for the hackathon

mashehu:
🚧 First very hacky version 🚧

Based on https://www.nytimes.com/games/wordle/index.html and a clone implementation from https://github.com/WebDevSimplified/wordle-clone

- pipeline wordle runs for all pipelines names (even >5 letters). To make it a bit easier it allows also entering shorter pipeline names, to get more hints.

- started with module version

Known bug: If a letter is entered twice and one of the instances is actually on the right position, the incorrect occurrence does still flip to orange (for example when the right word is "chipseq" and I enter "cageseq" both "e"s are orange), because each letter is tested individually without knowledge about the others. I would be thankful, if somebody knows how to solve this elegantly.

Screenshot 2022-02-23 at 11 47 17

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1030"></a>


mashehu:
I removed  the modules section and moved everything in one file. 

Copies the emoji summary of the guesses at the end of the tries (=win or lose), e.g.:
nf-cordle 1
🟨🟨⬛️🟨🟨⬛️
🟨🟨🟨🟨⬛️⬛️
🟩🟩🟩🟩🟩🟩

New pipeline name available every 12 hours (3am and 15pm CET for now)

Doesn't save the state in a cookie, so you have multiple tries and no state is saved...

mashehu:
I guess the easiest to test it would be to deploy it on the live server, so that people from the outreach team can play with it.

ewels (approved):

PR #1051, opened 2022-03-07
Remove remaining references to suffix and output in modules guidelines

jfy133:


markdown/developers/adding_modules.md
```diff
@@ -300,7 +300,7 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
 
 ### General
 
-- All non-mandatory command-line tool options MUST be provided as a string via the `$args` variable, which is assigned to using the `task.ext.args` variable. The value of `task.ext.args` is supplied from the `modules.config` file by assigning a string value to `ext.args`.
+- All non-mandatory command-line tool arguments MUST be provided as a string via the `$args` variable, which is assigned to using the `task.ext.args` variable. The value of `task.ext.args` is supplied from the `modules.config` file by assigning a string value to `ext.args`.
```
jfy133, line 303:
```suggestion
- All non-mandatory command-line tool non-file arguments MUST be provided as a string via the `$args` variable, which is assigned to using the `task.ext.args` variable. The value of `task.ext.args` is supplied from the `modules.config` file by assigning a string value to `ext.args`.
```

mahesh-panchal (approved):

mahesh-panchal:
It would be nice to change the `if` error message too, but I'm not sure what should go there. Can make another pull request later

jfy133:
> It would be nice to change the `if` error message too, but I'm not sure what should go there. Can make another pull request later

I think also there is a bit of inconsistency now in modules, would be worth going back through thos at the same time

PR #1060, opened 2022-03-09
bug fixes for nf-cordle

mashehu:
- add instructions modal (modeled after the original one from wordle)
Screenshot 2022-03-09 at 15 21 15

- fix colors in light mode
- cookie expires now before a new word is available
- allow only valid pipeline names

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1060"></a>


ewels (approved):
Nice!

PR #1078, opened 2022-03-16
Docs reorganization

mashehu:
- merge nav elements into Docs
- switch `developers` to `contributing`

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1078"></a>


ewels:
Right, about this - I was just thinking as I stood in front of the microwave. Everyone is working on these docs files at the moment, if we move them it's going to be a merge-conflict party 😰 

Maybe it doesn't matter as we can just overwrite with the latest state, but if we want to merge during the hackathon it will be a pain.

What do you think about doing softlinks instead in the short term? Then once activity has died down a bit we can move them properly and put in some `.htaccess` redirects?

mashehu:
Finally, got all features I wanted running: 
Screenshot 2022-03-23 at 13 23 11
Doesn't include the docs yet, because I wanted to wait for https://github.com/nf-core/tools/pull/1458 to be merged. But it shouldn't be too difficult to include them in the current setup (I imagine a second parsing function which adds tools sections based on the yaml-header of the markdown files to the `$sidebar_nav_elements` array)

ewels:
I get a 404 when I try to view any docs 🤔 eg. http://localhost:8888/contributing

includes/footer.php
```diff
@@ -87,15 +87,16 @@
       <div class="col-sm-6 col-lg-3 mb-3 d-print-none">
         <h5>For Authors</h5>
         <ul class="list-unstyled">
-          <li><a href="/developers/guidelines">Guidelines</a></li>
-          <li><a href="/developers/adding_pipelines">Adding a new pipeline</a></li>
-          <li><a href="/developers/modules">Adding a new module</a></li>
-          <li><a href="/developers/release_checklist">Release checklist</a></li>
+          <li><a href="/contributing/guidelines">Guidelines</a></li>
```
ewels, line 90: Were we not going to prefix these URLs with `docs/`?

ewels:
ok so http://localhost:8888/usage/installation works but not http://localhost:8888/usage (which is where the main nav takes me).

http://localhost:8888/developers/adding_pipelines works but not http://localhost:8888/contributing/adding_pipelines (or http://localhost:8888/contributing/)

ewels:
ok so the new softlink makes http://localhost:8888/contributing/adding_pipelines work now. But I still get no new nav sidebar 🤔

mashehu:
> ok so the new softlink makes http://localhost:8888/contributing/adding_pipelines work now. But I still get no new nav sidebar 🤔

yeah, that's a bit due to the rerouting and the soft links. Only `nf-co.re/docs/` get rerouted through `documentation.php` and get a sidebar, the others are just rendered as normal markdown files.

ewels:
ahh getting better - that nav dropdown animation 🤩

ewels:
What do you think about bumping up the content width? It's getting a bit narrow now that we have the two sidebars.. Maybe time to start thinking about a fluid content container? Or at least a bigger maximum width?

image

mashehu:
yeah, wasn't sure how to get it wider we are already on xxl (didn't think about `.container-fluid` 🙃 ). but it is also too narrow for my taste.

includes/header.php
```diff
@@ -152,31 +152,10 @@ function gtag() {
...
-              <a class="dropdown-item" href="/developers/adding_pipelines">Adding a new pipeline</a>
-              <a class="dropdown-item" href="/developers/modules">Modules</a>
-              <a class="dropdown-item" href="/developers/release_checklist">Release checklist</a>
-              <a class="dropdown-item" href="/tools/docs/latest/pipeline_lint_tests/">Lint error codes</a>
-              <a class="dropdown-item" href="/developers/sync">Template synchronisation</a>
-              <a class="dropdown-item" href="/developers/developer_tutorials">Developer tutorials</a>
-              <a class="dropdown-item" href="/developers/editor_plugins">Code editor plugins</a>
-              <a class="dropdown-item" href="/developers/design_guidelines">Graphic design guidelines</a>
+              <a class="dropdown-item" href="/docs/usage/introduction">Usage</a>
+              <a class="dropdown-item" href="/docs/contributing/guidelines">Contributing</a>
```
ewels, line 158: I'm not sure that this really adds much now.. It doesn't save a click most of the time. I think it's probably better to just have a flat link to the docs?

ewels: Ideally a new page with some kind of overview, but for now could be `/docs/usage/introduction` I guess

PR #1090, opened 2022-03-17
Module tutorials

### LINKED ISSUE #1040: Remove duplicate new DSL2 module tutorials
There are currently two similar tutorials / docs pages on how to create a DSL2 module:
- Developers --> [adding a new module](https://nf-co.re/developers/adding_modules)
- Developers --> developer tutorials --> [Create a DSL2 module](https://nf-co.re/developers/tutorials/dsl2_modules_tutorial).

A merged tutorial on how to add a new module. Probably Developers --> Adding a new module is easier to find, so I would propose to merge all the information there, but up for discussion!

Emiller88:
Closes #1040

I just went with the latest written tutorial based on the git history. I think it'll need to be updated as well.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1090"></a>


Emiller88:
@ggabernet You should be able to view them in Gitpod now! I can't figure out who wrote the two tutorials, they might be able to merge the two better. @lescai added the images for the first one, and I think maybe @jfy133 wrote the second judging by the git blame?

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
...
-```
+---
 
-Our final script will therefore look like this:
+## Writing a new module
 
-```nextflow
-// Import generic module functions
-include { initOptions; saveFiles; getSoftwareName } from './functions'
+Check out [Tutorial: Create a DSL2 Module](https://nf-co.re/developers/tutorials/dsl2_modules_tutorial) for how to create a new DSL2 Module
```
lescai, line 45: the link points to the old one but I suppose that's the result of the old page being up and the new not rendered yet, so on the gitpod version we open that one.
however, when this page is rendered, wouldn't this link point to itself?

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
...
 
-    conda (params.enable_conda ? "bioconda::fgbio=1.3.0" : null)
-    if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
-        container "https://depot.galaxyproject.org/singularity/fgbio:1.3.0--0"
-    } else {
-        container "quay.io/biocontainers/fgbio:1.3.0--0"
-    }
+2. Install [`Nextflow`](https://www.nextflow.io/docs/latest/getstarted.html#installation) (`>=21.04.0`)
+3. Install any of [`Docker`](https://docs.docker.com/engine/installation/), [`Singularity`](https://www.sylabs.io/guides/3.0/user-guide/) or [`Conda`](https://conda.io/miniconda.html)
+4. [Fork and clone the nf-core/modules repo locally](#uploading-to-nf-coremodules)
```
lescai, line 57: I understand this is meant to make the checklist leaner, but a tutorial in my view should be a step-by-step. and if you fork before start coding, people should get the instructions about forking before the instructions about the code.
in this way, they're linked to content that's placed after the guidelines on testing.
the order of the description therefore doesn't follow the order of actions that should be taken.

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
...
-    tuple val(meta), path("*_umi_converted.bam"), emit: umibam
-    path "*.version.txt"          , emit: version
+   Make a new branch for your module and check it out
 
-    script:
-    def software = getSoftwareName(task.process)
-    def prefix   = options.suffix ? "${meta.id}${options.suffix}" : "${meta.id}"
+   ```bash
+   git checkout -b fastqc
+   ```
```
lescai, line 68: not particularly attached to previous screenshots, but there's 2 ways of creating a branch and one is through GitHub itself on the browser. We are presenting only one of the 2 options here.
This is usually a personal preference and I believe a tutorial should give people the option to choose.

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
 
 ### Module guidelines
 
-The nf-core community has agreed on a minimal set of [guidelines](https://github.com/nf-core/modules#guidelines), intended to make module most suitable for general use, i.e. to be shared across a wide variety of community workflows.
+The nf-core community has agreed on a minimal set of [guidelines](https://nf-co.re/developers/modules#guidelines), intended to make module most suitable for general use, i.e. to be shared across a wide variety of community workflows.
```
lescai, line 14: I cannot add a comment where there isn't a change but the paragraph before introduction says the tutorial uses as use-case the FGBIO module (previous tutorial), this one seems to be using FASTQC

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
...
-mkdir tmpFolder
-
-fgbio --tmp-dir=${PWD}/tmpFolder \\
-FastqToBam \\
--i ${reads} \\
--o "${prefix}_umi_converted.bam" \\
---read-structures $read_structure \\
---sample ${meta.id} \\
---library ${meta.id} ${options.args}
-```
```
lescai, line 156: Again, I understand the goal is to make the page lean but this tutorial basically links to existing code of the FASTQC modules: a novice should follow the links and read the code of the entire script, making a connection between the instructions on this page and a ready-to-use finalised code.
Personally, I believe explaining the code bit by bit is part of the goal of a tutorial, although I understand it takes a bit of effort to write (and maintain).

markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -11,414 +11,298 @@ If you create a new module with the goal of contributing the code to *nf-core*,
...
-As described above, *nf-core tools* has created already the following files
+    4. [`./tests/modules/fastqc/nextflow.config`](https://github.com/nf-core/modules/blob/master/tests/modules/amps/nextflow.config)
 
-```bash
-tests/modules/fgbio
-└── fastqtobam
-    ├── main.nf
-    └── test.yml
-```
+        Some modules MAY require additional parameters added to the test command to successfully run. These can be specified with an `ext.args` variable within the process scope of the `nextflow.config` file that exists alongside the test files themselves (and is automatically loaded when the test workflow `main.nf` is executed).
```
lescai, line 114: this part is really crucial, and it's probably one of the most nested feature of the new template.
I understand it's well explained in the module guidelines page, but I also think one should go through it here again with an example that refers to the specific use-case that's used for the tutorial here.
We can't take this for granted, and a user without much experience will find this part difficult to navigate through without a concrete example within the context of this page.

lescai (changes requested):
Hi there,
I'm afraid I've had several comments along the page.
In short I believe a tutorial should be as much as possible self-contained, rather than forcing the user to navigate several links to add more details to the explanation and understand how to write the code it should be guided through.
Also more in-page examples with code for the specific use-case should be needed in my view, rather than directing the unexperienced user to a well-written whole script where the different parts are not isolated and explained.

jfy133:
> @ggabernet You should be able to view them in Gitpod now! I can't figure out who wrote the two tutorials, they might be able to merge the two better. @lescai added the images for the first one, and I think maybe @jfy133 wrote the second judging by the git blame?

The general modules one is @drpatelh I think, and I helped out tweaking the DSL2 Modules tutorial.

But I agree with all of @lescai 's points above for the latter - this is meant to be really 'hand-holdy' for new contributors. The former is meant to more be 'specs' and a quick-reference for experienced module writers.

Emiller88:
@lescai Thanks for the review!

@jfy133 Hmm, so should they live in parallel, or is there a way we can combine them? Maybe the reference should live in the CONTRIBUTING file of the modules repo?

lescai:
@Emiller88 I think they are 2 different things, so they should live in parallel.
- The module guidelines indicate what coding style and practice one should adhere to when writing a module for nf-core: can be placed on the website either under modules or under any guideline(s) area.
- A tutorial (any), should present a use-case (in order be applied and not a guideline) and guide someone (hand-hodly to quote @jfy133 :P) on each step they should follow when they do something for the first time.

Emiller88:
Gotcha, I agree with you two. I think the real issue was the page title `adding modules` was confusing, and the developer reference should point new users to the tutorial.

@ggabernet since you made this issue, any thoughts?

jfy133:
> Gotcha, I agree with you two. I think the real issue was the page title `adding modules` was confusing, and the developer reference should point new users to the tutorial.
> 
> @ggabernet since you made this issue, any thoughts?

Ah yeah, `adding_modules` was the original page, and we then later added the dsl2_modules tutorial (but in the tutorials folder, long beofre we discussed this reorganisation)

lescai (commented):
ok main page is fine I think thanks!
the tutorial itself now points to the old one, which is however outdated in the parameters part (at least), i.e. uses the old syntax "addParams" instead of "ext.args"

Emiller88:
Gotcha, do we want to address that here, or in a fresh PR?

My vote is a fresh PR so this doesn't get held up.

Emiller88:
Gotcha, do we want to address that here, or in a fresh PR?

My vote is a fresh PR so this doesn't get held up.

lescai:
sure! makes sense to PR one thing at a time

lescai (approved):

PR #1093, opened 2022-03-17
gitpod config to be used for tutorials

lescai:
This PR is meant to create a specific .gitpod.yml config file, to be used for running tutorials on a gitpod environment.
Reason for branch: the master branch contains a .gitpod.yml that serves an environment needed for the website development.
This PR changes 2 files:
1) .gitpod.yml
simplified and uses the nf-core/gitpod image containing nextflow and nf-core tools
also opens automatically the usage tutorial page
2) .vscode/settings.json
this reconfigs VScode in order not to open the welcome page, and allow full width display of the nf-core tutorial page

A separate PR is going to be submitted in order to insert the gitpod button into the tutorial page.

relates to https://github.com/carpentries-incubator/workflows-nextflow/issues/68


ewels:
See the orphan branch I made for this: https://github.com/nf-core/nf-co.re/tree/gitpod_config_training_orphan

what do you think about that approach, to keep the branch as minimal as possible?

lescai:
done but now it conflicts on three files which I don't have anymore on my fork.
I've followed the suggestions but everything is up to date on my side. 
Any good suggestions?

ewels:
If we're doing an orphan branch then you can't PR to it from here, as there's no shared git history. You can also just delete everything as you did in the last commit though, I guess that works too.

Those conflicting files will have been changed since you forked this branch. You'll need to resolve the conflict locally by accepting that they should be deleted despite the upstream changes.

lescai:
alright should be good now :) thanks!

lescai:
this is now working.
as I mentioned on PR https://github.com/nf-core/nf-co.re/pull/1094 that's connected to this one, the variable passed through the URL conflicts with url of the repo if contains url-characters, therefore the only solution I could think about was making substitutions then parsed in the config.
if there's more elegant suggestions, they're welcome of course.
this is tested and working.

.gitpod.yml
```diff
@@ -1,31 +1,19 @@
...
-    addComment: false
-    # add a "Review in Gitpod" button to the pull request's description (defaults to false)
-    addBadge: true
\ No newline at end of file
+- port: 8080
+  onOpen: ignore
+tasks:
+- init: git checkout gitpod_exercises
+- command: |
+        parsed=`echo $url | tr -s "[" "/" | tr -s "+" "."`
```
mahesh-panchal, line 17: So this parses the string 
```
https://gitpod.io/#url=nf-co+re[usage[tutorials[nf_core_usage_tutorial/https://github.com/nf-core/nf-co.re/blob/gitpod_config_training/
```
Parts of the URL:
```
https://gitpod.io/#                                              // Gitpod prefix
url=nf-co+re[usage[tutorials[nf_core_usage_tutorial              // Value pair, separate by commas
https://github.com/nf-core/nf-co.re/blob/gitpod_config_training/ // Repo + branch to open
```

Does URL encoding work (it says the values are URL encoded)? https://www.urlencoder.org/
```
https://carpentries-incubator.github.io/workflows-nextflow/01-getting-started-with-nextflow/index.html
```
becomes
```
https%3A%2F%2Fcarpentries-incubator.github.io%2Fworkflows-nextflow%2F01-getting-started-with-nextflow%2Findex.html
```
How can I test this? I tried editing the other page https://github.com/nf-core/nf-co.re/pull/1094/files
but then I can't get to the button in the preview browser. 

We may just be over complicating things here though.
What would be nice is just:
teaching_view

So a preview panel with the website (using `gp preview URL` and it's navigable)
And then as large a terminal as possible. I moved the terminal position to the right to have them side by side.
( Menu > View > Appearance > Panel position > right ).
Then they can do all their work in an exercises folder and not have to worry about all the other distractions along the side.

mahesh-panchal: I was too focused on carpentries material before and forgot we have two use cases here. 

For the nf-core training material we want buttons on each page as they are standalone training guides, so we should definitely see if URL encoding is sufficient to open the correct URL with gp preview. 

The Carpentries Workshop though just needs to start from the setup page and can have a simpler link.

lescai:
Hi @mahesh-panchal I'll merge this one so you'll be able to fully test the link on PR https://github.com/nf-core/nf-co.re/pull/1094#issuecomment-1072775128 which relies on the existence of the correct config in this branch to work.
It's been tested on the forked repo but I've had to change repo and branch name of course before submitting it for merging into nf-core so until this branch is merged it won't work :)

PR #1103, opened 2022-03-18
Try to add  `--outdir <OUTDIR>` to example nextflow commands

ewels:
This is now a required parameter and is not set in the `-profile test` config.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1103"></a>


drpatelh (approved):
Thanks!

PR #1108, opened 2022-03-23
add paramter type to parameters page

### LINKED ISSUE #1106: Add parameter type to pipeline Parameters page
The parameter type is defined in the parameter schema but it isn't rendered in the Parameters page on the website. A couple of users have tripped up thinking e.g. the `--trim_nextseq` parameter is boolean when in fact it should be an integer. Be good if this is documented on the website properly.

mashehu:
Open to better ideas on where to place them, instead of just beneath there name:
Screenshot 2022-03-23 at 13 47 16

Closes #1106 
 

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1108"></a>

### DIFF
--- a/includes/pipeline_page/docs_schema.php
+++ b/includes/pipeline_page/docs_schema.php
@@ -44,7 +44,12 @@ if(file_exists($gh_pipeline_schema_fn)){
     if(array_key_exists("description", $param)){
       $description = parse_md($param['description'])['content'];
     }
-
+    # parameter type
+    $type = '';
+    if (array_key_exists("type", $param) && strlen(trim($param['type'])) > 0 && $param['type'] != 'object') {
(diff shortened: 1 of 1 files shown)


ewels:
Bit irritating that we don't have pipelines in GitPod and that the update script doesn't work because of lack of config and stuff. I wonder if we can make that more fault tolerant or something 🤔

ewels:
Anyway, looks good I think 👍🏻 

Only alternative that I can think of is having it as a label on the right side, like `default` and `required`. Gets a bit busy over there though. I think where you have it is better.

Tempting to by more pythonic and have it in brackets after the parameter name - eg. `--input` (str)
But I think that makes it less beginner friendly. So yeah, let's stick with what you've done I think 👍🏻

ewels (approved):
LGTM 👍🏻

PR #1111, opened 2022-03-28
Gitpod tweak prebuild

ewels:
Change prebuild tasks to run in one terminal instead of two.

Before, as the tasks tan in two terminals, the two were running in parallel. By keeping everything in one terminal, `docker-compose up` should now only run _after_ the pull is complete.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1111"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -2,8 +2,10 @@
 # based on https://github.com/gitpod-io/template-docker-compose
 
 tasks:
-  - init: docker-compose pull
-  - command: chmod a+x /workspace/nf-co.re && docker-compose up
+  - init: >
+      docker-compose pull &&
+      chmod a+x /workspace/nf-co.re
+    command: docker-compose up
 
 ports:
   - port: 8888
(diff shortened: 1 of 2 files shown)


ewels:
You now get this if you try to open the GitPod environment too quickly 😅 

image

mashehu (approved):
Because we are already tweaking this, I added option to ignore the db port opening, which was bothering me 🙂

ewels:
Yeah I saw, good catch! I also just added a VSCode settings file, which I have always hated 😱  But I can't see any other way to get Gitpod to play nicely..

mashehu (approved):
Added auto-theming option. Didn't find anyway to disable the docker welcome screen, which is the first thing I have to close everytime...

ewels:
Ah is the Docker welcome screen different to the VSCode welcome screen?

ewels:
I wonder if we even need the docker VSCode plugin installed by default to be honest. Not sure what it actually adds?

Same with GitLens which is also super spammy. If connecting to local VSCode then people can have these installed in their local versions but I wonder if it's better to be more minimalistic in the default setup in the browser..

mashehu:
agree to both.

PR #1118, opened 2022-04-04
Add theme colour to HTML header

ewels:
Added a new meta tag to the header called `theme-color`. The main purpose of this is to tint the top status bar of Safari (only when set to use "Compact Mode" in the preferences).

I wasn't able to use the exact nf-core green as it was too close to the colour for the maximise button. So I darkened it and desaturated it a little. I also tried grey but that is a little generic-feeling.

MacOS Safari 15:

Screenshot 2022-04-04 at 23 44 57

Safari in iOS:

File 04-04-2022, 23 47 57

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1118"></a>

### DIFF
--- a/includes/header.php
+++ b/includes/header.php
@@ -45,7 +45,7 @@ if (isset($subtitle) && strlen($subtitle) > 0) {
   <meta name="description" content="<?php echo $page_meta; ?>">
   <meta name="author" content="Phil Ewels">
   <meta name="color-scheme" content="light dark">
-  <meta name="theme-color" content="#206e43">
+  <meta name="theme-color" content="#1d9655">
   <link rel="shortcut icon" href="/assets/img/logo/nf-core-logo-square.png" type="image/png" />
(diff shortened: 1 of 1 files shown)


includes/header.php
```diff
@@ -45,6 +45,7 @@
   <meta name="description" content="<?php echo $page_meta; ?>">
   <meta name="author" content="Phil Ewels">
   <meta name="color-scheme" content="light dark">
+  <meta name="theme-color" content="#206e43">
```
mashehu, line 48:
```suggestion
  <meta name="theme-color" content="#1d9655">
```

How about the nf-core-color we use in the dark theme instead? 
Screenshot 2022-04-05 at 09 55 42

ewels: Yeah, if that works! I was just doing trial and error trying to find something that safari would load.

PR #1123, opened 2022-04-06
remove sticky top for sidebars in pipeline and modules pages, increase number of elements in ToC dropdown

mashehu:
not sure how to handle the multilevel ToCs in the dropdown. They were too short/not useful for pipeline pages, so I include now h1 and h2, but display them on the same level... 

Don't know a good solution how to handle sticky ToCs with buttons at the bottom. Removed it completely for now for these cases.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1123"></a>


ewels (approved):
Quick merge as it's a hotfix, can discuss further..

PR #1132, opened 2022-04-12
Hopefully this will fix 'git push' on some PRs

ewels:
Not sure if this will work or not, but worth a try.

Based on testing with PR https://github.com/nf-core/nf-co.re/pull/1094

Doing `git push` there fails because the PR comes from a fork branch `master` which is the same as the target. So, GitHub and `hub` create a different branch name for working locally. Then, when you try to push that, it doesn't match the remote and you [get this error](https://github.com/nf-core/nf-co.re/runs/5997784202?check_suite_focus=true):

```
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use
    git push git@github.com:lescai/nf-co.re.git HEAD:master
To push to the branch of the same name on the remote, use
    git push git@github.com:lescai/nf-co.re.git HEAD
To choose either option permanently, see push.default in 'git help config'.
Error: Process completed with exit code 128.
```

According to [this old comment](https://github.com/github/hub/issues/821#issuecomment-255531557), the changes in this PR should hopefully get `git push` to work again 🤞🏻 

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1132"></a>

### DIFF
--- a/.github/workflows/fix-linting.yml
+++ b/.github/workflows/fix-linting.yml
@@ -43,9 +43,13 @@ jobs:
         if: steps.prettier_status.outputs.result == 'fail'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # push.default stuff is because branch names can vary
+        # See https://github.com/github/hub/issues/821#issuecomment-255531557
         run: |
-          git config --global user.email "core@nf-co.re"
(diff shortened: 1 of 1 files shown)


ewels:
I'd quite like to test this before bed. Currently failing on `master` so should be low risk to self-merge 👀

PR #1144, opened 2022-04-29
add back $param_body definition

mashehu:
Got lost during the merge here: https://github.com/nf-core/nf-co.re/commit/4e1f67a72cf960e682b2718fcd8a3c4db3bd2f75

### DIFF
--- a/includes/pipeline_page/docs_schema.php
+++ b/includes/pipeline_page/docs_schema.php
@@ -45,6 +45,7 @@ if (file_exists($gh_pipeline_schema_fn)) {
         if (array_key_exists('description', $param)) {
             $description = parse_md($param['description'])['content'];
         }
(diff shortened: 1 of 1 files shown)


maxulysse (approved):

includes/pipeline_page/docs_schema.php
```diff
@@ -170,6 +171,16 @@ function print_param($is_group, $param_id, $param, $is_required = false) {
             $row_class .= ' param-docs-hidden collapse d-print-block';
         }
 
+        # Body
+        $param_body =
+            '<div id="' .
```
ewels, line 176: Ugh, looks grim when split over a line. Might start using `sprintf` a bit more, now that we're using prettier..

PR #1159, opened 2022-05-13
Rename bytesize-cutandrun_220517 to bytesize-cutandrun.md

ewels:
The website needs the `.md` file extension to find the event.

Also I'm not sure that we need the date suffix in the URL. We already have it in a 2022 folder and I like nice pretty URLs 🤩


FranBonath (approved):
Thanks for figuring this out!

PR #1190, opened 2022-06-06
Add debugging tip for configuration docs

jfy133:
Based  on my own experience just now... adds how to manually find out why a job fails to submit when a job is rejected by a (slurm) cluster

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1190"></a>

### DIFF
--- a/markdown/usage/configuration.md
+++ b/markdown/usage/configuration.md
@@ -167,47 +167,6 @@ Usually you would save configuration such as this to a config file [as described
 
 If you think that the defaults in the pipeline are way off, please let us know! Then we can adjust the defaults to the benefit of all pipeline users.
 
-# Debugging Configuration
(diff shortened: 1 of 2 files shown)


ewels (changes requested):
Actually, I'm not sure that this really belongs in configuration? Feels like it would fit better in troubleshooting? https://nf-co.re/docs/usage/troubleshooting

jfy133:
I was sort of in two minds... troubleshooting is mostly about running the pipeline itself, not working out why a configuration isn't working (I fear troubleshooting will just become a dumping ground), but we can try it. I've added a link to the configuration page to the troubleshooting as a compromise

ewels (approved):
Nice 👌🏻

PR #1209, opened 2022-07-14
Corrected use of single quotes which prevent shell construct evaluation

lquayle88:
The "step_by_step_institutional_profile.md" contained a number of code lines wherein single quotes were used which would have prevented correct evaluation of the shell constructs which they contained. I replaced the single quotes with double quotes on the relevant lines which should allow the code to work as I believe it was intended.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1209"></a>


mashehu (approved):

mashehu:
Thanks for finding and fixing this!

PR #1217, opened 2022-08-01
remove broken broken_link checker (too noisy 😞)

mashehu:
I guess it is time to say good-bye 👋🏻

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1217"></a>


ewels:
Can always add it back again if / when https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/121 is implemented.

Could also look into implementing that as a PR. I actually don't think it should be all that difficult to do. Mostly redirecting the output from the npm `markdown-link-check` package to `/dev/null` and then giving a nicer output when printing from `errors.txt`. It's a fairly short `entrypoint.sh` script in the action.

ewels (approved):

PR #1227, opened 2022-08-11
add copy-button to code-blocks with bash

mashehu:
Aug-11-2022 15-17-20

I could have maybe done it with some html injection in PHP, but javascript was easier...
Only limited to blocks with `.language-bash` for now.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1227"></a>


ewels (approved):
Super nice!

> Only limited to blocks with `.language-bash` for now.

Any reason not to do it for everything?

mashehu:
I don't see a use-case why one would copy want to copy a log output, for example. Or do you think this hinders discoverability?

ewels:
I just don't see a downside really. But yeah I guess log output is not super useful 🤔 And in fairness basically everything is either bash or console / log output.

I'm sure someone will complain soon enough if there are others missing :)

PR #1229, opened 2022-08-12
Fix duplicated tweets

### LINKED ISSUE #610: Check for duplicate tweets before posting
The automated twitter posts about pipeline releases are great, though sometimes releases can get double-postings (often because of something odd at a user-level with the release I think).

It might be nice to do a simple check before posting a tweet that there isn't already a tweet with identical text. This should avoid the simplest duplicate problems. Or perhaps instead of checking the entire tweet, just check the text at the start: `Pipeline release! nf-core/xxxx xxxx`

mashehu:
fixes #610 by fetching all nf-core tweets and searching for matching pipelinename+version.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1229"></a>

### DIFF
--- a/update_pipeline_details.php
+++ b/update_pipeline_details.php
@@ -259,7 +259,7 @@ if (count($tweets) > 0) {
                 $old_tweets = $connection->get('statuses/user_timeline', $params);
                 // Check for substring of the new tweet in old tweets
                 foreach ($old_tweets as $old_tweet) {
-                    preg_match('/.*?\(/', $tweet, $match);
+                    preg_match('/Pipeline release!.*?\(/', $tweet, $match);
                     if (strpos($old_tweet->text, $match[0]) !== false) {
                         $already_tweeted = true;
                         echo 'Already tweeted at https://twitter.com/nf_core/status/' .
@@ -274,7 +274,7 @@ if (count($tweets) > 0) {
                 }
             }
             if (!$already_tweeted) {
-                // $connection->post('statuses/update', ['status' => $tweet]);
+                $connection->post('statuses/update', ['status' => $tweet]);
                 echo "Sent tweet: $tweet\n";
             }
         }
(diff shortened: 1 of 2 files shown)


update_pipeline_details.php
```diff
@@ -242,11 +242,41 @@ function sort_datestamp($a, $b) {
...
+                        break;
+                    }
+                }
+
+                if (count($old_tweets) == 0 || $already_tweeted) {
+                    break;
+                }
+            }
+            if (!$already_tweeted) {
+                // $connection->post('statuses/update', ['status' => $tweet]);
```
ewels, line 277: I think this needs to be uncommented?

update_pipeline_details.php
```diff
@@ -242,11 +242,41 @@ function sort_datestamp($a, $b) {
...
+                $params = [
+                    'count' => 200,
+                    'page' => $page,
+                    'exclude_replies' => true,
+                    'exclude_retweets' => true,
+                ];
+                $old_tweets = $connection->get('statuses/user_timeline', $params);
+                // Check for substring of the new tweet in old tweets
+                foreach ($old_tweets as $old_tweet) {
+                    preg_match('/.*?\(/', $tweet, $match);
```
ewels, line 262: This regex looks odd at a glance. Is it right?

ewels (changes requested):

update_pipeline_details.php
mashehu: oops 🙃 I was very scared to actually send out a tweet while testing 😁

update_pipeline_details.php
mashehu: I made it stricter now (I think/hope).

update_pipeline_details.php
ewels: Better safe than sorry! 😅

update_pipeline_details.php
ewels: Looks good! I hadn't properly understood what it was doing but I get it now. I thought that you were using the regex to check for the name + version directly, had my noticed the second strpos check.

ewels (approved):

ewels:
Now we wait! 🤞🏻

PR #1273, opened 2022-10-03
show most recent event as current

mashehu:
fixes the index header showing not the most recent event. We should think about listing other events there as well somehow.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1273"></a>


ewels (approved):
Thanks!

PR #1277, opened 2022-10-03
Migrate stats from json files to MySQL database

mashehu:
For now, this migrates everything need for the pipelines pages, but not yet stats (that's why `stats.php` still contains the old code).
The entries corresponding to older stats from `nfcore_stats.json` and `nfcore_issue_stats.json` can be manually migrated into the database with `import_stats.php` (the `contributor_stats.json` files can be completely reconstructed from github.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1277"></a>


mashehu:
@nf-core-bot fix linting

PR #1286, opened 2022-10-04
Prettier double quotes

ewels:
At the moment all nf-core repos use double quotes for Prettier linting, except this one.

I think it's because we wanted to keep single quotes in the PHP code, but that doesn't mean that we have to have single quotes in all other files (YAML and Markdown front-matter mostly).

This PR updates the prettier config so that the single quotes setting only applies to `*.php` files (check the first commit).

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1286"></a>


ewels:
Closing due to site rebuild. Tracked in https://github.com/nf-core/nf-co.re/issues/1497

PR #1290, opened 2022-10-05
Fix stats plots

mashehu:
missing data points are breaking the stats plots for now. I fixed it already for the clones plots, this also does it for repo views.


mashehu:
Merging it, because it works with my local clone of the live db and I guess @ewels is quite busy atm 🤠

ewels (approved):
LGTM!

PR #1292, opened 2022-10-05
fix colors in dark mode

mashehu:
there might be something more wrong, but this is a quick fix to limit the damage.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1292"></a>


mashehu:
@nf-core-bot fix linting

ewels (approved):
Yeah definitely a bit suspicious that this is needed! But hopefully Bootstrap 3.6 will ship with native dark mode soon and we can strip most of this out anyway 🤞🏻

PR #1293, opened 2022-10-05
Migrate stats page to db

mashehu:
switched half of the code stats cards and the leader board (which for some reason loaded the `stats.json` increasing the load even further) a second time to be sourced from the database.

Load times are still rough...

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1293"></a>


mashehu:
@nf-core-bot fix linting

maxulysse:

ewels:
Leaving this as for now the new site will still load the old stats page. So this _may_ still be relevant for a bit. Maybe.

mashehu:
done in #2695

PR #1333, opened 2022-10-10
From: https://github.com/nf-core/nf-co.re/issues/234 Expand --json flag explanation 

vbarrera:
From https://github.com/nf-core/nf-co.re/issues/234

Expand --json flag explanation for nf-core list to indicate that the output is not filtered.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1333"></a>


ctuni (approved):
More understandable now!

PR #1366, opened 2022-10-11
Update creating_with_nf_core.md

GeOdette:
Added a schema for the output of nf-core create command

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1366"></a>


maxulysse:
@nf-core-bot fix linting pretty please :pray:

ewels (changes requested):
Apologies, but I don't think that we should include this in the documentation.

The files created by `nf-core create` change on almost every update of tools, so this documentation is already out of date and will be a lot of work to keep in sync.

We could try to write some automation to keep this part of the docs automatically updated, but it'd be quite a lot of work for not much benefit I think.

So apologies, thank you for the work and contribution, but I will close this one 🙏🏻

PR #1375, opened 2022-10-11
Fix syntax for nf-core modules lint example

ameynert:


markdown/developers/tutorials/dsl2_modules_tutorial.md
```diff
@@ -260,7 +260,7 @@ Now that you've completed code development, you are ready to check if your code
 This can also be done easily using _nf-core tools_ just by changing folder into the parent _modules_ directory and typing the command
```
abhi18av, line 261: @ameynert , thanks for the initiative - I thought that if the user moves into the `modules` directory then it might not be correct to say `-d modules`


Have you tried this locally?

ameynert: Yes, I was testing this on cellranger/counts locally, works fine, but I was in the parent directory. Will fix to '.'

abhi18av (approved):

PR #1381, opened 2022-10-11
Subworkflows docs

edmundmiller:


FranBonath:
@ewels draft workflow that doesn't have any functional links currently. Maybe this be closed without merge?

ewels (approved):

ewels:
The TODO comment points to a non-existent tutorial, yes. But the second part of the diff replaces a chunk of duplicated docs with a link to the primary source, which I think is a good change.

The comment doesn't really matter, so happy to merge as-is here. If @mashehu permits it 😬 👀

mashehu:
done

PR #1408, opened 2022-10-26
Fix scrollspy and ToC

mashehu:
following a lead by @sofiahag, scrollspy works again if we target the first `<nav>` inside `.toc`. Don't fully understand, why it stopped working and it works with this 🤷🏻 
Additionally, I changed the order of element styling such that "On this page" and  "back to top" is always visible.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1408"></a>

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -298,8 +298,7 @@ function get_self_url($strip_query = true) {
 }
 
 function generate_toc($html_string) {
-    $toc =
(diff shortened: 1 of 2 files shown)


sofiahag (approved):
Ah, great, you found it! 
I tried finding the exact element, but struggled 👏

mashehu:
@nf-core-bot fix linting

PR #1413, opened 2022-10-30
Add `run with` buttons in pipelines pages

mribeirodantas:
This is a proof of concept for the idea of adding buttons in every
pipeline page with instructions on how to run them with `nf-core launch`,
`nextflow run` and the Tower CLI.

Signed-off-by: Marcel Ribeiro-Dantas <mribeirodantas@seqera.io>

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1413"></a>

Preview below:
Screenshot 2022-10-30 at 12 28 54
Screenshot 2022-10-30 at 12 28 48

### DIFF
--- a/includes/pipeline_page/sidebar.php
+++ b/includes/pipeline_page/sidebar.php
@@ -63,8 +63,10 @@ $embed_video = array_values($embed_video)[0];
     <div class="row border-bottom pb-2">
         <div class="col-12">
             <h6><i class="fas fa-terminal fa-xs"></i> Run with
-            <p class="btn btn-success d-print-none" id="btn_nfcore" onclick="changeText('nf-core launch <?php echo $pipeline->full_name . $release_cmd; ?> -profile test --outdir <OUTDIR>');changeInfo('');">nf-core</p>
-            <p class="btn btn-success d-print-none" id="btn_nxf" onclick="changeText('nextflow run <?php echo $pipeline->full_name . $release_cmd; ?> -profile test --outdir <OUTDIR>');changeInfo('');">Nextflow</p>
+            <p class="btn btn-success d-print-none" id="btn_nfcore" onclick="changeText('nf-core launch <?php echo $pipeline->full_name .
+                $release_cmd; ?> -profile test --outdir <OUTDIR>');changeInfo('');">nf-core</p>
+            <p class="btn btn-success d-print-none" id="btn_nxf" onclick="changeText('nextflow run <?php echo $pipeline->full_name .
+                $release_cmd; ?> -profile test --outdir <OUTDIR>');changeInfo('');">Nextflow</p>
             <p class="btn btn-primary d-print-none" id="btn_tower" onclick="changeText('tw launch https://nf-core/sarek');changeInfo('Read how to configure the Tower CLI <u><a href=\'https://github.com/seqeralabs/tower-cli/#2-configuration\'>here</a></u>.');">Tower</p></h6>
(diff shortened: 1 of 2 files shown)


mribeirodantas:
@nf-core-bot fix linting

mashehu:
Nice idea! How about we model this closer to how github does it and have them basically as tabs, instead of buttons:
Screenshot 2022-10-31 at 10 12 21

mribeirodantas:
That could be an option. I was a bit worried that some people wouldn't realize you can click the other options if it isn't explicitly with buttons (some shape structure around the words), but if we get some feedback that underlined text is fine, sure 😄

mashehu:
This is how it would like as tabs:
dark-mode
light-mode

mribeirodantas:
> I found the UX a bit confusing with the buttons to be honest. I expected that something similar to clicking on the launch button in the header would happen.

Hm.. that's a good point. But at least you knew it was clickable 😬.

I liked the GIFs you posted. Do you mind committing these changes to this PR?

mribeirodantas:
Great, @mashehu ! Just rendered locally your commit and it looks better indeed! Do you think we should do something to the "Run with" to draw more attention? Maybe coloring it green? Bold?

mashehu:
I wouldn't do more color (especially in dark mode there is already quite a bit of coloring). We could give it a bit more vertical distance to the rest, similar to how it is done on npmjs.com:

Screenshot 2022-11-01 at 14 26 45

ewels:
Haven't tested but going from the screenshots I wonder if it needs some spacing between the command input box and the buttons?

mribeirodantas:
On GitHub, there is more space due to the thick line under the options that appear depending on what you clicked. For our case, because we don't have this line, I think it's fine, but I wouldn't really mind adding more space between the input box and the button texts.

includes/pipeline_page/sidebar.php
```diff
@@ -60,13 +60,41 @@
...
+                    <button class="nav-link text-muted active" id="nfcore-tab" data-bs-toggle="tab" data-bs-target="#pipeline-nfcore-run-cmd" type="button" role="tab" aria-controls="nfcore" aria-selected="true">nf-core</button>
+                </li>
+                <li class="nav-item" role="presentation">
+                    <button class="nav-link text-muted" id="nf-tab" data-bs-toggle="tab" data-bs-target="#nf" type="button" role="tab" aria-controls="nf" aria-selected="false">Nextflow</button>
+                </li>
+                <li class="nav-item" role="presentation">
+                    <button class="nav-link text-muted" id="tw-tab" data-bs-toggle="tab" data-bs-target="#tw" type="button" role="tab" aria-controls="tw" aria-selected="false">Tower</button>
+                </li>
+            </ul>
+            <div class="tab-content">
```
ewels, line 77:
```suggestion
            <div class="tab-content mt-2">
```

ewels:
Thanks @mribeirodantas !

mashehu:
Now it looks to  me more disjointed from the tab headers
Screenshot 2022-11-02 at 12 39 15
maybe adding some padding instead?

ewels:
Ah, I assumed it had the top leading border already (and I couldn't see it as it was duplicated). Sorry, tried to test on GitPod but couldn't get pipelines in so did it blind.

I was expecting it to look like this:
image

With the input box below

mashehu:
I removed that border with a `border-bottom-0` to the `nav-tabs`, because it looked weird together with the border of the input field.

PR #1418, opened 2022-10-31
fixing the stats page again 🤞🏻 

mashehu:
Not sure if it is that simple, but live debugging on the server worked after I load `stats.json` only once in `stats.php`

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1418"></a>

### DIFF
--- a/public_html/stats.php
+++ b/public_html/stats.php
@@ -546,8 +546,6 @@ A list of these repositories can be found <a href="#core_repos">below</a>.</p>
 <table class="table table-sm mt-5">
   <tbody>
 <?php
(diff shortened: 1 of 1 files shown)


ewels (approved):
Wait, what? It was loading the multi-hundred-megabyte JSON file _twice_ for every page load?? 🙈

ewels:
Can't really review on my phone properly to check whether these variables are being loaded in other contexts etc or whatever. But feel free to merge if you're confident.

mashehu:
it has the same variable name as the import above, so should all be fine.

PR #1423, opened 2022-11-02
Event page filtering

mashehu:
Too many people asked for it, so I broke and throw in a quick implementation 🙂 

Added bytesize as it's own category.

The two column layout was an idea by @FranBonath 
Screenshot 2022-11-02 at 14 31 07

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1423"></a>

### DIFF
--- a/public_html/assets/js/nf-core.js
+++ b/public_html/assets/js/nf-core.js
@@ -445,6 +445,21 @@ $(function () {
       bootstrap.ScrollSpy.getInstance(dataSpyEl).refresh();
     });
   });
+
+  // Filter events with buttons
+  $('.events-toolbar .events-filters button').click(function () {
+    $(this).blur().toggleClass('active');
+    var showclasses = [];
+    $('.events-toolbar .events-filters button.active').each(function () {
+      showclasses.push($(this).data('bsTarget'));
+    });
+    if (showclasses.length > 0) {
+      $('.event-list .card').filter(showclasses.join(', ')).show();
+      $('.event-list .card').not(showclasses.join(', ')).hide();
+    } else {
+      $('.event-list .card').show();
+    }
+  });
 });
 
 function scroll_to(target_el, offset) {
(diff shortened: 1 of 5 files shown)


ewels:
Love the filters 👍🏻 

- Don't love the side-by-side view sorry. Upcoming is often empty / very small list compared to past. So just shows as a massive half-page of empty space to me, which looks a bit weird?
- Can we have a text filter box like we do for the pipelines / modules?
- What do you think about making the entire event box a link?

mashehu:
Now back to one column layout, but with a bit more distance between them, to make them more distinguishable.

Added the text search (only on card content)

> - What do you think about making the entire event box a link?
👎🏻 I think having only text and button as click-able items is more in line with other websites and reduces instances of being sent to a new page by accident.

ewels:
Ahhh the search box is so good 😍  Going to make it a million times easier to find specific bytesize talks 💯 

Just small stuff now:

- More spacing below buttons, above "Upcoming events" header:
    - image
- Subheading to split events by year?
    - Might be nice to have some visual shift to break up the huge list, though could look crappy when filtering 🤔 

Now that we're back to full width it does strike me that the bars are mostly empty 😆  What about a more tabular view instead? Might make it easier to scan / read and make better use of the screen real estate. As in, have a column for title, column for date, column for speaker etc. Can still be divs, doesn't actually have to be a table (so can be responsive).

Equally, happy to merge after the spacing is added if you don't want to go down this rabbit hole 😉

ewels:
Looks great! 🎉

ewels:
Ok I know it's merged, but now I'm thinking that it's not actually that useful to do an `AND` filtering. Like, never.

What do you think about adding a button that is `All` and then only being able to select one button at a time? Would be easier to flick through the categories then. Personally finding it a bit unintuitive when I click hackathons followed by tutorials and nothing seems to update (as tutorials are all hidden below the hackathons, which remain).

PR #1437, opened 2022-11-07
Add local hackathon for the IGDR in Rennes France

LouisLeNezet:
We would like to participate to the Hackathon and do a local event in Rennes inside our Institute.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1437"></a>


LouisLeNezet:
Would it be better to squashed everything in one commit before merging to the master ?

mashehu:
> Would it be better to squashed everything in one commit before merging to the master ?

nah, no worries, just hit merge when you are happy. 🙂

PR #1449, opened 2022-11-10
Simplify text in buttons on Join page

ewels:
As we get more buttons on the Join page, they're getting a bit wordy.

Now:

image

Before:

image

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1449"></a>

### DIFF
--- a/public_html/join.php
+++ b/public_html/join.php
@@ -24,20 +24,20 @@ include '../includes/header.php';
 <p>We use a few different tools to organise the nf-core community -
     you are welcome to join us at any or all!</p>
 <p class="text-center">
-  <a href="https://nf-co.re/join/slack" class="mb-2 btn btn-lg btn-success">
-    <i class="fab fa-slack"></i> Join nf-core Slack
+  <a href="https://nf-co.re/join/slack" class="mb-2 me-3 btn btn-lg btn-success">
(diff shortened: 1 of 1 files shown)


public_html/join.php
```diff
@@ -24,20 +24,20 @@
 <p>We use a few different tools to organise the nf-core community -
     you are welcome to join us at any or all!</p>
 <p class="text-center">
-  <a href="https://nf-co.re/join/slack" class="mb-2 btn btn-lg btn-success">
-    <i class="fab fa-slack"></i> Join nf-core Slack
+  <a href="https://nf-co.re/join/slack" class="mb-2 me-3 btn btn-lg btn-success">
```
mashehu, line 27: because we use the brand colors for the rest:
```suggestion
  <a href="https://nf-co.re/join/slack" class="mb-2 me-3 btn btn-lg btn-success" style="background-color: #4a154b;">
```

mashehu (approved):

public_html/join.php
ewels: Ok, but has a green border so needs to be btn-dark as well.
```suggestion
  <a href="https://nf-co.re/join/slack" class="mb-2 me-3 btn btn-lg btn-dark" style="background-color: #4a154b;">
```

PR #1461, opened 2022-11-21
Clarify usage of optional outputs in output tuples

pinin4fjords:
From conversation in https://nfcore.slack.com/archives/CE4K7FEHE/p1669023424969749, I'm proposing a clarification to the associated documentation regarding optional outputs in otherwise non-optional tuples. 

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1461"></a>

### DIFF
--- a/markdown/developers/modules.md
+++ b/markdown/developers/modules.md
@@ -510,6 +510,14 @@ process {
 
 3. Optional inputs are not currently supported by Nextflow. However, passing an empty list (`[]`) instead of a file as a module parameter can be used to work around this issue.
 
+4. Optional outputs SHOULD be marked as optional:
+
+     ```nextflow
+     tuple val(meta), path('*.tab')                   , optional:true, emit: tab
+     ```
+
+5. Since it is not currently possible to mark individual elements of a tuple as optional, optional outputs SHOULD NOT be placed in output channels with mandatory outputs. 
+
 ### Module parameters
 
 1. A module file SHOULD only define input and output files as command-line parameters to be executed within the process.
--- a/markdown/developers/modules.md
+++ b/markdown/developers/modules.md
@@ -513,7 +513,7 @@ process {
 4. Optional outputs SHOULD be marked as optional:
 
    ```nextflow
-   tuple val(meta), path('*.tab')                   , optional:true, emit: tab
+   tuple val(meta), path('*.tab'), emit: tab,  optional: true
    ```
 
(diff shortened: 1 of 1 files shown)


pinin4fjords:
@nf-core-bot fix linting

markdown/developers/modules.md
```diff
@@ -510,6 +510,14 @@ process {
 
 3. Optional inputs are not currently supported by Nextflow. However, passing an empty list (`[]`) instead of a file as a module parameter can be used to work around this issue.
 
+4. Optional outputs SHOULD be marked as optional:
+
+   ```nextflow
+   tuple val(meta), path('*.tab')                   , optional:true, emit: tab
+   ```
+
+5. Since it is not currently possible to mark individual elements of a tuple as optional, optional outputs SHOULD NOT be placed in output channels with mandatory outputs.
```
mahesh-panchal, line 519:
```suggestion
5. Each output file SHOULD be emitted in it's own channel, along with the `meta` map if provided ( the exception is the versions.yml ).
```

mahesh-panchal: @drpatelh @jfy133 This is usually enforced right? Each file should be in it's own channel?

pinin4fjords: Ahh, that's new on me, I have a couple of modules with multiple-file output tuples

jfy133: yes, we enforce that - as we don't want to make assumptions what people want to use each output file from (what you may consider should 'travel together' may not be what others do...). So yes, lots of re-joining tuples...

markdown/developers/modules.md
```diff
@@ -510,6 +510,14 @@ process {
 
 3. Optional inputs are not currently supported by Nextflow. However, passing an empty list (`[]`) instead of a file as a module parameter can be used to work around this issue.
 
+4. Optional outputs SHOULD be marked as optional:
+
+   ```nextflow
+   tuple val(meta), path('*.tab')                   , optional:true, emit: tab
```
mahesh-panchal, line 516:
```suggestion
   tuple val(meta), path('*.tab'), emit: tab,  optional: true
```
This is just a style suggestion. I prefer the emits to line up and spaces

mahesh-panchal (approved):

markdown/developers/modules.md
pinin4fjords: Sure, I just copy/pasted directly from a module, but happy to adjust.

PR #1465, opened 2022-11-23
transcript bytesize nf-core/nascent

FranBonath:


markdown/events/2022/bytesize-nascent.md
```diff
@@ -23,3 +23,107 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[12:29](https://www.youtube.com/watch?v=chayGGPTnfM&t=749)
+Let's talk a little bit about the transcript identification because that's the most interesting part of the pipeline and what makes it unique. There's a couple of different options, as I said. First, if you're doing GRO-seq, I have some great support for that. If anyone would like to support other assays or would like to see it supported, please open an issue. So first is GroHMM, and this is what kind of sparked us getting into the nextflow and going into bigger pipelines. It was difficulty reproducing this and running this on big enough machines to actually use it because it's an R package. It was released in 2015 by Minho Chae, Charles Danko, and Lee Kraus, actually just down the street at UT Southwestern.
```
christopher-hakkaart, line 82:
```suggestion
Let's talk a little bit about the transcript identification because that's the most interesting part of the pipeline and what makes it unique. There's a couple of different options, as I said. First, if you're doing GRO-seq, I have some great support for that. If anyone would like to support other assays or would like to see it supported, please open an issue. So first is GroHMM, and this is what kind of sparked us getting into the nextflow and going into bigger pipelines. It was difficult reproducing this and running this on big enough machines to actually use it because it's an R package. It was released in 2015 by Minho Chae, Charles Danko, and Lee Kraus, actually just down the street at UT Southwestern.
```

markdown/events/2022/bytesize-nascent.md
```diff
@@ -23,3 +23,107 @@ It is our hope that these talks / videos will build an archive of training mater
...
... new assay that just came out in 2022, and it's very exciting. I just left this in up at the top in Figure A. This is just also illustrating the difficulty in reproducing these, and this is on the exact same data sets. You can see the difference in the HOMER and the GroHMM results and just how much they vary and just a slight tweaking a tool. You'd expect maybe better performance, but you wouldn't expect completely different results based on what tool you're using. Down here at the bottom, this is the PINTS identification method. Just in a rudimentary way, it works very similarly to MACS2. This is a potential true peak based on the density of this, and it's very easy to pick out. It does some algorithms, picks up the local background noise from these, and these are the light blue. You can see in the purple from those, that's then a potential peak that it needs to test and see, is that actually a peak or is it just more noise from the assay itself? What PINTS is doing is really picking up these TSS start sites. As you can see from the read pile up here, it's just picking up the TSS site rather than the entirety of the transcript, which might actually lead out all the way along here.
```
christopher-hakkaart, line 91:
```suggestion
Let's now jump into PINTS identification. This is a new assay that just came out in 2022, and it's very exciting. I just left this in up at the top in Figure A. This is just also illustrating the difficulty in reproducing these, and this is on the exact same data sets. You can see the difference in the HOMER and the GroHMM results and just how much they vary by just a slight tweaking of a tool. You'd expect maybe better performance, but you wouldn't expect completely different results based on what tool you're using. Down here at the bottom, this is the PINTS identification method. Just in a rudimentary way, it works very similarly to MACS2. This is a potential true peak based on the density of this, and it's very easy to pick out. It does some algorithms, picks up the local background noise from these, and these are the light blue. You can see in the purple from those, that's then a potential peak that it needs to test and see, is that actually a peak or is it just more noise from the assay itself? What PINTS is doing is really picking up these TSS start sites. As you can see from the read pile up here, it's just picking up the TSS site rather than the entirety of the transcript, which might actually lead out all the way along here.
```

christopher-hakkaart (changes requested):
Some very minor things but otherwise excellent

FranBonath:
@nf-core-bot fix linting

christopher-hakkaart (approved):
Looks great - thanks for making the changes!

PR #1467, opened 2022-11-25
Tweak ungrouped parameter tooltip 

jfy133:
As the selection only happens 'after' pressing this button, this confused me (I tried to find a way of 'selecting' the ungrouped parameters myself with non-existint checkboxes...)

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1467"></a>

### DIFF
--- a/public_html/assets/js/nf-core-schema-builder.js
+++ b/public_html/assets/js/nf-core-schema-builder.js
@@ -1515,7 +1515,7 @@ function generate_group_row(id, param, child_params) {
                 <div class="col-auto align-self-center schema_row_config border-start" title="Open settings" data-bs-toggle="tooltip">
                     <i class="fas fa-cog"></i>
                 </div>
-                <div class="col-auto align-self-center schema_group_move_params" title="Select parameter(s) to be moved into this group" data-bs-toggle="tooltip">
+                <div class="col-auto align-self-center schema_group_move_params" title="Moved ungrouped parameter(s) to this group" data-bs-toggle="tooltip">
                     <i class="fas fa-folder-download"></i>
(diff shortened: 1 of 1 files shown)


mashehu:
Hmm, not sure about this. Your wording makes it less clear to me how I can choose which parameters to move. How about `Select and move`?

jfy133:
Ok I had a small typo, but still as the button is in the group, just `select and move` alone would imply to me to select the group not the ungrouped parmaeters.

What about (if not too long): `Select and move ungrouped parameters to here`?

mashehu:
Sorry, should have written my complete proposal: I meant `Select and move parameter(s) into this group`

jfy133:
I would still mention `ungrouped` but otherwise I think that would be OK.

I still worry about `select` though... maybe just `Move ungrouped parameters into this group` is sufficient? They will do it the first time and then realise they can select?

mashehu:
> They will do it the first time and then realise they can select?

With that argumentation we can leave it at the current state 😉  Again, `move` only would let me wonder how to select them in the first place. 

The `select` here has the same function as the `add` in  `Add help text`, that it is a clear action which I will be able to perform when clicking on the button. Not that the button does that for me (which wouldn't be the case with`move` either)

jfy133:
Ok good point... OK I'll think about it a bit further. I sort of feel like just `Move` is more intuitive, _selecting_ the parameters is a 'bonus' is more what I was implying.

jfy133:
I looked at it  with fresh eyes and actually it makes sense to me, I think I was having a bad day..!

PR #1476, opened 2022-12-02
add explicit time zone to event start times, update calender-links version

mashehu:
Our ics files from events did not work with the macOS calendar app, as reported by @matthdsm. This is because the timezone is not  reported in the correct format, as brought up on the [library repo](https://github.com/spatie/calendar-links/issues/129).

Explicitly adding the time zone seems to fix it.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1476"></a>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "description": "PHP dependencies for the main nf-core website.",
     "require": {
         "abraham/twitteroauth": "^2.0",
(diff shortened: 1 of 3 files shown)


mashehu:
@nf-core-bot fix linting

matthdsm (approved):

PR #1486, opened 2022-12-08
Attempt at writing guidance of which inputs types to use when in modules

jfy133:
I definitely made this overly complicated, sorry 🤦‍♀️

### DIFF
--- a/markdown/developers/modules.md
+++ b/markdown/developers/modules.md
@@ -762,6 +762,104 @@ To see some more advanced examples of these keys in use see:
 - [Set ext.prefix based on task inputs](https://github.com/nf-core/rnaseq/blob/e049f51f0214b2aef7624b9dd496a404a7c34d14/conf/modules.config#L297)
 - [Set ext.args based on both parameters and task inputs](https://github.com/nf-core/rnaseq/blob/e049f51f0214b2aef7624b9dd496a404a7c34d14/conf/modules.config#L377-L381)
 
+## Guidance on module inputs
+
+There are various ways information that a given software needs can be entered into a module to make the module work.
+We aim to make nf-core modules as flexible as possible, to ensure pipelines are not constrained by 'particularities' of a given module. In otherways, we want to make modules work for pipelines, not pipelines to work for modules.
+Here we will provide some general guidance on how and when to use the various ways of passing information into a module.
+
+The five ways of inserting information into a module are:
+
+- `input:`: a distinct input channel defined by in the module itself
+- `val(meta)`: a common entry in nf-core module tuple-based input channels that is a list of information associated with a given file
+- `ext.args`: defined in `modules.conf` configuration files that is used to insert (normally non-mandatory) custom flags/options to the tools used in the module
+- `ext.prefix`: defined in `modules.conf` configuration files that is used for defining file names (typically based on a `meta.id` element)
+- `ext.<custom>`: defined in `modules.conf, custom 'variables' that can be passed into the module, defined by the writer of the module
+
+In general the priority order is:
+
+- All files should be inserted into a module via `input:` channels
+- All non-file inputs: `args` > `meta` > `ext.custom`
+  - Where custom pipeline-specific `meta` keys can be inserted into an `args`, rather than module-hardcoded `meta` keys
+
+## ext.args
+
+This is the main way to pass non-file inputs (flags, options, strings, etc.) to a tool within a module. Every nf-core module must include an `$args` variable within the command of the tool itself.
+
+Pipeline developers can insert any information stored in their `meta` maps to this variable via a `modules.config` (see below). Users can also add their own additional personal parameters or information to this via a custom `modules.conf` file (at their own risk).
+
+This can be done via closures, and be dynamically constructued using conditions within lists.
+
+```nextflow
+ext.args = {
+  [
+    "-a" ,
+    "-b ${meta.id}",
+    run_mode ? "-c" : ""
+    ].join(' ') }
+```
+
+It is important to note that it is problematic to have something in ` config` that keeps a _module_ working, as configs can be overwritten. Therefore if the _module_ (not the tool!) requires a particular parameter to execute such information can be passed via an `input:` channel (see below). However, a pipeline can make sure that the _tool_ keeps working (i.e., non-file arguments) via their own `modules.conf`.
+
+### input chanels
+
+This is the way for passing _files_ into a module. This is to ensure Nextflow correctly stages them in the working directory of a process.
+
+Input channels MAY, in rare cases, be used for other _mandatory_ information (e.g. passing in strings or booleans) in cases where they cannot reasonably passed to the module via `ext.args`.
+
+Non-`val()` input channels are mainly distinguished from `ext.args` because `val(inputs)` are only customisable by the pipeline (not the user).
+
+Information stored in a `meta` map that needs to be passed to a `val()` input channel can be done so via the [`.multiMap{}`](https://www.nextflow.io/docs/latest/operator.html#multimap) Nextflow operator.
+
+## meta
+
+`meta` lists can be used to pass pipeline-mandatory parameters to a tool, giving most flexibility to a pipeline developer.
+
+Pipeline developers can use whatever key name they want to record a given bit of information, and pass this into the module either via `ext.args` (or in some rare cases, a dedicated input channel, see above).
+
+Module developers SHOULD NOT assume 'hardcoded' `meta` fields and thus not use these within a module, but rather leave this to the pipeline developer to construct the correct parameters via `ext.args` in a config.
+
+For example, a module developer SHOULD NOT hardcode the use of a field such as `${meta.reference}`
+
+```bash
+tool \\
+  run \\
+  $args \\
+  -c ${meta.reference}
+```
+
+This then requires a pipeline developer to confirm to a particular naming scheme, that may not be compatible in the way other modules store the same information in their `meta`. Therefore we
+
+Rather, the pipeline developer can insert this into the module via `args`, where the command is:
+
+```bash
+tool \\
+  run \\
+  $args
+```
+
+and the modules config
+
+```nextflow
+ext.args = { "-c ${meta.reference}" }
+```
+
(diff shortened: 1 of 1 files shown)


markdown/developers/modules.md
```diff
@@ -762,6 +762,104 @@ To see some more advanced examples of these keys in use see:
...
+- `val(meta)`: a common entry in nf-core module tuple-based input channels that is a list of information associated with a given file
+- `ext.args`: defined in `modules.conf` configuration files that is used to insert (normally non-mandatory) custom flags/options to the tools used in the module
+- `ext.prefix`: defined in `modules.conf` configuration files that is used for defining file names (typically based on a `meta.id` element)
+- `ext.<custom>`: defined in `modules.conf, custom 'variables' that can be passed into the module, defined by the writer of the module
+
+In general the priority order is:
+
+- All files should be inserted into a module via `input:` channels
+- All non-file inputs: `args` > `meta` > `ext.custom`
+  - Where custom pipeline-specific `meta` keys can be inserted into an `args`, rather than module-hardcoded `meta` keys
```
mahesh-panchal, line 783:
```suggestion
In general:

- All files, mandatory, or non-mandatory, must have a corresponding `input:` channel ( type `path` ).
- Command line options essential to the functioning of the tool, must also be `input:` channels ( type `val` ), or encoded as a sensible default that can be overridden by checking `ext.args`. 
- Filenames should be defined by `ext.prefix` and a file extension. 
- Sample specific command-line options should be defined based on fields of the `meta` map. If the command-line option is essential to the function of the tool, then the channel operator `multiMap` should be used to construct the string and pass it to an `input:` channel. If the command-line option is not essential to the operation of the tool or certain options should be defined by what is in the `meta` map, then the command-line options should be defined using `ext.args`.  
```

markdown/developers/modules.md
```diff
@@ -762,6 +762,104 @@ To see some more advanced examples of these keys in use see:
...
+
+This can be done via closures, and be dynamically constructued using conditions within lists.
+
+```nextflow
+ext.args = {
+  [
+    "-a" ,
+    "-b ${meta.id}",
+    run_mode ? "-c" : ""
+    ].join(' ') }
```
mahesh-panchal, line 799:
```suggestion
ext.args = {
  [
    "-a" ,
    "-b ${meta.id}",       // Set's an option value to be the meta.id
    run_mode ? "-c" : ""   // run_mode is a variable defined in the main.nf of the module
    ].join(' ') }
```

markdown/developers/modules.md
```diff
@@ -762,6 +762,104 @@ To see some more advanced examples of these keys in use see:
...
+```
+
+It is important to note that it is problematic to have something in ` config` that keeps a _module_ working, as configs can be overwritten. Therefore if the _module_ (not the tool!) requires a particular parameter to execute such information can be passed via an `input:` channel (see below). However, a pipeline can make sure that the _tool_ keeps working (i.e., non-file arguments) via their own `modules.conf`.
```
mahesh-panchal, line 802: > However, a pipeline can make sure that the _tool_ keeps working (i.e., non-file arguments) via their own `modules.conf`. 

This is not true. A custom config completely overrides `ext.args` and if an option is supplied, the onus ends up on the user to make sure the tool keeps functioning.

mahesh-panchal (commented):
I'll provide some more comments in a bit.

markdown/developers/modules.md
jfy133: Ah true!

jfy133:
@mahesh-panchal please feel free to just directly make chages on my fork, no need for suggestions I definitely trust you over myself here ;)

markdown/developers/modules.md
mahesh-panchal, line 802:
```suggestion
It is important to note that it is problematic to have something in a `config` that keeps a _module_ working, as configs can be overwritten. Therefore if the _module_ (not the tool!) requires a particular parameter to execute such information must be passed via an `input:` channel (see below).
```

mahesh-panchal:
OK. What do you think?

markdown/developers/modules.md
```diff
@@ -762,6 +762,222 @@ To see some more advanced examples of these keys in use see:
...
+## Guidance on module inputs
+
+There are various ways information that a given software needs can be entered into a module to make the module work.
+nf-core modules are intended to be flexible as possible, and ensure pipelines are not constrained by 'particularities' of a given module.
+In other words, we want to make modules work for pipelines, not pipelines to work for modules.
+Here we provide some general guidance on how and when to use the various ways of passing information into a module.
+
+The four ways of passing information into a module are:
+
+- `input:`: a distinct input channel defined by in the module itself
```
jfy133, line 774:
```suggestion
- `input:`: a distinct input channel defined in the module itself
```

jfy133:
Looks good, even if maybe it is a bit too wordy - however the examples help a lot.

I'm not sure if @FriederikeHanssen will agree entirely about the non-file input: channels, but lets see ;).

jfy133:
Yup, that's also what I originally felt as well, but was repeatedly argued against (because of 'maintainability' reasons), and that the pipeline developer also can keep control over this stuff via `ext.args` etc.

Of course there is also the risk that a user could overwrite this with a custom modules file, but in my specific cases I'm not expecting users to be advanced or comfortable enough to fiddle with those so I've been safe so far with that...

but anyway. Like I said, lets see what others say :+1:

mahesh-panchal (approved):

markdown/developers/modules.md
```diff
@@ -762,6 +764,222 @@ To see some more advanced examples of these keys in use see:
...
+- `val(meta)`: a common entry in nf-core module tuple-based input channels that is a list of information associated with a given sample/entity or associated files ( e.g. `meta.single_end` describing the endedness of the sequence files associated with the sample).
+- `ext.args`: defined in `modules.conf` configuration files that is used to insert non-mandatory custom flags/options to the tools used in the module.
+- `ext.prefix`: defined in `modules.conf` configuration files that is used for defining file names (typically based on a `meta.id` element).
+
+Note: Custom ext fields `ext.<custom>` are discouraged as most use-cases can be accomodated using the above techniques.
+
+In general:
+
+- All files, mandatory, or non-mandatory, MUST have a corresponding `input:` channel ( type `path` ).
+- Command-line options essential to the functioning of the tool MUST also be `input:` channels ( type `val` ), or encoded as a sensible default that can be overridden by checking `args` (e.g., seeding options for reproducibilty).
```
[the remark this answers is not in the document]
mahesh-panchal: True.
However, now that I've started actually using this, I feel that this way goes towards better maintainability. For those that are asked to add new modules to workflows, it signals quicker what needs to be added, as there's no record of what's a mandatory non-file options otherwise. And as I say above, it makes it easier to change workflow options for users.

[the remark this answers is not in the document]
FriederikeHanssen: It could be a bit frustrating if we change the rules every couple of months tbh :D  we had a poll about it and all.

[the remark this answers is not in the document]
mahesh-panchal: I'm hoping this won't need to be changed again. Do we have good reasons for maintaining mandatory non-file options via `ext.args`?

pinin4fjords (approved):
Realise there's controversy, but I really like it, thanks for the work.

markdown/developers/modules.md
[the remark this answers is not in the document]
FriederikeHanssen: I really don't see an advantage of passing the mandatory params all through the pipeline and subworkflows. Imo it will add a ton of boilerplate code and make it much harder to see what a module is actually run with. How would you handle arguments that are mandatory but sample specific, i.e. rely on meta map info, that is not standardised across pipelines? The whole extra args/overwritting args thing has been fairly common for us. So the ease of just providing an updated config to users (and having this even more simplified with what @maxulysse is working on here https://github.com/nf-core/sarek/pull/891). For the rare modules that need some weird exception like controlfreec, ascat,gsea we can always have custom solutions of course.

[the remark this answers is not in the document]
mahesh-panchal: > I really don't see an advantage of passing the mandatory params all through the pipeline and subworkflows. 

- Pipeline stability, when a user changes `ext.args`. A user can only change what is meant to be configured by the user. The rest is kept in line with the workflow code.
- A module user can see explicitly which parameters are needed ( and clarified by the meta.yml ). 
  - Pipeline authors need not be as experienced with the tool to get it to work. 
- Single place in code where mandatory arguments can be checked, compared to both workflow and `modules.config`.

> Imo it will add a ton of boilerplate code and make it much harder to see what a module is actually run with. 

These channels are mandatory input. It should be less difficult, rather than thinking all the input is provided, and then it breaks because something is missing from `ext.args` (I've done this many times). Why would not being explicit about something make it harder to run? Also having all the mandatory input come through a single point of entry (i.e. workflow only) makes it easier to see what the module is taking as input. 

> How would you handle arguments that are mandatory but sample specific, i.e. rely on meta map info, that is not standardised across pipelines? 

This is explicit in the text. What can be done to make it clearer?

> The whole extra args/overwritting args thing has been fairly common for us. So the ease of just providing an updated config to users (and having this even more simplified with what @maxulysse is working on here [nf-core/sarek#891](https://github.com/nf-core/sarek/pull/891)). For the rare modules that need some weird exception like controlfreec, ascat,gsea we can always have custom solutions of course.

I think this proposed solution is not the right way to go about it. However, I'm not familiar enough with those modules to really understand what your needs are. The solution proposed keeps mandatory options explicit in separate places, and adds another option on top of the current way of providing input. While it may make the code more lean ( which I'm generally in favour of ), I think this does not really help readability, or understanding. I'm wondering if this is going in the direction of mandatory options for the pipeline, but not the module. In which case, I think that the module should have a patch file in the pipeline where the module code maintains stability for the extra parameters needed for the pipeline to remain functional in the face of changes to `ext.args`.

In my mind, `ext.args` are meant to be User configurable options, that don't break the workflow if changed. The onus should be on the pipeline developer to keep it stable, and that may need to be done using a patch.

PR #1492, opened 2022-12-13
Homepage: Use markdown component for code example

ewels:
This is pretty cool 😉

### DIFF
--- a/src/components/HomepageCodeExample.md
+++ b/src/components/HomepageCodeExample.md
@@ -1,0 +1,18 @@
+```bash
+# Install nextflow
+curl -s https://get.nextflow.io | bash
+mv nextflow ~/bin/
+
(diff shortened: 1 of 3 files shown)


mashehu (approved):

mashehu:
You could even leave the markdown inline and put it inside a `<Markdown>` component from `components/Markdown.astro`

ewels:
I tried that first, but the `<Markdown>` component has been depreciated and removed. I briefly tried to do the same thing with the recommended replacement but I couldn't get it to work. Would be cool if you can figure it out! Feel like it must be possible still if it was before, just need the syntax and naming.

ewels:
Can come back to this to try to do a component another time maybe. For now, prettier-astro plugin is destroying my markup and I can't be bothered to waste any more time trying to fix it.

PR #1498, opened 2022-12-15
bytesize transcript custom scripts

FranBonath:


markdown/events/2022/bytesize-custom_scripts.md
```diff
@@ -22,3 +22,86 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[2:20](https://www.youtube.com/watch?v=3aA5-s8PAF0&t=140)
+This is my first pipeline, which is a Nextflow pipeline that I've written. As you can see it contains a single process called myScript, which is going to take string values as an input and give standard output. In the script block there you'll see that all it's doing is taking the strings defined with the name "str" from my input, and it's going to turn all the lowercase letters into uppercase letters, so nothing too complicated. This is a really simple, in this case, single line of code. In reality, your script could be much, much larger and also written in a different language. In the workflow block down the bottom here, we have "this", "that", and "other" being the three string inputs that we're taking from the channel. I'm piping that into my process and then just using the view operator to show that my output in my screen terminal window.
```
christopher-hakkaart, line 42:
```suggestion
This is my first pipeline, which is a Nextflow pipeline that I've written. As you can see it contains a single process called myScript, which is going to take string values as an input and give standard output. In the script block there you'll see that all it's doing is taking the strings defined with the name "str" from my input, and it's going to turn all the lowercase letters into uppercase letters, so nothing too complicated. This is a really simple, in this case, single line of code. In reality, your script could be much, much larger and also written in a different language. In the workflow block down the bottom here, we have "this", "that", and "other" being the three string inputs that we're taking from the channel. I'm pipeing that into my process and then just using the view operator to show that my output in my screen terminal window.
```

markdown/events/2022/bytesize-custom_scripts.md
```diff
@@ -22,3 +22,86 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[11:55](https://www.youtube.com/watch?v=3aA5-s8PAF0&t=715)
+Just to summarize, what I've covered today is that NextFlow can use custom scripts written from many different languages. Scripts can be stored in both the bin or the templates directory. And both of these will be available to the Nextflow script. Meaning that you don't need to specify an absolute or a relative path as you're executing a script. It's really fantastic to do this because it makes your scripts much more portable and usable by others. Dependencies can be managed using conda and containers. In the examples I've shown, you can see that it can be quite simple or much more complex with the use of mold containers to help you store all those together with singularity and conda images. And with that, I will finish. I think we're probably about where I thought we'd be for time. And I'll finish on this. And if there are any questions, I'm happy to do my best to answer them. Thanks very much.
```
christopher-hakkaart, line 78:
```suggestion
Just to summarize, what I've covered today is that NextFlow can use custom scripts written from many different languages. Scripts can be stored in both the bin or the templates directory. And both of these will be available to the Nextflow script. Meaning that you don't need to specify an absolute or a relative path as you're executing a script. It's really fantastic to do this because it makes your scripts much more portable and usable by others. Dependencies can be managed using conda and containers. In the examples I've shown, you can see that it can be quite simple or much more complex with the use of mulled containers to help you store all those together with singularity and conda images. And with that, I will finish. I think we're probably about where I thought we'd be for time. And I'll finish on this. And if there are any questions, I'm happy to do my best to answer them. Thanks very much.
```

christopher-hakkaart (approved):

markdown/events/2022/bytesize-custom_scripts.md
FranBonath: mulled containers were used a lot in the talk, maybe we should have another swipe through, to change them all. /FBo

christopher-hakkaart: Good call! I can't believe I missed so many! Too worried about the audio matching I guess. I've taken another pass at it

PR #1512, opened 2023-01-02
Bytesize nftest transcript

FranBonath:


markdown/events/2022/bytesize_nftest.md
```diff
@@ -22,3 +22,81 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[1:14](https://www.youtube.com/watch?v=K9B7JRkMpQ4&t=74)
+So what that looks like. If you haven't memorized the Hello Nextflow script, here's a quick reminder. We start saying all of these different greetings in different languages, and we pipe them through, and then we view it lastly. And so what that looks like on the nf-test side is then we have a name of the test, this spoiler plate up here, and then we have a script, and we're actually calling, I believe, the remote script here, and then a name of the test on this line, and then we have what we expect, such as the success of it, how many tasks we're going to have, which I think is also important as you're changing things, and then we have several certs of what's happening in the standard out of these greetings, for example. And that's the power of what you can do with nf-test here, and I'll let your imagination run wild with that of all the things you can do with it from there.
```
Emiller88, line 36:
```suggestion
So what that looks like. If you haven't memorized the Hello Nextflow script, here's a quick reminder. We start saying all of these different greetings in different languages, and we pipe them through, and then we view it lastly. And so what that looks like on the nf-test side is then we have a name of the test, this boiler plate up here, and then we have a script, and we're actually calling, I believe, the remote script here, and then a name of the test on this line, and then we have what we expect, such as the success of it, how many tasks we're going to have, which I think is also important as you're changing things, and then we have several asserts of what's happening in the standard out of these greetings, for example. And that's the power of what you can do with nf-test here, and I'll let your imagination run wild with that of all the things you can do with it from there.
```

markdown/events/2022/bytesize_nftest.md
```diff
@@ -22,3 +22,81 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[10:40](https://www.youtube.com/watch?v=K9B7JRkMpQ4&t=640)
+What this does is it just creates a quick little snapshot with the md5sums of all those files that you specified. And you can do standard out, standard error, et cetera as well on those based on the workflow. But we found that to be a little hairy for right now. As you can see, this is automated. And whenever you rerun your test, it checks for these md5sums and will tell you. The other thing is it's really easy when you know you changed some stuff or you know these are going to change. You just run it with the `-update_snapshots` flag and it updates all the snapshots for you automatically. Then you just commit this file and it will check again against them.
```
Emiller88, line 69:
```suggestion
What this does is it just creates a quick little snapshot with the md5sums of all those files that you specified. And you can do standard out, standard error, et cetera as well on those based on the workflow. But we found that to be a little hairy for right now. As you can see, this is automated. And whenever you rerun your test, it checks for these md5sums and will tell you. The other thing is it's really easy when you know you changed some stuff or you know these are going to change. You just run it with the `--update-snapshot` flag and it updates all the snapshots for you automatically. Then you just commit this file and it will check again against them.
```

markdown/events/2022/bytesize_nftest.md
```diff
@@ -22,3 +22,81 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[12:57](https://www.youtube.com/watch?v=K9B7JRkMpQ4&t=777)
+Those are all great and lots of great things that you should go read about. But I think as a community, we have more important things to focus on. What are some realistic testing priorities? We should probably first convert our CI tests to pipeline tests. I think those are great to reproduce locally and quickly. Next, we can then add more pipeline tests for various params and pathways that you might want to make sure don't break, I think most people are just testing that all their liners work and then hope that everything downstream works, whereas you might want to check all of your skips, for example, and make sure that you get the right amount of processes and that certain things aren't getting run.
```
Emiller88, line 78:
```suggestion
Those are all great and lots of great things that you should go read about. But I think as a community, we have more important things to focus on. What are some realistic testing priorities? We should probably first convert our CI tests to pipeline tests. I think those are great to reproduce locally and quickly. Next, we can then add more pipeline tests for various params and pathways that you might want to make sure don't break, I think most people are just testing that all their aligners work and then hope that everything downstream works, whereas you might want to check all of your skips, for example, and make sure that you get the right amount of processes and that certain things aren't getting run.
```

markdown/events/2022/bytesize_nftest.md
```diff
@@ -22,3 +22,81 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[14:22](https://www.youtube.com/watch?v=K9B7JRkMpQ4&t=862)
+A couple of us have been toying around a lot how a roll out plan could look like. This is just a quick update on this and is no way indicative of exactly what's going to happen. First we want to start with pipelines as a proving ground because I think that's where we had the most ground to gain compared to being stuck with just GitHub actions. So it's already in methyl-seq and demultiplexed. If you're looking for some quick examples and you want to implement this today and switch those over, just that step one, we're going to need to update the template. I've got a PR started for that, but we need to settle on some best practices or how we want to do things so that as you switch between nf-core pipelines and contributing, things are familiar.
```
Emiller88, line 84:
```suggestion
A couple of us have been toying around a lot how a roll out plan could look like. This is just a quick update on this and is no way indicative of exactly what's going to happen. First, we want to start with pipelines as a proving ground because I think that's where we had the most ground to gain compared to being stuck with just GitHub actions. So it's already in methyl-seq and demultiplex. If you're looking for some quick examples and you want to implement this today and switch those over, just that step one, we're going to need to update the template. I've got a PR started for that, but we need to settle on some best practices or how we want to do things so that as you switch between nf-core pipelines and contributing, things are familiar.
```

Emiller88 (changes requested):
Just a few things I caught!

FranBonath:
Thanks @Emiller88!

Emiller88 (approved):
Looks good to me!

PR #1559, opened 2023-01-25
Clarify final release stages, DOI bits

pinin4fjords:
I've tweaked the docs here in two ways:

 - I found the instruction to update the version on the fork's `dev` a little odd. I never have my fork's dev ahead of the upstream, I just synch with the upstream, make branches and PR to upstream dev. So I've tweaked accordingly, but happy to undo that if I'm the odd one there.
 - Less controversially, I've re-jigged the final parts about DOI updates, which no longer work. We don't have direct master write, so I'm instructing the user to include the DOI updates with the other post-release tasks, and then ask a core team member to copy the DOI bit to master (we all agree that's not ideal but the best we can do for now).


maxulysse (approved):

pinin4fjords:
Thanks @maxulysse

PR #1603, opened 2023-03-03
transcript bytesize nf-core/mag

FranBonath:
contains apart from the transcript also the youtube and figshare links.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1603"></a>


markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[9:14](https://www.youtube.com/watch?v=IiorfDHeoLo&t=554)
+Here you can see for example the total length of the resulting assemblies and then compared for different pipeline ones were different assembly settings were used. The lower two pipeline settings correspond to a sample-wise assembly and using either only short or short and long read, so hybrid data, and the upper two settings correspond to a core assembly. Again with short or short and long reads, and what we can see is that the total length of the resulting assemblies significantly increased both by using the hybrid setting, and by applying the core assembly setting. Similar results we also see when looking at the number of MAGs, so the number of genomes that could be retrieved from this data, and also when looking at the L50 values. This indicates that the actual setting that is used for the assembly within this pipeline can have a relatively huge impact on the results. It's definitely good that the pipeline provides different settings, so that you can really choose the correct setting for input data, and it might also be worth to compare different settings.
```
skrakau, line 62:
```suggestion
Here you can see, for example, the total length of the resulting assemblies compared for different pipeline runs for which different assembly settings were used. The lower two pipeline settings correspond to a sample-wise assembly and using either only short or short and long reads, so hybrid data, and the upper two settings correspond to a co-assembly, again with short or short and long reads. And what we can see is that the total length of the resulting assemblies significantly increased both by using the hybrid setting, and by applying the co-assembly setting. Similar results we also see when looking at the number of MAGs, so the number of genomes that could be retrieved from this data, and also when looking at the N50 values. This indicates that the actual setting that is used for the assembly within this pipeline can have a relatively huge impact on the results. It's definitely good that the pipeline provides different settings, so that you can really choose the correct setting for your input data, and it might also be worth to compare different settings.
```

markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
+Another topic I would like to shortly mention is the resource requirements, because this came up quite often in the Slack channel, and it's also somehow difficult to estimate in advance, because it really differs depending on the input data. The main requirements are both for memory and time, coming from the assembly step. As I mentioned already it really differs for different input data sets and I collected some numbers just to give you a rough idea for different pipeline ones that were run by Daniel Straub on our compute cluster. For one rather small sample, which was a cultural sample, both MegaHIT and SPAdes required less than 25 gigabytes and were finished in a couple of hours. However, for a larger river sample data set, MegaHIT took already more than 100 gigabytes of RAM, and it took more than one day to finish, and SPAdes even took more than 900 gigabytes of memory, and it required more than nine days. There was another very large data set containing 15 soil samples for which also a core assembly was performed and for this MegaHIT required one terabyte and more than 17 days, and SPAdes could not even be run because it would have required more than two terabytes of memory.
```
skrakau, line 65:
```suggestion
Another topic I would like to shortly mention is the resource requirements, because this came up quite often in the Slack channel, and it's also somehow difficult to estimate in advance, because it really differs depending on the input data. The main requirements are both for memory and time, coming from the assembly step. As I mentioned already it really differs for different input data sets and I collected some numbers just to give you a rough idea for different pipeline runs that were run by Daniel Straub on our compute cluster. For one rather small sample, which was a culture sample, both MegaHIT and SPAdes required less than 25 GB and were finished in a couple of hours. However, for a larger river sample data set, MegaHIT took already more than 100 GB of RAM, and it took more than one day to finish, and SPAdes even took more than 900 GB of memory, and it required more than nine days. There was another very large data set containing 15 soil samples for which also a co-assembly was performed and for this MegaHIT required one TB and more than 17 days, and SPAdes could not even be run because it would have required more than two TB of memory.
```

markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[11:55](https://www.youtube.com/watch?v=IiorfDHeoLo&t=715)
+This just shows that even for smaller data sets, you cannot run this on your laptop. In general, one can say that it depends on the sequencing depths, the number of samples, the complexity of the analyte metagenome, and also on the applied tool and setting. For this it might be worth noting that, that both assembly tools are run by default but MegaHIT requires much fewer resources than SPAdes, and if you do not want to compute a hybrid assembly it might make sense to consider the `--skip_spades` parameter. Additionally, the core assembly also increases the required resources because it pools samples. At least for one individual task, the required memory and time is much higher. This is something important to keep in mind, because also if you want to run it on larger data sets, you might want to provide a custom config file in order to adjust the resources required for your particular data set.
```
skrakau, line 68:
```suggestion
This just shows that even for smaller data sets, you cannot run this on your laptop. In general, one can say that it depends on the sequencing depths, the number of samples, the complexity of the underlying metagenome, and also on the applied tool and setting. For this it might be worth noting that both assembly tools are run by default but MegaHIT requires much fewer resources than SPAdes, and if you do not want to compute a hybrid assembly it might make sense to consider the `--skip_spades` parameter. Additionally, the co-assembly also increases the required resources because it pools samples. At least for one individual task, the required memory and time is much higher. This is something important to keep in mind, because also if you want to run it on larger data sets, you might want to provide a custom config file in order to adjust the resources required for your particular data set.
```

markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[12:53](https://www.youtube.com/watch?v=IiorfDHeoLo&t=773)
+With this we have seen how we can run the nf-core pipeline for modern metagenomic data sets. As I mentioned already at the beginning, it can also handle ancient DNA. For this James and Maxime added an ancient DNA validation sub-workflow. This is particularly interesting because, as we know at least, there's no other such pipeline which can handle ancient DNA. What this essentially does is that it performs identification of possible ancient contigs by modeling ancient DNA damage patterns, and then polishes the contigs in order to remove the errors that are caused by the presence of such ancient DNA damages in order to allow more unpaired downstream analysis. This might be interesting for some of you to know that this pipeline can also handle ancient metagenomic data analysis.
```
skrakau, line 71:
```suggestion
With this we have seen how we can run the nf-core/mag pipeline for modern metagenomic data sets. As I mentioned already at the beginning, it can also handle ancient DNA. For this James and Maxime added an ancient DNA validation sub-workflow. This is particularly interesting because, as far as we know at least, there's no other such pipeline which can handle ancient DNA. What this essentially does is that it performs identification of possible ancient contigs by modeling ancient DNA damage patterns, and then polishes the contigs in order to remove the errors that are caused by the presence of such ancient DNA damages in order to allow more unimpaired downstream analyses. This might be interesting for some of you to know that this pipeline can also handle ancient metagenomic data analysis.
```

markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
+
+[13:52](https://www.youtube.com/watch?v=IiorfDHeoLo&t=832)
+With this, I'm already at the end of my presentation, just a few words on the outlook. The next release James already prepared, it just requires one more review. It contains another optional binning tool, namely CONCOCT. It will also allow optionally the bin QC with CheckM and GUNC. For the midterm future, it would be also very nice if a functional annotation step could be added, so depending on the strategy, either using HUMAnN 3 or eggNOG, and also a standalone long read assembly option would be very nice by using, for example, the tool meta-flye, such that the pipeline could be also run without short read data.
```
skrakau, line 74:
```suggestion
With this, I'm already at the end of my presentation, just a few words on the outlook. The next release James already prepared, it just requires one more review. It contains another optional binning tool, namely CONCOCT. It will also allow optionally the bin QC with CheckM and GUNC. For the midterm future, it would be also very nice if a functional annotation step could be added, so depending on the strategy, either for example using HUMAnN 3 or eggNOG, and also a standalone long read assembly option would be very nice by using, for example, the tool meta-flye, such that the pipeline could be also run without short read data.
```

skrakau (approved):
Thanks @FranBonath for this! I have some small suggestions, other than that it looks good to me.

markdown/events/2023/bytesize_mag.md
```diff
@@ -21,3 +24,89 @@ It is our hope that these talks / videos will build an archive of training mater
...
...ted. By using additionally longer reads, this can improve the contiguity of such resulting assemblies. The pipeline also performs a genome binning step and optionally also a binning refinement step, then can taxonomically classify the resulting bins and also provides a comprehensive QC statistics. Furthermore, it can utilize sample-wise group information. This can be used for the core assembly. This is important if you have data sets where you know that certain strains are present across multiple samples, such as within longitudinal data sets. Because the core assembly can improve or increase the sequencing depth, this also allows to recover more lower abundant genomes. Additionally, the group information is also used for the computation of core abundances, which is used in the genome binning step. Furthermore, the pipeline also allows the handling of ancient DNA, because it's containing ancient DNA validation sub-workflow, which is rather specific for this pipeline. A previous version of this pipeline was already published at the beginning of this year in NAR Genomics and Bioinformatics, so if someone's interested in more details, you can also have a look at this application note.
```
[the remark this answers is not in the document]
skrakau: Soory, for some reason the suggested changes are not highlighted here (and for some of the following suggestions as well), not sure why ...

PR #1612, opened 2023-03-07
Add safety team to main aboutpage until governence is ready

jfy133:


christopher-hakkaart (approved):
Yup - good call! Lock it in

PR #1682, opened 2023-03-27
Fixed git push arguments on Adding new pipeline.

KarNair:
Fixed the push command to ```git push -all origin``` in the **Push To GitHub** section of **Adding New Pipelines** page to be up to date with git push arguments used on nf-core create docs.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1682"></a>


KarNair:
@nf-core-bot fix linting

jfy133:
Uhh I'm not actually sure of this myself :sweat_smile: @mashehu @mirpedrol @ewels  can you guys confirm that's best practise according to `nf-core create`?

mirpedrol:
Hi :) I think this depends a lot on the user and I wouldn't specify how to push things

KarNair:
@mirpedrol In that case, maybe both ```git push``` and ```git push --all origin``` could be added to the instructions along with a brief comment on which one to use when.

KarNair:
@nf-core-bot  fix linting

mirpedrol (approved):
LGTM :)

mashehu (commented):
sorry, but I think this will confuse quite a few people. Pushing several branches is quite an edge case and not a standard thing while working with nf-core, I guess.

jfy133:
Ooof yes, I agree, that's quite extreme... where does it say this in `nf-core create`?

jfy133 (commented):
I think actually given

> Make sure _not_ to initialise it with _any_ file, `README` or `LICENSE`: you just want an empty repository.

It implies that no other changes should be made at all before pushing

Therefore could you instead remove the note and put a warning between the 'make sure not' and 'once created' paragraphs saying 'Do not make any other changes/commits' etc'

_then_ you can add `--all`  to the command - as then the TEMPLATE, dev, and master are all pushed in once. 

Or even, add the warning and replace the `note` with something like 'we don't normally recommend using `--all` however do it in this case to have all the nf-core-required branches to be ready on the remote' or something like that?

What do you think @mashehu

mashehu:
agree with @jfy133 and would go with his suggestions.

maxulysse (approved):

maxulysse:
not sure about that, maybe we can just drop it

ewels:
Also the docs above this are way too lenient about whether to use the template or not. It's now a [requirement in the guidelines](https://nf-co.re/docs/contributing/guidelines/requirements/use_the_template).

I'll submit a new PR to address these issues.

PR #1692, opened 2023-03-27
YouTube embedding

vlebars:
Tried to add YouTube video embedding, need to merge the request to see if the change actually works on the website


christopher-hakkaart:
You can preview changes in a gitpod environment or by using a [local servier](https://github.com/nf-core/nf-co.re#running-a-local-server). The video looks looks good. Is there a consensus on where on the page it should sit? E.g., at the bottom under other sessions?

vlebars:
Indeed I looked to other event pages and the video was not at the same place and size, I did a new pull request ( #1699 ) to embed the video like on other pages, so all pages look the same.

PR #1699, opened 2023-03-28
Added video embedding on four 2022 event pages

vlebars:
I added video embedding on four 2022 event page, with the same template as the other event with working embedding.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1699"></a>


vlebars:
I reopened a new pul request so the videos look the same as on the other pages. I changed all four requested pages.

FranBonath (approved):
looks good :)

PR #1717, opened 2023-03-28
Teams notifications

atrigila:
Addressing https://github.com/nf-core/nf-co.re/issues/1455 but I need feedback if this is correct and also if this page is appropriate. 

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1717"></a>

### DIFF
--- a/markdown/usage/introduction.md
+++ b/markdown/usage/introduction.md
@@ -81,6 +81,7 @@ Nextflow works best when you have an active internet connection, as it is able t
 
 - Hyphens matter! Core Nextflow command-line options use one (`-`), whereas pipeline-specific parameters use two (`--`)
 - Specify `--email your@email.com` to receive emails when your pipeline completes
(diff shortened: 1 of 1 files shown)


christopher-hakkaart:
Is Slack also supported? I haven't checked but it had a feeling it was?

matthdsm:
Yes, slack should also be supported. If I'm not mistaken @Midnighter added that part, but it _should_  amount to the same thing

matthdsm:
The selection for slack vs teams is done based on the url
https://github.com/nf-core/tools/blob/master/nf_core/pipeline-template/lib/NfcoreTemplate.groovy#L208

So the method for enabling the notification stays the same.

matthdsm:
the docs for slack are here
https://api.slack.com/messaging/webhooks

So if you were to add that next to the teams ones I think it's good to go

matthdsm (approved):
Looks good to me 😄 Good work

atrigila:
Thanks for the review @matthdsm @muffato @christopher-hakkaart :)

muffato (approved):

PR #1743, opened 2023-04-05
Link to mulled name tool

mirpedrol:
add a link to the helper tool to obtain mulled container names
follow-up from https://github.com/nf-core/tools/pull/2199


Midnighter (approved):
Some unrelated docs changes it seems but looks good to me!

mirpedrol:
oh, that was me committing changes in random branches and never pushing them 😅 thanks!

PR #1780, opened 2023-05-31
Fix header and footer links

mashehu:


package-lock.json
```diff
@@ -33,6 +30,9 @@
             },
             "devDependencies": {
                 "@prettier/plugin-php": "^0.18.4",
+                "bootstrap": "5.1.1",
```
mirpedrol, line 33: This version has `^` in other places, should it have it here too?

mirpedrol (approved):

PR #1796, opened 2023-06-16
Bytesize transcript python package

FranBonath:


markdown/events/2023/bytesize_python_packaging.md
```diff
@@ -36,4 +37,90 @@ This is a good thing to do for a few reasons:
...
+
+(question cont.) I'll just drop the link in the chat. If someone doesn't know what we're talking about.
+
+(answer cont) That links for cookie cutter itself, right, which is just like a generic templating tool. There are cookie cutter projects which people have created like template repositories. Specifically for python, if that makes sense.
+
+[26:09](https://www.youtube.com/watch?v=hOuS6mXCwhk&t=1569)
+(question) We do have another question in the chat. Someone is asking why not pyprojects.toml.
```
ewels, line 114:
```suggestion
(question) We do have another question in the chat. Someone is asking why not `pyprojects.toml`?
```

markdown/events/2023/bytesize_python_packaging.md
```diff
@@ -36,4 +37,90 @@ This is a good thing to do for a few reasons:
...
...a structured file format, and the other one is pyproject.toml, which is the newer and better way of doing things. Pyproject.toml is nice because it's also a standard for many other python tools with configs. If you want to use black to lint your code, which you should, because black is amazing, you'll put your settings in pyproject.toml. If you use, I don't know, mypi for type linting or any of these flake8 tools or whatever and it will be linting tools and stuff. They all stick their settings in pyproject.toml, which is great because you have one config file for everything to do with your python project. That is much nicer and you can also do all of your setuptools python stuff in there. There are a couple of things which I found I think are missing. Correct me if I'm wrong, I don't think you can point it to a requirements.txt file for all requirements. It's quite useful having that file sometimes, maybe it doesn't matter... I think the setup tools website says it's like in beta and it might change, so I thought I'd play it safe today and go for setup.cfg, which is newish, but fairly safe. But yeah, pyproject.toml is, if you can make it work for you, probably a nicer way to do it.
```
ewels, line 116:
```suggestion
(answer) This is something else I was debating on the start. This is a bit of history here. When I started creating my first python projects you always used that `setup.py` file and you still can. It's a bit like how Nextflow config files are just a groovy script, where you can do whatever you like. `setup.py` is the same. It's just a python script, where you can do whatever you like.Which is wonderful and horrifying! Slowly over the last... the python community moves slowly... so for the last many years, there's been a move away from that way of doing things into more standardized file types and there are two which are being used: there's a `setup.config` file, which is exactly the same thing but in a structured file format, and the other one is `pyproject.toml`, which is the newer and better way of doing things. `pyproject.toml` is nice because it's also a standard for many other python tools with configs. If you want to use black to lint your code, which you should, because black is amazing, you'll put your settings in `pyproject.toml`. If you use, I don't know, mypy for type linting or any of these flake8 tools or whatever and it will be linting tools and stuff. They all stick their settings in `pyproject.toml`, which is great because you have one config file for everything to do with your python project. That is much nicer and you can also do all of your setuptools python stuff in there. There are a couple of things which I found I think are missing. Correct me if I'm wrong, I don't think you can point it to a `requirements.txt` file for all requirements. It's quite useful having that file sometimes, maybe it doesn't matter... I think the setup tools website says it's like in beta and it might change, so I thought I'd play it safe today and go for `setup.cfg`, which is newish, but fairly safe. But yeah, `pyproject.toml` is, if you can make it work for you, probably a nicer way to do it.
```

ewels (approved):
Amazing work 🙌🏻

PR #1797, opened 2023-06-21
Add tools api pages

mashehu:
ran `lazydocs . --output-path docs/md` in the tools repo and then copied them for testing purposes into the content directory (in the future they will be just symlinked there).


mashehu:
Fixed in #2268

PR #1799, opened 2023-06-22
use paginated github query

mashehu:


mashehu:
@nf-core-bot fix linting

update_pipeline_details.php
```diff
@@ -42,14 +43,37 @@ function get_gh_api($gh_api_url) {
...
-        die("\n-------- END ERROR " . date('Y-m-d h:i:s') . " --------\nCould not fetch $gh_api_url \n");
+    $first_page = true;
+    $next_page = false;
+    $res = [];
+    while ($first_page || $next_page) {
+        // reset loop vars
+        $first_page = false;
+        // Get GitHub API results
+        if ($next_page) {
+            $gh_api_url = $next_page;
```
ewels, line 54: Where is this variable coming from? Just want to check that we're not breaking stuff by overwriting it here..

update_pipeline_details.php
```diff
@@ -42,14 +43,37 @@ function get_gh_api($gh_api_url) {
...
+            die("\n-------- END ERROR " . date('Y-m-d h:i:s') . " --------\nCould not fetch $gh_api_url \n");
+        }
+
+        array_push($res, ...$tmp_results);
+        // Look for URL to next page of API results
+        $next_page = false;
+        $m_array = preg_grep('/rel="next"/', $http_response_header);
+
+        if (count($m_array) > 0) {
+            preg_match('/<([^>]+)>; rel="next"/', array_values($m_array)[0], $matches);
```
ewels, line 70: Oof, this is the easiest way? There's no way to get the headers into an associative array or anything? Fair enough..

ewels (approved):
Two minor comments / queries. Otherwise LGTM ✅

update_pipeline_details.php
mashehu: it's the input parameter to this github query function

update_pipeline_details.php
mashehu: I could simply count the page up.

PR #1804, opened 2023-06-26
add extra for empty defaults

mashehu:
this should fix an error when `'default'` has already been removed from the object

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1804"></a>


mirpedrol (approved):
LGTM!

PR #1808, opened 2023-06-28
Add Phil's suggestions

mashehu:
- [x] Header+Footer


src/components/ThemeSwitch.svelte
```diff
@@ -1,55 +1,78 @@
...
-        const getPreferredTheme = () => {
-            if (storedTheme) {
-                return storedTheme;
-            }
-
-            return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
-        };
     </script>
 </svelte:head>
+<div class="dropdown">
```
ewels, line 17:
```suggestion
<div class="dropdown-center">
```

ewels (commented):
Looking good! I guess this is still a WIP, but still - quick thoughts:

- [x] It'd be nice to have a vertical bar separator between the dark mode switch + socials (and maybe the Join button)
- [x] Not all socials have a tooltip. Actually I'm not sure that these need tooltips, might be a bit visually noisy. If we keep them, add a delay?
- [x] I think keep the socials monochrome. Drop opacity as they are brighter than everything else currently. Then can bring up opacity on hover.
- [ ] Footer looks much better, but vertical alignment / equal padding is off a bit
- [ ] Maybe supporters should be monochrome too?

src/components/ThemeSwitch.svelte
mashehu: doesn't work. not sure why.

PR #1814, opened 2023-06-30
First pass at corralling guidelines into a single folder

robsyme:
I'm still checking for broken links, but this moves things like modules and subworkflows into the "Guidelines" section of the website.

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1814"></a>


jfy133:
When I think about this being a collection of quick-reference stuff, I would also personally put `Test data guidelines`, `Pipeline Release Review Guidelines` and also `Release checklist` to the list :)

I also think something has gone wrong here:

image

PR #1832, opened 2023-07-06
Basic training Sept 2023

mribeirodantas:
This is a draft, a work in progress. It's still not completely clear how much we'll change from the March training. Adding more recordings? Simply directing attendees to the playlists and being live on Slack to reply? It's still being discussed.


ewels:
Closed in favour of https://github.com/nf-core/nf-co.re/pull/1834

PR #1863, opened 2023-07-13
transcript bytesize nf-validation

FranBonath:
now with a PR to the new webpage!


mirpedrol:
ah it does! all good then 😄

PR #1874, opened 2023-07-14
trying to fix `nf-core launch`

mashehu:


ewels:
As discussed in Slack, fixed by changing the `.htaccess` redirects on the oldsite for trailing slashes. I added the `200` code in https://github.com/nf-core/website/commit/75aebd137efc056d40cd06be7474fb9eecced5b1

Leaving this open in case you want to delete the `launch.astro` file still..

PR #1884, opened 2023-07-19
reduce built time through content collections

mashehu:


mashehu:
@nf-core-bot fix linting

mashehu:
doesn't work due to memory leak 🤷🏻

PR #1887, opened 2023-07-21
first try with view transitions

mashehu:


mashehu:
On hold until https://github.com/withastro/astro/issues/7765 is fixed

PR #1891, opened 2023-07-24
Add playwright testing

mashehu:
This tests:
- that the homepage loads with correct title
- that pipeline name redirects work
- that SSR is working in the aws previews
- that the pipeline listing is working


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  run_test:
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
```
alvaromartmart, line 19: you can speed up this step across different runs if you make the github actions workflow cache npm dependencies, check out
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: npm ci
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Waiting for 200 from the Netlify Preview
+      uses: jakepartusch/wait-for-netlify-action@v1.4
+      id: waitFor200
+      with:
+        site_name: "nf-core"
+        max_timeout: 1000 # 10 Minutes
```
alvaromartmart, line 28:
```suggestion
        max_timeout: 600 # 10 Minutes
```

tests/nf-core.spec.ts
alvaromartmart, line 1: A couple of suggestions:
- it is recommended to use [page object models](https://playwright.dev/docs/pom) to encapsulate common locators and functionality as it makes easier to create, read and maintain the tests
- Playwright's [soft assertions](https://playwright.dev/docs/test-assertions#soft-assertions) can be helpful for tests where you check several things within a page, so the test is allowed to "cover more ground" even if the outcome is still a failure.

.github/workflows/build-json-files-and-md-cache.yml
```diff
@@ -23,9 +23,23 @@ jobs:
...
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
 
+      - if: ${{ steps.cache-npm.outputs.cache-hit != 'true' }}
+        name: Install dependencies
+        run: npm install
```
alvaromartmart, line 42:
```suggestion
        run: npm ci
```

PR #1894, opened 2023-07-24
Update nf-core logos

mribeirodantas:
Fix buggy dash in teh nf-core logo


mribeirodantas:
I don't know much about Astro, so I'm trying to fix the image size without bugging the dash 😅  WIP

mashehu:
I don't see this bug on any of my browsers. can you check with the updated svgs?

mribeirodantas:
It's working now, @mashehu 🥳.

As for the bug, you should be able to reproduce it on Safari (Version 17.0 (19616.1.20.11.3) here)

mashehu (approved):

PR #1906, opened 2023-08-02
Use Netlify _redirects for Slack invite URL

ewels:


public/_redirects
```diff
@@ -1,3 +1,6 @@
+# Slack invite URL
+/join/slack https://join.slack.com/t/nfcore/shared_invite/zt-zj2p3l6d-oLK8OcMGXYuPra0ye1ptdg 200
```
mashehu, line 2: Not sure if it needs to be a `200` redirect, but fine as it is.

mashehu (approved):

public/_redirects
ewels: I was wondering if it would hide the ugly URL and act as a CDN, but no.. But it does work at least!

ewels:
```suggestion
/join/slack https://join.slack.com/t/nfcore/shared_invite/zt-zj2p3l6d-oLK8OcMGXYuPra0ye1ptdg
```

PR #1913, opened 2023-08-07
Disable AWS File preview for large HTML documents

MatthiasZepper:
Unfortunately, larger HTML documents will not render in the AWS file preview on the website, [but throw a Netlify error](https://nf-co.re/viralrecon/2.6.0/results/viralrecon/results-3731dd3a32a67a2648ea22c2bd980c224abdaee2/platform_illumina/pipeline_info/?file=execution_report_2023-03-23_22-44-40.html). For an example, see e.g. the [8 MB execution reports in Viralrecon bucket](https://nf-co.re/viralrecon/2.6.0/results/viralrecon/results-3731dd3a32a67a2648ea22c2bd980c224abdaee2/platform_illumina/pipeline_info/). 

The reason seems to be that Netlify like AWS Lambda uses 6291556 bytes as payload size limit for some functions. Thus, this PR introduces an additional size check, that will disable a preview when the HTML size limit is nearing this limit. Anything below 5.5 MB is presumed safe across-the-board, and for HTML files larger than that, the content is encoded and size-checked prior to display. The function to check the size has been adopted from [Zac Charles' _Deep Dive: Lambda’s Response Payload Size Limit_ article](https://zaccharles.medium.com/deep-dive-lambdas-response-payload-size-limit-8aedba9530ed). 

_Of note: With [Response Streaming](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html), handling larger payloads may be possible, but I don't think that this extra effort is justified given that the files can be downloaded and viewed easily._


MatthiasZepper:
This does not solve the issue, unfortunately. In contrast to local tests, the function still crashes in the deployment preview. I will have to come up with something else...

src/components/pipeline/AWSFilePreview.astro
```diff
@@ -28,10 +28,15 @@ const hasPreview = [
     'gif',
     'svg',
 ].includes(extension);
-const tooBig = extension !== 'html' && file.size > 10000000;
+let tooBig = extension !== 'html' && file.size > 10000000;
```
mashehu, line 31:
```suggestion
let tooBig =  file.size > 5500000;
```
Why not simply like this (because we will probably have the same problem with other file types?

Then we don't need the exception below.

MatthiasZepper: The approach chosen was based on the false? assumption that it would only fail when the `<iframe>` component gets rendered, while fetching larger `content` asynchronously would still succeed. 

Since you had already written an elaborate `const tooBig = extension !== 'html' && file.size > 10000000` if-clause for fetching content, I presumed this part of the code was already well tested with files up to this size.

For now, I agree that simply lowering the limit will be the most straightforward solution. If we feel like overengineering at some point, using readable streams and converting chunks from them to blobs would technically allow handling bigger data in the browser.

mashehu: Nah, I just guesstimated based on rendering time (didn't run into that error locally). Let's just lover the limit and have the same for all and we should be good for now. Maybe when we see too many Multiqc reports bigger than that size we can start looking for alternative approaches.

mashehu (approved):

src/components/pipeline/AWSFilePreview.astro
MatthiasZepper: Convincing argument! I agree.

PR #1938, opened 2023-08-21
Make the governance page name badges better

ewels:
Governance page username badge improvements:

* Made a new sub-component and massively reduced markup in the MDX
* Move maintainers' affiliation info into the component
* Removed maintainers list

Maintainers list is now much, much easier to read:

CleanShot 2023-08-22 at 00 26 41@2x

Bed time now, but remaining - can't get the text to align properly:

CleanShot 2023-08-22 at 00 25 52@2x

Should be a nice tight-wrapping circle, like this:

CleanShot 2023-08-22 at 00 26 03@2x

Help to fix this appreciated 🙏🏻

PR #1940, opened 2023-08-22
Add minHeadingDepth option for ToC

ewels:
Pre-commit rejected my commit and I didn't noticed, thought that this was appended onto the end of https://github.com/nf-core/website/pull/1938

Anyway, here it is.. :)

### DIFF
--- a/src/components/sidebar/SidebarToc.svelte
+++ b/src/components/sidebar/SidebarToc.svelte
@@ -10,7 +10,11 @@
         hidden?: boolean;
     }[];
 
+    export let minHeadingDepth: number = 1;
     export let maxHeadingDepth: number = 4;
+
+    // filter out headings that are lower than min_heading_depth
+    headings = headings.filter((h) => h.depth >= minHeadingDepth);
 
     // filter out headings that are higher than max_heading_depth
     headings = headings.filter((h) => h.depth <= maxHeadingDepth);
--- a/src/content/about/governance.mdx
+++ b/src/content/about/governance.mdx
@@ -1,6 +1,7 @@
 ---
 title: Governance
 description: Governance of the nf-core community
+minHeadingDepth: 2
 maxHeadingDepth: 2
 ---
 
--- a/src/content/config.ts
+++ b/src/content/config.ts
@@ -34,6 +34,7 @@ const about = defineCollection({
         title: z.string(),
         description: z.string(),
         md_github_url: z.string().url().optional(),
+        minHeadingDepth: z.number().optional(),
         maxHeadingDepth: z.number().optional(),
     }),
 });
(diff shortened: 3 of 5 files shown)


mashehu (approved):

src/components/sidebar/SidebarToc.svelte
```diff
@@ -10,23 +10,29 @@
...
+    export let minHeadingDepth: number = 1;
     export let maxHeadingDepth: number = 4;
 
+    // filter out headings that are lower than min_heading_depth
+    headings = headings.filter((h) => h.depth >= minHeadingDepth);
+
     // filter out headings that are higher than max_heading_depth
     headings = headings.filter((h) => h.depth <= maxHeadingDepth);
 
-    const min_heading_depth = Math.min(...headings.map((h) => h.depth));
```
ewels, line 18: @mashehu i left this in, as I think it's doing a different thing (despite the name). The config could be set to 1 as default, but the page could only have h3 and down.. So we still need to check the minimum on the page, right?

src/components/sidebar/SidebarToc.svelte
```diff
@@ -10,23 +10,29 @@
...
     onMount(() => {
         // set the first heading as active on initial load
-        if (!$currentHeading) {
-            currentHeading.set(headings[0]?.slug);
-        }
+        activeHeading = $currentHeading || headings[0].slug;
         currentHeading.subscribe((slug) => {
+            //check if any heading has the same slug as the currentHeading
+            const heading = headings.find((h) => h.slug === slug);
+            console.log('currentHeading', slug, heading);
```
ewels, line 34: Debug left in

src/components/sidebar/SidebarToc.svelte
mashehu: Ah, yes, true. Will merge the two.

src/components/sidebar/SidebarToc.svelte
```diff
@@ -10,23 +10,30 @@
...
     export let maxHeadingDepth: number = 4;
 
+    // filter out headings that are lower than min_heading_depth
+    headings = headings.filter((h) => h.depth >= minHeadingDepth);
+
     // filter out headings that are higher than max_heading_depth
     headings = headings.filter((h) => h.depth <= maxHeadingDepth);
 
-    const min_heading_depth = Math.min(...headings.map((h) => h.depth));
+    minHeadingDepth = Math.min(...headings.map((h) => h.depth));
```
ewels, line 22: This now overwrites the supplied config, I think? Needs a different variable name and then a second `max` call on the two..

mashehu: No, it should be fine. Because the supplied config is already filtering out higher level headings, so the `min` hear can't be lower then minHeadingDepth.

ewels: Right, yup 👍🏻

PR #1941, opened 2023-08-22
switch to rehype-mermaid to make graphs SSR

mashehu:


mashehu:
This solution needs to run playwright during build and I can't get that working on netlify 😞

PR #1954, opened 2023-08-24
combine filter logic for events and for pipelines

mashehu:


mashehu:
@nf-core-bot fix linting

PR #1987, opened 2023-09-21
update meta map guidelines for reference file inputs

ramprasadn:
As discussed in the nf-core maintainers meeting, I have updated the guidelines for creating new modules to add a meta map to channel definitions that carry reference file inputs.


maxulysse:
@ramprasadn there are some conflicts.
If you could solve them that would be fabulous, and we could finally merge that one

src/content/docs/contributing/modules.md
```diff
@@ -367,9 +367,11 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
 
-2.  Non-file mandatory arguments or arguments needed to modify the command to make the module run with no error, SHOULD be provided as value channels (for example `lib_type` in [salmon/quant](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf)) - see 'Input/output options' below.
+2.  All channel definitions that carry reference inputs, like fasta, intervals, and sequence dictionary SHOULD contain a meta map (for example, [see here](https://github.com/nf-core/modules/blob/master/modules/nf-core/gatk4/collectreadcounts/main.nf)). We do this because there are pipelines that support multiple genomes, and they use meta information to distinguish between different genomes.
 
-3.  All _non-mandatory_ command-line tool _non-file_ arguments MUST be provided as a string via the `$task.ext.args` variable.
+3.  Non-file mandatory arguments or arguments needed to modify the command to make the module run with no error, SHOULD be provided as value channels (for example `lib_type` in [salmon/quant](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf)) - see 'Input/output options' below.
```
cmatKhan, line 372: I think rather than referring to 'non-file' in the current points 3. and 4. ... It would be better to separate that into its own bullet point. eg,

3. `file` arguments, whether or not they are mandatory, __should be__ provided as input channels (for example, the `fai` argument in [samtools faidx](https://nf-co.re/modules/samtools_faidx/))

 I'm torn as to whether that should come before or after the current point 3.

On the one hand, it seems like mandatory arguments are more important, and hence addressed first.

On the other, it is probably a more necessary reminder to include non-mandatory file arguments, which the author themselves may never use, as module input.

mashehu:
This PR is still against the old website structure. @ramprasadn would you mind opening a new PR with this changes applied to the restructured docs, so we can get this merged?

PR #1991, opened 2023-09-25
fix too strict toc filtering

mashehu:


ewels (approved):
oops, my bad. Thanks for finding + fixing 🙏🏻

PR #2099, opened 2023-11-09
Add using closure in modules.conf to process directives that use params.*

mahesh-panchal:
Updates docs to include that process directives which include `params.*` should be supplied as closures.


mashehu (approved):

mashehu:
@nf-core-bot fix linting

PR #2102, opened 2023-11-10
try to fix update_stats page

mashehu:
switching more to database based data and disabling twitter


ewels (approved):
I will merge blind, as I figure it's already broken so nothing to lose 😅

PR #2130, opened 2023-11-29
Added custom script stuff

RaqManzano:
I added something to start with, let me know what you think and if this is what you had in mind @FranBonath.

### DIFF
--- a/src/content/docs/contributing/nf_core_basic_training.md
+++ b/src/content/docs/contributing/nf_core_basic_training.md
@@ -160,6 +160,12 @@ nf-core lint
 
 # Adding Modules to a pipeline
 
+A module is a  single `process` built to be reusable and self-contained so it can be used within different Nextflow pipelines. They encapsulate a specific function or task, for example running a single tool such as [`FastQC`](https://github.com/nf-core/modules/blob/master/modules/nf-core/fastqc/main.nf). You can import and use modules like functions in a Nextflow subworkflow, this makes your workflow more readable and maintainable.
+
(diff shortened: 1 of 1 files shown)


src/content/docs/contributing/nf_core_basic_training.md
```diff
@@ -214,6 +220,83 @@ Open ./modules/local/demo/module.nf and start customising this to your needs whi
...
+    args = argparser()
+    maf2bed(maf_file=args.mafin, bed_file=args.bedout, extra=args.extra)
+
+
+if __name__ == "__main__":
+    main()
+
+```
+
+Now, let's go back to our module:
```
FranBonath, line 263:
```suggestion
Since `maf2bed.py` is in the `bin` directory we can directory call it in the script block of our new module [module name] . You only have to be careful with how you call variables (some explanations on when to use `${variable}` vs. `$variable`):
```

src/content/docs/contributing/nf_core_basic_training.md
```diff
@@ -214,6 +220,83 @@ Open ./modules/local/demo/module.nf and start customising this to your needs whi
...
+    script: // This script is bundled with the pipeline in bin
+    def args = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}"
+
+    """
+maf2bed.py --mafin $maf --bedout ${prefix}.bed
+    """
+```
+
+We are now able to use our custom script through the module
```
FranBonath, line 296: Here I would also add how to add the include statement and the process to our `demotest.nf` workflow and how to add the maf input as input tuple together with meta.

FranBonath (approved):

FranBonath:
The core group has a retreat on the 5th of February, would be nice if we had this PR merged. It doesn't have to be completed or anything, I would just like to show how far we have come already in one file/preview.

maxulysse:
There are a couple of conflicts that prevent this PR to be merged...

PR #2133, opened 2023-12-01
 modules addition in basic training

FriederikeHanssen:


src/content/docs/contributing/nf_core_basic_training.md
```diff
@@ -681,19 +685,86 @@ INFO     Use the following statement to include this module:
...
+    //
+    FASTQC (
+        INPUT_CHECK.out.reads
+    )
+    ch_versions = ch_versions.mix(FASTQC.out.versions.first())
+
+    SALMON_INDEX()
+```
+
+Now we are still missing an input for our module. In order to build an index, we require the reference fasta. Luckily, the template pipeline has this already all configured, and we can access it by just using `params.fasta` and `view` it to  insppect the channel content. (We will see later how to add more input files.)
```
FranBonath, line 728: Could we add a section where we show how to determine which input channels are required?

src/content/docs/contributing/nf_core_basic_training.md
```diff
@@ -681,19 +685,86 @@ INFO     Use the following statement to include this module:
...
+
+include { FASTQC  } from '../modules/nf-core/fastqc/main'
+include { MULTIQC } from '../modules/nf-core/multiqc/main'
+include { SALMON_INDEX } from '../modules/nf-core/salmon/index/main'
+
+```
+
+This makes the module now available in the workflow script and it can be called with the right input data.
+
+<!-- TODO/TO DISCUSS here the user now needs to know about how to get their fasta. We could do this here or add a new point for this above -->
```
FranBonath, line 708: My logic would be having separate points:
- install the module
- add the inclusion statement and the process
- identify which inputs are required, add (if necessary) a new param, add the input channels with optional channel manipulation
- check the output channels to be used in another process

FranBonath (approved):
very good start!

FranBonath:
@FriederikeHanssen do you think we can merge this PR until the retreat? Independent of how far we come, because then we could show one preview page instead of several PRs.

FranBonath:
@nf-core-bot fix linting please :)

PR #2137, opened 2023-12-05
Add guidance on bad exit codes

mahesh-panchal:
Add guidance on how to deal with tools with unexpected error codes.

### DIFF
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -845,7 +845,7 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
 11. Occasionally, some tools do not exit with the expected exit code 0 upon successful use of the tool. In these cases one can use
     the `||` operator to run another useful command when the exit code is not 0 (for example, testing if a file is not size 0).
 
-    ```nextflow {4}
+    ```nextflow {6}
     script:
     """
     tool \\
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -845,7 +845,7 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
 11. Occasionally, some tools do not exit with the expected exit code 0 upon successful use of the tool. In these cases one can use
     the `||` operator to run another useful command when the exit code is not 0 (for example, testing if a file is not size 0).
 
-    ```nextflow
+    ```nextflow {4}
     script:
     """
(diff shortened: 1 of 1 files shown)


src/content/docs/contributing/modules.md
```diff
@@ -842,6 +842,23 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     """
     ```
 
+11. Occasionally, some tools do not exit with the expected exit code 0 upon successful use of the tool. In these cases one can use
+    the `||` operator to run another useful command when the exit code is not 0 (for example, testing if a file is not size 0).
+
+    ```nextflow
```
mashehu, line 848:
```suggestion
    ```nextflow {4}
```

src/content/docs/contributing/modules.md
```diff
@@ -842,6 +842,23 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
+    tool \\
+      --input $input \\
+      --summary ${prefix}.summary \\
+      || test -s ${prefix}.summary
+    """
+    ```
+
+    See the [Bash manual on file operators](https://tldp.org/LDP/abs/html/fto.html) for examples of properties of files which could be tested.
+    Alternate suggestions include using `grep -c` to search for a valid string match, or other tool which will appropriately
+    error when expected output is not successfully created.
```
mashehu, line 860:
```suggestion
    error when the expected output is not successfully created.
```

mashehu (approved):

src/content/docs/contributing/modules.md
mahesh-panchal: What does this do?

mashehu: highlight a specific line in the codeblock: https://nf-co.re/docs/contributing/website_markdown#highlight-lines

src/content/docs/contributing/modules.md
```diff
@@ -842,6 +842,23 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     """
     ```
 
+11. Occasionally, some tools do not exit with the expected exit code 0 upon successful use of the tool. In these cases one can use
+    the `||` operator to run another useful command when the exit code is not 0 (for example, testing if a file is not size 0).
+
+    ```nextflow {4}
```
mahesh-panchal, line 848:
```suggestion
    ```nextflow {6}
```

PR #1722, opened 2023-03-29
Add tips for version extraction

LouisLeNezet:
fix #1033 

<a href="https://gitpod.io/#https://github.com/nf-core/nf-co.re/pull/1722"></a>

PR #2141, opened 2023-12-06
Include again string version extraction tips

mahesh-panchal:
It seems the tips to extract the version from the tool output was lost in transition to the new website. https://github.com/nf-core/website/pull/1722

Here's a slightly revised version.

### DIFF
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -746,7 +746,7 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
         fastqc: \$( fastqc --version | sed -e "s/FastQC v//g" )
-        samtools: \$( samtools --version 2>&1 | sed 's/^.*samtools //; s/Using.*\$// )
+        samtools: \$( samtools --version |& sed '1!d ; s/samtools //' )
     END_VERSION
     ```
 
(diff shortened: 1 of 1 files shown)


src/content/docs/contributing/modules.md
```diff
@@ -761,6 +761,25 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
+    tool --version | sed '1!d'
+    ```
+
+    - `sed '1!d'` Extracts only line 1 of the output printed by `tools --version`.
+    - If the line extraction doesn't work, then it's likely the version information is written to stderr, rather than stdout.
+      In this case capture stderr using `|&` which is shorthand for `2>&1 |`.
+    - `sed 's/pattern/replacement/'` can be used to remove parts of a string. `.` matches any character, `+` matches 1 or more times.
+    - You can separate `sed` commands using `;`. Often the pattern : `sed filter line ; replace string` is enough to get the version number.
+    - It is not necessary to use `echo`, `head`, `tail`, or `grep`.
+    - Use `|| true` for tools that exit with a non-zero error code: `command --version || true` or `command --version | sed ... || true`.
```
mashehu, line 779:
```suggestion
    - Use `|| true` for tools that exit with a non-zero error code: `command --version || true{:bash}` or `command --version | sed ... || true{:bash}`.
```
for inline code highlighting: https://rehype-pretty-code.netlify.app/#highlight-inline-code

src/content/docs/contributing/modules.md
```diff
@@ -761,6 +761,25 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
+    - The line to process can also be selected using a pattern instead of a number: `sed -nr '/pattern/p'`.
```
mashehu, line 780:
```suggestion
    - The line to process can also be selected using a pattern instead of a number: `sed -nr '/pattern/p'{:bash}`.
```

mashehu (approved):

PR #2169, opened 2024-01-11
don't use pip cache

mashehu:
caching of setup-python works only if there is a requirements.txt


mashehu:
@nf-core-bot fix linting

PR #2217, opened 2024-01-22
Add deprecation guidelines

mahesh-panchal:
Adds guidelines on how to deprecate a module.

### DIFF
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -1406,10 +1406,10 @@ check_conda_channels // boolean: check conda channels
 Sometimes modules or subworkflows become outdated and need to be deprecated (available, but no longer recommended).
 These modules or subworkflows should not be deleted as they could be used in private repositories, or used on other
 platforms. The recommended procedure is, once the alternative is available on nf-core modules, add a comment to the
-top of the module code saying this module is deprecated, and an `assert`` in the code body to print a deprecation
+top of the module code saying this module is deprecated, and an `assert` in the code body to print a deprecation
 message like so:
 
-```nextflow
+```groovy
 /*
 WARNING: This module has been deprecated. Please use nf-core/modules/path/to/new/module
 
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -1401,6 +1401,30 @@ outdir               //    path: base directory used to publish pipeline results
 check_conda_channels // boolean: check conda channels
 ```
 
+## Deprecating a module
+
+Sometimes modules or subworkflows become outdated and need to be deprecated (available, but no longer recommended).
+These modules or subworkflows should not be deleted as they could be used in private repositories, or used on other
+platforms. The recommended procedure is, once the alternative is available on nf-core modules, add a comment to the
+top of the module code saying this module is deprecated, and an `assert`` in the code body to print a deprecation
+message like so:
+
+```nextflow
+/*
+WARNING: This module has been deprecated. Please use nf-core/modules/path/to/new/module
+
+Reason:
+Add why this module is no longer fit for purpose.
+*/
+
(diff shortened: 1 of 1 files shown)


src/content/docs/contributing/modules.md
```diff
@@ -1401,6 +1401,33 @@ outdir               //    path: base directory used to publish pipeline results
...
 
+## Deprecating a module
+
+Sometimes modules or subworkflows become outdated and need to be deprecated (available, but no longer recommended).
+These modules or subworkflows should not be deleted as they could be used in private repositories, or used on other
+platforms. The recommended procedure is, once the alternative is available on nf-core modules, add a comment to the
+top of the module code saying this module is deprecated, and an `assert`` in the code body to print a deprecation
+message like so:
+
+```nextflow
```
mashehu, line 1412:
```suggestion
```groovy
```

src/content/docs/contributing/modules.md
```diff
@@ -1401,6 +1401,33 @@ outdir               //    path: base directory used to publish pipeline results
 check_conda_channels // boolean: check conda channels
 ```
 
+## Deprecating a module
+
+Sometimes modules or subworkflows become outdated and need to be deprecated (available, but no longer recommended).
+These modules or subworkflows should not be deleted as they could be used in private repositories, or used on other
+platforms. The recommended procedure is, once the alternative is available on nf-core modules, add a comment to the
+top of the module code saying this module is deprecated, and an `assert`` in the code body to print a deprecation
```
mashehu, line 1409:
```suggestion
top of the module code saying this module is deprecated, and an `assert` in the code body to print a deprecation
```

mashehu (approved):

src/content/docs/contributing/modules.md
```diff
@@ -1401,6 +1401,33 @@ outdir               //    path: base directory used to publish pipeline results
...
 
+## Deprecating a module
+
+Sometimes modules or subworkflows become outdated and need to be deprecated (available, but no longer recommended).
+These modules or subworkflows should not be deleted as they could be used in private repositories, or used on other
+platforms. The recommended procedure is, once the alternative is available on nf-core modules, add a comment to the
+top of the module code saying this module is deprecated, and an `assert` in the code body to print a deprecation
+message like so:
+
+```groovy
```
mashehu, line 1412:
```suggestion
```groovy title="main.nf"
```

src/content/docs/contributing/modules.md
```diff
@@ -1401,6 +1401,33 @@ outdir               //    path: base directory used to publish pipeline results
...
+
+Reason:
+Add why this module is no longer fit for purpose.
+*/
+
+process OLD_MODULE {
+  ...
+
+  script:
+  assert true: "This module has been deprecated. See code for the reason and an alternative."
```
JoseEspinosa, line 1424: Couldn't we put the reason in a variable and output it in the message directly?

JoseEspinosa: 👍 exactly, so that the user does not need to check the reason in the main of the module

mahesh-panchal:
Looks like this when run:
```console
$ nextflow run main.nf 
N E X T F L O W  ~  version 23.10.1
Launching `main.nf` [friendly_goldberg] DSL2 - revision: f5ac28c9ad
[-        ] process > TASK -
ERROR ~ Error executing process > 'TASK'

Caused by:
  
WARNING: This module has been deprecated. Please use nf-core/modules/path/to/new/module

Reason:
This module is no longer fit for purpose because ...

. Expression: false -- Check script 'main.nf' at line: 18

Source block:
  assert false: deprecation_message
  """
      """

Tip: you can replicate the issue by changing to the process work dir and entering the command `bash .command.run`

 -- Check '.nextflow.log' file for details


```

maxulysse (approved):

PR #2230, opened 2024-01-24
Filter bar tweaks

ewels:
Went a bit crazier on this one. The switching behaviour bothered me, and it felt like quite a lot of what we had there was a bit unnecessary / took too much space. Ended up condensing it right down and using the same style for desktop and mobile. Say if anyone feels strongly about this.

Before:

CleanShot 2024-01-24 at 20 39 43@2x

After:

CleanShot 2024-01-24 at 20 40 01@2x

---

Before:

CleanShot 2024-01-24 at 20 40 23@2x

After:

CleanShot 2024-01-24 at 20 40 35@2x


mashehu:
Not sure if I like it on the event page.
Before:

After:
Screenshot 2024-01-25 at 09 26 51

Slight Compromise: 
Screenshot 2024-01-25 at 09 28 41

mashehu:
we also lose the counts of elements for the different filters. Overall I like it for the mobile view, but not for larger, too much information and functionality (eg. doube-click to select just one type) being lost

ewels:
Yup, fair enough. Added back at large screen sizes. I agree it works well.

Added an icon for search for better differentiation between elements.

mashehu (approved):

PR #2234, opened 2024-01-24
Bytesize admonition

ewels:
It's bugged me for ages that we have the same Bytesize talk intro in every single bytesize markdown file. Also that you have to manually duplicate the title into a markdown heading.

Updated the website to detect a title that starts with `bytesize` and then show the standard text as an admonition above the page content. Manually removed all of the duplicated content. There was a lot 😆 

Before:

CleanShot 2024-01-24 at 22 42 29@2x

After:

CleanShot 2024-01-24 at 22 42 51@2x


ewels:
NB: Many event pages just lost their ToC sidebar as they no longer need headings. May be worth looking into making full width when no headings? Not sure how difficult that'd be. Separate PR anyway.

mashehu:
> NB: Many event pages just lost their ToC sidebar as they no longer need headings. May be worth looking into making full width when no headings? Not sure how difficult that'd be. Separate PR anyway.
Screenshot 2024-01-26 at 07 48 39

This makes the text too wide. Alternative general approach for event pages: similar to the blog have a left sidebar with other events

Screenshot 2024-01-26 at 07 50 46

ewels:
> have a left sidebar with other events

Like it!

FranBonath (approved):

PR #2236, opened 2024-01-25
Add Vale and GitHub Action

edmundmiller:
I was talking to Drew about the Seqera docs in Boston and he mentioned they were using Vale. I had played with it before. 

Wondering how bad it's going to be with our docs.


mashehu (approved):

mashehu (approved):

edmundmiller:
@mashehu Should be good for a first attempt, and we'll see what goes wrong 😬

PR #2314, opened 2024-02-21
Bytesize: Special interest groups - add YouTube

ewels:
@maxulysse - could you please confirm that this is the correct URL? (You said you trimmed it, is it the same YouTube ID?).


maxulysse:
> @maxulysse - could you please confirm that this is the correct URL? (You said you trimmed it, is it the same YouTube ID?).

I did the trimming in Youtube

maxulysse (approved):

PR #2327, opened 2024-02-23
Add additional information for multi-tool modules

mahesh-panchal:
## Multi-tool module guidelines

- It seems there's already a lot of guidelines present for multi-tool modules.
  - Modules that optimize resource usage should therefore be referred to as multi-tool modules.
  - Fixing the term makes it easier to search for in the page since the information is scattered.
- Clever use of the Github search box should be able to help people find these modules in nf-core repositories.
  - No need to label them.
- Only allow multi-tool modules that have an explicit chain order.
  - Makes naming easier, e.g. `SAMTOOLS/COLLATEFASTQ`
  - Tools used solely for compression don't need to be in the name. e.g. `bwa mem -> samtools sort/view`.
    - How do we name modules that chain over multiple packages?

## Connected discussions

- Slack discussion: https://nfcore.slack.com/archives/C043UU89KKQ/p1702456394353949
- Example of PR that should be kept local: https://github.com/nf-core/modules/pull/4571
  - Reason for exclusion: the Groovy code is potentially too complex to understand for the average user. Pipeline developers should be able to understand quickly what a module is doing. 
- Example of PR that could be included: https://github.com/nf-core/modules/pull/3310

### DIFF
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -611,15 +611,16 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
 
     For guidelines regarding multi-tool modules, please search this page for the phrase `multi-tool`.
 
-    Existing local multi-tool modules can be searched for using the Github search box searching across
-    the nf-core org, for terms such as `args2` `samtools` `collate` `fastq`.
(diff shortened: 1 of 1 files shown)


maxulysse:
> How do we name modules that chain over multiple packages?

I guess it's something that we will see on a case by case basis

maxulysse (approved):

PR #2340, opened 2024-03-01
New shop page

ewels:
Embeds spreadshop, with some CSS customisation

🤫 

@netlify /shop


edmundmiller:
Will this auto-redirect to the US shop? 🤔

ewels:
Not sure, you tell me! https://deploy-preview-2340--nf-core.netlify.app/shop/

For me it's showing GBP, so it might be a bit funky.

I was also thinking about having a FAQ and stuff - so might be worth having our own landing page before going through to the shop pages. Could then have links to the different shop sites, along with some FAQ. What do you think?

edmundmiller:
I'm getting pounds 😞 

Yeah that could work. I bet we could render one or the other based on location.

ewels:
Ok, I googled a bit more and also came to the same conclusion - that the auto redirect wouldn't work.

Just put together an index page and then location-specific embed pages under a dropdown. This should be more reliable. Also the FAQ mean that people hopefully know what they're getting themselves into.

Let me know what you think.

edmundmiller (approved):
That looks lovely now! Love the FAQ before a shop renders. I agree that should be more reliable.

nvnieuwk:
One small thing I've noticed is that by selecting the `Belgium: Dutch Euro` option, it actually displays the page in French (which can be an issue because not all Flemish people speak French fluently, asking for a friend :grin:)

FriederikeHanssen (approved):
lgtm

jfy133 (approved):
I would add a note on the embed page that makes sure to check all pages to see all designs e.g. pipeline specific ones, or mention they exist in the description

maxulysse (approved):

ewels:
Thanks all - I'll add some "highlights" boxes above the button to go to the shop:

* pipeline specific swag
* customise your colours
* big range of products

With subtext to explain how to eg. see the full list of products.

ewels:
Done 🎨

PR #2345, opened 2024-03-04
Try to unconfuse stdout/err capturing

pontus:
I found the point about capturing stdout and stderr hard to parse and confusing and have tried to make it less so.

This also adds an example of suggested behaviour when tools do not use stdout/stderr but capture such outputs to files internally.

### DIFF
--- a/src/content/docs/contributing/modules.md
+++ b/src/content/docs/contributing/modules.md
@@ -773,10 +773,14 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     }
     ```
 
-10. In some cases, STDOUT and STDERR need to be saved to file, for example for reporting purposes. Use the shell command `tee` to redirect the
-    streams to both file and it's original stream. This allows for the streams to be captured by the job scheduler's stream logging capabilities
-    and print them to screen when Nextflow encounters an error. In particular, when using `process.scratch`, the log files may not be preserved when
-    the job scheduler relinquishes the job allocation.
+10. In some cases, STDOUT and STDERR may need to be saved to file, for example for reporting purposes. Use
+    the shell command `tee` to capture the streams but preserve the streams. This allows for the streams to
+    be captured by the job scheduler's stream logging capabilities and print them to screen when Nextflow
+    encounters an error.
+
+    This also makes certain they are captured by Nextflow whereas things written to files only may be lost when
+    the job scheduler relinquishes the job allocation lost if using `process.scratch` - unless those files are
+    designated outputs from the process.
 
     ```nextflow {7-8}
     script:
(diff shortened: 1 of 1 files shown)


src/content/docs/contributing/modules.md
```diff
@@ -773,10 +773,14 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
-    and print them to screen when Nextflow encounters an error. In particular, when using `process.scratch`, the log files may not be preserved when
-    the job scheduler relinquishes the job allocation.
+10. In some cases, STDOUT and STDERR may need to be saved to file, for example for reporting purposes. Use
+    the shell command `tee` to capture the streams but preserve the streams. This allows for the streams to
+    be captured by the job scheduler's stream logging capabilities and print them to screen when Nextflow
+    encounters an error.
+
+    This also makes certain they are captured by Nextflow whereas things written to files only may be lost when
+    the job scheduler relinquishes the job allocation lost if using `process.scratch` - unless those files are
+    designated outputs from the process.
```
mashehu, line 783:
```suggestion
    This ensures that they are captured by Nextflow. If information is only written to files, it could potentially be lost when the job scheduler gives up the job allocation. 
    This is particularly true if `process.scratch` is used, unless those files are specifically marked as outputs from the process.
```

src/content/docs/contributing/modules.md
```diff
@@ -790,6 +794,19 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     """
     ```
 
+    Similarly, if the tool captures STDOUT or STDERR to a file itself, it is best to send those to the
+    corresponding streams as well. Since a timeout may mean execution is aborted, it may make most sense
+    to have background tasks do that.
+
+    ```nextflow {7-8}
```
mashehu, line 801:
```suggestion
    ```nextflow {3-4}
```

mashehu (approved):

src/content/docs/contributing/modules.md
mahesh-panchal: The outputs are lost even if they're designated outputs and `process.scratch` is used. Nextflow copies the designated output files only when the script successfully completes

src/content/docs/contributing/modules.md
```diff
@@ -790,6 +794,19 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
...
+    script:
+    """
+    tail -F stored_stderr.log >&2 &
+    tail -F stored_stdout.log &
```
mahesh-panchal, line 805: I only knew about `-f`. Today I learned about `-F`. Nice.

src/content/docs/contributing/modules.md
```diff
@@ -773,10 +773,14 @@ The key words "MUST", "MUST NOT", "SHOULD", etc. are to be interpreted as descri
     }
     ```
 
-10. In some cases, STDOUT and STDERR need to be saved to file, for example for reporting purposes. Use the shell command `tee` to redirect the
-    streams to both file and it's original stream. This allows for the streams to be captured by the job scheduler's stream logging capabilities
-    and print them to screen when Nextflow encounters an error. In particular, when using `process.scratch`, the log files may not be preserved when
-    the job scheduler relinquishes the job allocation.
+10. In some cases, STDOUT and STDERR may need to be saved to file, for example for reporting purposes. Use
+    the shell command `tee` to capture the streams but preserve the streams. This allows for the streams to
```
mahesh-panchal, line 777:
```suggestion
    the shell command `tee` to simultaneously capture and preserve the streams. This allows for the streams to
```

src/content/docs/contributing/modules.md
mashehu: updated the suggestion accordingely

mashehu:
@nf-core-bot fix linting

PR #2346, opened 2024-03-04
Split page int subpages and upgrade to astro 4

mashehu:


mashehu:
Closing in favor for #2486

PR #2362, opened 2024-03-14
Hackathon March 2024: Add Seqera as a local site

ewels:


ewels:
@nf-core-bot fix linting

PR #2363, opened 2024-03-14
Map of hackathon sites

ewels:
Quick and dirty implementation of a map showing the locations of the hackathon sites. Did this to get a screenshot for the intro slides, but it'd be nice to have it in the page properly!

Needs some magic from @mashehu to dynamically grab the coordinates from the MDX files somehow, instead of having to hardcode in a YAML file like I did here.

Note that I think I got an off-by-one error when making the YAML 🙈


mashehu (approved):
Too late for this one, but will be handy for future events.

jfy133:
Browser looks, ok, but map covers first couple of lines of table below it

PR #2376, opened 2024-03-18
Improve events' location meta fields

### LINKED ISSUE #673: New event page meta fields
It would be nice to rethink the event YAML fields. At the moment we're sort of hijacking the `location` field as it was a quick and dirty fix when we moved online, but would be nice to have dedicated fields for Slides / DOI, YouTube, Bili Bili, Zoom and so on.

Each should probably have the option of a string or a list (eg. when there are multiple talks at an event) and each should be able to just take a url string or a dict with more options (text + URL).

Could also be nice to add custom Font Awesome icons for Bili Bili and DOI etc.

joaquimgamero:
Closes #673

- Moves `location...` related fields to an objects inside an array of `locations` in the schema
- Links are now its own array (with custom icons like github, bilibili, youtube)
- Locations themselves can also have links (useful for links of the event venue, for example)

💡 Markdown files still not updated, looking for feedback first


src/components/event/EventHeader.astro
```diff
@@ -36,6 +36,7 @@ const end = frontmatter.end;
                                         return (
                                             <li class="list-item">
                                                 <a class="text-white underline" href={loc}>
+                                                    // Add logic to display icon depending on URL
```
mashehu, line 39:
```suggestion
                                                    // TODO: Add logic to display icon depending on URL
```

joaquimgamero:
Wondering how will this impact the UI if an event will have multiple locations, how should we adapt this section specifically, for example?

Screenshot 2024-03-18 at 12 02 29

src/content/config.ts
```diff
@@ -26,9 +26,19 @@ const events = defineCollection({
...
-            locationName: z.string().optional(),
-            locationURL: z.string().url().or(z.string().startsWith('#')).or(z.array(z.string().url())).optional(),
-            locationLatLng: z.array(z.number(), z.number()).optional(),
+            locations: z.array(
+                z.object({
+                    locationName: z.string().optional(),
+                    locationURL: z.string().url().or(z.string().startsWith('#')).or(z.array(z.string().url())).optional(),
+                    locationLatLng: z.array(z.number(), z.number()).optional(),
+                    // Fields more relevant for online events
+                    slides: z.array(z.string()).optional(),
```
mashehu, line 35:
```suggestion
                    slides: z.array(z.string().url()).optional(),
```
also for the rest, so we have automatic URL validation

src/content/config.ts
```diff
@@ -26,9 +26,19 @@ const events = defineCollection({
...
+                    doi: z.array(z.string()).optional(),
+                    youtube: z.string().optional(),
+                    biliBili: z.string().optional(),
+                    zoom: z.string().optional(),
```
joaquimgamero, line 39: Maybe I misunderstood and these should be added just as items in the `locationURL` array?

mashehu: hmm, I think we use locationURL quite often for venue URLs, so might be fine to keep them seperated

mashehu: yeah, we had something like that for the old website. it's a bit of a pain to cover all different urls, but it will be mostly youtube, so maybe it's fine? 🤔

joaquimgamero: In the end I went a bit further, a small summary:
- Locations is an array now
- Locations can have their own links (useful for the event's venue link)
- There is a new separate `links` array which represent links for the event in general, here some custom icons are added for youtube, github, bilibili, etc...

Example:

Screenshot 2024-03-18 at 17 06 38

mashehu: very nice!
need just to play with the layout a bit to make the columns more balanced (and maybe center the event export button).
because we have the icons, I don't think we need the bullets for the links list.
Thinking further: maybe we just have the icons and don't show the links at all (because they are usually quite ugly)(but make them more discoverable, so that's something we need to try out/discuss)

src/components/event/EventHeader.astro
```diff
@@ -26,17 +45,71 @@ const end = frontmatter.end;
         </div>
         <div class="col">
             <dl>
+                <dt>Locations:</dt>
+                <dd>
+                    <ul class="list-unstyled">
+                        {
+                            locations.map((location) => {
```
mashehu, line 52:
```suggestion
                            locations?.map((location) => {
```

src/components/event/EventHeader.astro
```diff
@@ -26,17 +45,71 @@ const end = frontmatter.end;
...
+                        }
+                    </ul>
+                </dd>
+
                 {
-                    frontmatter.locationURL && (
+                    frontmatter.links && frontmatter.links.length > 0 && (
                         <>
-                            <dt>{frontmatter.locationURL.length === 1 ? 'Link' : 'Links'}:</dt>
+                            <dt>Links:</dt>
```
mashehu, line 103: could we have "Link" if `frontmatter.links.length==1`

src/components/event/EventHeader.astro
```diff
@@ -26,17 +45,71 @@ const end = frontmatter.end;
...
                             <dd>
-                                <ul class="list-unstyled">
-                                    {frontmatter.locationURL.map((loc) => {
+                                <ul>
+                                    {frontmatter.links.map((link) => {
```
mashehu, line 106:
```suggestion
                                    {frontmatter.links?.map((link) => {
```

src/components/event/EventHeader.astro
```diff
@@ -12,6 +12,25 @@ if (frontmatter.end === undefined) {
...
+        return 'fa-zoom-us';
+    } else if (/^(https?:\/\/)?(www\.)?bilibili\.com\/.+$/gi.test(link)) {
+        return 'fa-bilibili';
+    } else if (/^(https?:\/\/)?(www\.)?github\.com\/.+$/gi.test(link)) {
+        return 'fa-github';
+    } else if (/^(https?:\/\/)?(www\.)?linkedin\.com\/.+$/gi.test(link)) {
+        return 'fa-linkedin';
+    } else if (/^(https?:\/\/)?(www\.)?twitter\.com\/.+$/gi.test(link)) {
+        return 'fa-twitter';
+    } else if (/^(https?:\/\/)?(www\.)?(docs\.google\.com)\/.+$/gi.test(link)) {
```
mashehu, line 30: slideshare would also be a nice case to handle

src/content/config.ts
joaquimgamero: I actually was thinking that the export button should be placed next to the "Edit" button, or in the same corner to say the least. Moving link icons below the event dates and leaving the right column for the locations might balance everything a bit better, I will try tomorrow!

mashehu (approved):
Great work!

src/content/config.ts
```diff
@@ -26,9 +26,21 @@ const events = defineCollection({
...
                 })
                 .optional(),
-            locationName: z.string().optional(),
-            locationURL: z.string().url().or(z.string().startsWith('#')).or(z.array(z.string().url())).optional(),
-            locationLatLng: z.array(z.number(), z.number()).optional(),
+            locations: z
+                .array(
+                    z.object({
+                        locationName: z.string().optional(),
+                        locationURL: z
```
ewels, line 33: What do folks think about renaming these to just `name`, `url` etc now that they're nested? The `location` seems a bit duplicative and unnecessary..

src/components/event/EventHeader.astro
joaquimgamero: Makes sense, thanks! https://github.com/nf-core/website/pull/2376/commits/7c7120d341a289eec67e1eb12e277b59f9ff10cf

src/components/event/EventHeader.astro
joaquimgamero: Yeap! I got rid of it after tinkering around: https://github.com/nf-core/website/pull/2376/commits/7c7120d341a289eec67e1eb12e277b59f9ff10cf

src/components/event/EventHeader.astro
joaquimgamero: Sure, added it here: https://github.com/nf-core/website/pull/2376/commits/7c7120d341a289eec67e1eb12e277b59f9ff10cf

joaquimgamero:
Made some layout tweaks to balance everything a bit more. I've also fixed the Export event dropdown being shown "under" the down facing triangle.

Screenshot 2024-03-19 at 10 47 36

src/components/event/EventHeader.astro
```diff
@@ -23,57 +46,75 @@ const end = frontmatter.end;
...
+                                                        <a
+                                                            class="m-1 btn btn-sm btn-outline-light"
+                                                            href={`https://www.openstreetmap.org/?mlat=${location.geoCoordinates[0]}&mlon=${location.geoCoordinates[1]}&zoom=16#map=16/${location.geoCoordinates[0]}/${location.geoCoordinates[1]}`}
+                                                            target="_blank"
+                                                        >
+                                                            <i class="fa-duotone fa-map-pin me-1" />
+                                                            See on map
+                                                        </a>
+                                                    )}
+                                                    :
```
mashehu, line 76:
```suggestion
                                                    :<a
                                                            class="m-1 ms-auto btn btn-sm btn-outline-light"
                                                            href={`https://www.openstreetmap.org/?mlat=${location.geoCoordinates[0]}&mlon=${location.geoCoordinates[1]}&zoom=16#map=16/${location.geoCoordinates[0]}/${location.geoCoordinates[1]}`}
                                                            target="_blank"
                                                        >
                                                            <i class="fa-duotone fa-map-pin me-1" />
                                                            See on map
                                                        </a>
                                                    )}

```

joaquimgamero:
@nf-core-bot fix linting

PR #2396, opened 2024-03-20
restructure tools training

FranBonath:
I split the training into several .md files. This helps the user of the training to navigate and the creators of the training to avoid merge conflicts.


ewels (approved):
LGTM!

PR #2343, opened 2024-03-03
Blog post: special interest groups

ewels:
@netlify /blog/2024/special_interest_groups

WIP, as written on a plane with no internet connection.

Ended up being a bit long, worth discussing whether we want to move some of this into some new documentation and slim down the blog post.

To do:

- [x] Link to bytesize talk in introduction
- [x] Potentially split out into docs + blog post
- [ ] Add photos / more artwork

@netlify /blog

PR #2419, opened 2024-03-28
Restructure docs contributing page

jfy133:
This restructures somewhat the contributing sections of docs.

I'm aware of a larger plan to completely overhaul docs but my frustration boiled over to a rage quit my current TODO  (😬) to at least 'patch fix' the following:

My main aim was to isolate the 'specifications' sections of the various component guidelines which are becoming a spaghetti mess/knowledge dumping ground.

I use the specifications sections to review various PRs, but they are getting 1. increasingly harder to find buried in the middle of the page, 2. other people are contributing new specifications (e.g. nf-test) outside that section.

Thus I've made a new subsection called specifications, and extracted these parts of the existing 'guideliens' page (and improved the test data specifications a bit).

I've also renamed the page titles  of the original  guidelines pages to group them together accordingly, and distinguish these from the specification pages.

Things to fix/consider:

- The new specifications left-hand menu seems to be empty
- Put the newly named 'guidelines' in their own separate subsection, however this will require page re-directions (particularly `modules.md` is linked a lot on various repo README)

But I would leave this to @mashehu and @ewels to consider..

@netlify /docs/contributing/component-specifications/modules


src/content/docs/contributing/guidelines/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Guidelines
+title: Guidelines for Pipeline
```
mashehu, line 2:
```suggestion
title: Guidelines for Pipelines
```

ewels:
Looks good! The side navbar doesn't work at the deepest level of nesting, and we _really_ need breadcrumbs now that we're getting this level of organisation as it's very difficult to climb back out. But the organisation of the content is definitely better 👍🏻 

My vote would still be to avoid using numbered lists and instead use (numbered?) headings, which can be linked to directly. But that's not a blocker for this PR.

jfy133:
> Looks good! The side navbar doesn't work at the deepest level of nesting, and we _really_ need breadcrumbs now that we're getting this level of organisation as it's very difficult to climb back out. But the organisation of the content is definitely better 👍🏻
> 
> My vote would still be to avoid using numbered lists and instead use (numbered?) headings, which can be linked to directly. But that's not a blocker for this PR.

I'm happy to make more linkable references for the specifciations!

You mean instead to have something lke

```
### 1 this MUST be like this
```

mashehu:
>  The side navbar doesn't work at the deepest level of nesting,

fixed in #2343

ewels:
@mashehu - different thing. The ToC nesting was fixed in that PR, but I'm talking about the left-sidebar.

This should probably have links to the other sibling pages, with the parent page(s) at the top or something:

CleanShot 2024-03-30 at 23 00 34@2x

ewels:
@jfy133 exactly - each point is already numbered so it should be a fairly easy change.

Instead of this:

```markdown
### General

1.  All mandatory and optional input files MUST be included in `input` channel definitions.

2.  Non-file mandatory arguments or arguments needed to modify the command to make the module run with no error, SHOULD be provided as value channels (for example `lib_type` in [salmon/quant](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf)) - see 'Input/output options' below.

3.  All _non-mandatory_ command-line tool _non-file_ arguments MUST be provided as a string via the `$task.ext.args` variable.

    - The value of `task.ext.args` is supplied from the `modules.config` file by assigning a closure that returns a string value to `ext.args`. The closure is necessary to update parameters supplied in a config with `-c`.
```

Something like this:

```markdown
### 1 General

#### 1.1  Mandatory input files in input channels

All mandatory and optional input files MUST be included in `input` channel definitions.

#### 1.2 Optional arguments in value channels

Non-file mandatory arguments or arguments needed to modify the command to make the module run with no error, SHOULD be provided as value channels (for example `lib_type` in [salmon/quant](https://github.com/nf-core/modules/blob/master/modules/nf-core/salmon/quant/main.nf)) - see 'Input/output options' below.

#### 1.3 Arguments in ext.args

All _non-mandatory_ command-line tool _non-file_ arguments MUST be provided as a string via the `$task.ext.args` variable.

The value of `task.ext.args` is supplied from the `modules.config` file by assigning a closure that returns a string value to `ext.args`. The closure is necessary to update parameters supplied in a config with `-c`.
```

This has a few benefits:

* It's hopefully easier to read - less indentation and better sections in text
* Each item has a heading and so an anchor link that can be linked to directly
* Each item will show up in the ToC sidebar

The downside is that the numbering can get messed up if we add something new in.

One optional workaround for this is that we don't add numbering in the markdown itself, but rather as a website feature that can be turned on via the page frontmatter. Then we can add in the numbers automatically in the rendering (no need to check that they are sequential manually) and also not include them in the header links, so that these don't change if something gets inserted. This is what we do in training.nextflow.io

For example: the `3.2 Create a transcriptome index file` but note that there is no `3.2` section of the docs renders on the page with numbers, but there is no number in the link or [markdown](https://github.com/nextflow-io/training/blob/f567173428311a11d84b56f6a602bde3497f5f32/docs/basic_training/rnaseq_pipeline.md?plain=1#L112):

https://training.nextflow.io/basic_training/rnaseq_pipeline/#create-a-transcriptome-index-file

..but, that needs _someone_ to add support for this new feature on the website 😉

JoseEspinosa (commented):
Don't have anything to add to @ewels and @mashehu comments but just want to say that this re-ordering is a very nice idea @jfy133 ❤️

FranBonath:
To me it seems that the link from Docs in the main menue is still pointing to the old page. Is that intentional? @jfy133

FranBonath:
Also I get broken links if I click on "Usage" and "Contributing". Is that a me issue or something else?

mahesh-panchal:
> Don't have anything to add to @ewels and @mashehu comments but just want to say that this re-ordering is a very nice idea @jfy133 ❤️

☝🏽 This.

jfy133:
@ewels now numbered!

@mashehu for @FranBonath 's second points - I think that will have to be something for you. For example the URL ending in `docs/contributing/nfcore_component_specifications/`, has on the left hand menu a dedicated clickable section to `docs/contributing`, but this is not a sub page but rather a column on the main `docs` index page...

@FranBonath which 'docs in the main menu' are you referring to?

jfy133:
> problem with the sidebar was inconsistent naming 😉 (also added a page for subsections, so you can click all titles now)

Oops my bad, didn't realise :sweat_smile:

maxulysse:
The deploy link button is not working

maxulysse (approved):
Looking good, I love it

ewels (approved):
Numbering is much nicer, thanks @jfy133 👍🏻 

Looking good! General wish list, all things that can be tackled in other PRs:

1. Frontmatter to add numbers to headers (not anchors), remove from markdown
2. Breadcrumbs for easier navigation
3. (Maybe) Left sidebar that shows current level of nav hierarchy - currently there's no way to get between "sibling" pages, and the list of _all_ docs is very long.

ewels:
I also am finding the docs extremely confusing - for example "Guidelines for pipelines" is probably more specifications (as I think you also think, given that the specifications index page links there). Then we have "Guidelines for Modules" which is different to "Specifications for modules". Blaaah 😵

jfy133:
> I also am finding the docs extremely confusing - for example "Guidelines for pipelines" is probably more specifications (as I think you also think, given that the specifications index page links there). Then we have "Guidelines for Modules" which is different to "Specifications for modules". Blaaah 😵

I agree, I don't mind re-naming the `Guidelines for pieplines` if you want, but that is a big set of files and I don't want to rock the boat too much... 😅

It's also why I didn't move all the 'guidelines' pages into it's own sub section (I'm worried about breaking URLs etc...)

FranBonath:
> @ewels now numbered!
> 
> @mashehu for @FranBonath 's second points - I think that will have to be something for you. For example the URL ending in `docs/contributing/nfcore_component_specifications/`, has on the left hand menu a dedicated clickable section to `docs/contributing`, but this is not a sub page but rather a column on the main `docs` index page...
> 
> @FranBonath which 'docs in the main menu' are you referring to?

Screenshot 2024-04-04 at 10 16 53

When I press the `Docs` button there, I get the old page.

PR #2421, opened 2024-03-30
Update publish date for SIGs blog post

ewels:
@netlify /blog/2024/special_interest_groups


ewels:
Meh, not sure it's worth modifying - it disappears from the blog page completely so would need to be careful with merge times for it not to randomly dissappear. And not sure that it really matters that the publish date was a little stale.

mashehu (approved):

PR #2436, opened 2024-04-04
Push the announcement date for Boston hackathon

ewels:
I feel like this is too far off, if this banner is _always_ there then people will start to ignore it 😆

@netlify /events/2024/hackathon-boston


mashehu (approved):
another option would be to use the text field in `announcement` to have the small banner

ewels:
True! Bring back the `<marquee>`!!

PR #2446, opened 2024-04-10
Reformat blog dates from MM/DD/YYYY to DD/MMM/YEAR

Joon-Klaps:
Reformat blog dates from MM/DD/YYYY to DD/MMM/YEAR

### DIFF
--- a/src/pages/blog/index.astro
+++ b/src/pages/blog/index.astro
@@ -127,8 +127,14 @@ const sections = years.map((year) => {
                                                 />
                                             ))}
                                         </div>
-                                        <span class="text-body-secondary text-small text-end flex-grow-1">
-                                            published at {new Date(post.data.pubDate).toLocaleDateString()}
+                                        <span class="text-body-secondary text-small text-end flex-grow-1 text-nowrap">
+                                            published at
+                                            <br \>
+                                                {new Date(post.data.pubDate).toLocaleDateString('en-GB', {
+                                                    day: 'numeric',
+                                                    month: 'short',
+                                                    year: 'numeric',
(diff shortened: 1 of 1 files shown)


Joon-Klaps:
> not so sure about this, don't want to make it confusing for US people... (with `toLocaleDateString()` US people should get MM/DD/YYYY, while Europeans should get DD/MM/YYYY (depending on their computer/browser locale settings) )

Even when it's written with the month abbreviation - '1 Feb 2024'?  Or you mean to drop the 'en-GB' in `toLocalDateString()`?

src/pages/blog/index.astro
```diff
@@ -127,8 +127,14 @@ const sections = years.map((year) => {
...
                                         </div>
-                                        <span class="text-body-secondary text-small text-end flex-grow-1">
-                                            published at {new Date(post.data.pubDate).toLocaleDateString()}
+                                        <span class="text-body-secondary text-small text-end flex-grow-1 text-nowrap">
+                                            published at
+                                            <br />
+                                            {new Date(post.data.pubDate).toLocaleDateString('en-GB', {
+                                                day: 'numeric',
+                                                month: 'short',
+                                                year: 'numeric',
```
mashehu, line 136:
```suggestion
                                            {new Date(post.data.pubDate).toLocaleDateString('en-GB', {
                                                year: 'numeric',
                                                month: 'short',
                                                day: 'numeric',
```
just so we follow ISO order 🙂

mashehu (approved):
sorry, checked now the rest of the code and we have exactly the same there. just some reordering needed

PR #2467, opened 2024-04-18
Upload animal genomics first session youtube videos

JoseEspinosa:
@netlify /special-interest-groups/animal-genomics


JoseEspinosa:
@nf-core-bot fix linting

mashehu (commented):
there is quite some logic connected to` event_types`. If possible we should stick to a low number for them, maybe this could be a `talk` instead?

JoseEspinosa:
> there is quite some logic connected to` event_types`. If possible we should stick to a low number for them, maybe this could be a `talk` instead?

Fair enough, we had [this](https://nfcore.slack.com/archives/CE7DN1U7M/p1713476978707249) thread this morning discussing about it. I am also ok if we leave it as `talk`. Wanted to try myself but if there is too much logic I will just remove it from this PR, in any case.

JoseEspinosa:
@nf-core-bot fix linting

ewels (changes requested):
I don't think that we need a second event page - it was only one event, even if we chopped the video into two pieces. If we can just have both videos embedded on the one page that'd be ideal (if you agree). Should be possible.

JoseEspinosa:
> I don't think that we need a second event page - it was only one event, even if we chopped the video into two pieces. If we can just have both videos embedded on the one page that'd be ideal (if you agree). Should be possible.

I added two pages because I wanted to list them with different labels but if both are listed as `talk` then I don't think there is a need to do it.

JoseEspinosa:
@nf-core-bot fix linting

ewels:
I think that there is a bug in the website about embedding YouTube videos on event pages currently. The fix for that can come in a different PR, but it'd be nice if @mashehu can confirm that (once fixed) it'll support multiple Youtube videos..?

Otherwise LGTM 👍🏻

mashehu:
#2471 fixes the broken embeds and can handle multiple embeds

mashehu (approved):
Ready to be merged!

PR #2492, opened 2024-05-06
Add new core members to governance page

nvnieuwk:
The title says it all :grin: 

@netlify /about/governance


JoseEspinosa (commented):
Maybe take the opportunity to also remove @ggabernet from core and add her to steering and alumni

ewels (approved):

PR #2441, opened 2024-04-05
Docs restructure

ewels:
Results of _[edit: "much more than"]_ a day's hacking by @mashehu and @ewels 

Still a lot left to do, see project board: https://github.com/orgs/nf-core/projects/67/views/1

Deployment preview: https://deploy-preview-2441--nf-core.netlify.app/docs

@netlify /docs

PR #2525, opened 2024-05-18
Fix link to tools docs

mirpedrol:
https://github.com/nf-core/website/pull/2441#issuecomment-2117472345

@netlify /docs/usage/Getting_started/introduction


ewels (approved):
Ah great, thanks! I think that should redirect so be ok, but better to use the proper links..

PR #2555, opened 2024-05-28
Added meta-omics overview svg

erikrikarddaniel:


jfy133 (approved):
Ugh baby said no, so I suggest merging in and I finish another time (of course i came a cross a bunch of issues - not related to you) which got in the way. 

Only question if the file location is a good idea... @mashehu @ewels is the location ok or should put in in an assets folder somewhere?

PR #2568, opened 2024-06-01
exclude npm packages from renovate updates

mashehu:


edmundmiller:
Missing the context here, did this break something?

mashehu:
Nope, I just prefer to do it manually. Maybe grouping them into one PR makes them easier to still keep around as a reminder.

edmundmiller (approved):
Thanks for the explanation! 

Are they automerging? Could we set it to request you on a review, and then you can manually pick the things you want to update? Any way to codify the bumps that you're handpicking?

Just looking for feedback before we start actually using it in nf-core modules.

PR #2569, opened 2024-06-02
Guidelines: Workflow name

ewels:
Make the workflow name guidelines a bit more verbose and detailed.

@netlify /docs/guidelines/pipelines/requirements/workflow_name

### DIFF
--- a/src/content/docs/guidelines/pipelines/requirements/workflow_name.md
+++ b/src/content/docs/guidelines/pipelines/requirements/workflow_name.md
@@ -7,6 +7,12 @@ menu:
 ---
 
 All nf-core pipelines should be lower case and without punctuation.
-This is to maximise compatibility with other platforms such as Docker Hub, which enforce such rules.
-We prefer that they are descriptive towards the data or analysis type the pipeline will be using or performing, and should be approved by the nf-core community [(`#new-pipelines`)](https://nfcore.slack.com/archives/CE6SDEDAA).
-In documentation, please refer to your pipeline as `nf-core/yourpipeline`.
+This is to maximise compatibility with other platforms.
+
+We prefer that names are self-descriptive towards the data or analysis type the pipeline will be using or performing.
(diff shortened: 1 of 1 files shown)


jfy133 (approved):
It works but I wonder if a bullet point list would make to easier to identify the criteria?

- lower case a-z 
- no special characters 
- descriptive and intuitive (to make it easily for users to know what it is for)


Or something like that, then people can tick each off in their head when Reading

src/content/docs/guidelines/pipelines/requirements/workflow_name.md
```diff
@@ -7,6 +7,12 @@ menu:
 ---
 
 All nf-core pipelines should be lower case and without punctuation.
-This is to maximise compatibility with other platforms such as Docker Hub, which enforce such rules.
-We prefer that they are descriptive towards the data or analysis type the pipeline will be using or performing, and should be approved by the nf-core community [(`#new-pipelines`)](https://nfcore.slack.com/archives/CE6SDEDAA).
-In documentation, please refer to your pipeline as `nf-core/yourpipeline`.
+This is to maximise compatibility with other platforms.
+
+We prefer that names are self-descriptive towards the data or analysis type the pipeline will be using or performing.
```
jfy133, line 12:
```suggestion
We strongly recommend that names are self-descriptive towards the data or analysis type the pipeline will be using or performing.
```

ewels:
@jfy133 - updated. Let me know what you think 👍🏻

jfy133 (approved):

jfy133:
Ah one last thing: maybe add a tip block saying we know that they're are a couple that don't have that name, but they are there for legacy reasons? To fend off the inevitably 'but sarek and eager'

ewels:
Good idea - added 👍🏻

christopher-hakkaart (approved):

christopher-hakkaart:
I think this is good to go as well

PR #2573, opened 2024-06-05
Fix Maxime's file icons so when SVG opened/viewed text is visible

jfy133:


mashehu (approved):
The icon text e.g. Bam is not centered on with respect to the black ribbon. And for my taste there should be only one ribbon also for multiple files (but then centered accordingly). I also don't like that one sees the q in the second fastq file in the back.

PR #2577, opened 2024-06-12
Update Neuroimaging sig

AlexVCaron:
Add new member @GuillaumeTh 

@netlify /special-interest-groups/neuroimaging


AlexVCaron:
@mashehu I requested a review from you, but was also wondering : is it ok to handle reviews internally in the SIG ? For minor modifications affecting only the neuroimaging SIG pages that is. Just want to get a heads up.

mashehu (approved):

mashehu:
yes, no need to wait for my 👍🏻 on this kind of small changes, but thanks for asking

PR #2608, opened 2024-07-02
Add comment about setting a default prefix that isn't just meta.id

SPPearce:
@netlify /docs/guidelines/components/modules


mashehu (approved):

mashehu:
failing playwright tests can be ignored for now

sites/docs/src/content/docs/guidelines/components/modules.md
```diff
@@ -390,13 +390,19 @@ Channel names MUST follow `snake_case` convention and be all lower case.
...
-- As a result of using this syntax, if the module has the same named inputs and outputs then you can add a line in the `script` section like below (another example [here](https://github.com/nf-core/modules/blob/e20e57f90b6787ac9a010a980cf6ea98bd990046/modules/lima/main.nf#L37)) which will raise an error asking the developer to change the `args.prefix` variable to rename the output files so they don't clash.
+- As a result of using this syntax, if the module could _potentially_ have the same named inputs and outputs add a line in the `script` section like below (another example [here](https://github.com/nf-core/modules/blob/e20e57f90b6787ac9a010a980cf6ea98bd990046/modules/lima/main.nf#L37)) which will raise an error asking the developer to change the `args.prefix` variable to rename the output files so they don't clash.
 
   ```nextflow
   script:
   if ("$bam" == "${prefix}.bam") error "Input and output names are the same, set prefix in module configuration to disambiguate!"
   ```
 
+- If the input and output files are likely to have the same name, then an appropriate default prefix may be set, for example:
```
jfy133, line 400:
```suggestion
- If the input and output files are likely to have the same name, then an appropriate default prefix MAY be set, for example:
```

I  feel this should be left for the developer to decide how the resulting file should be called. The error should make them aware of this.

mahesh-panchal: Setting the default prefix does give the control to the developer, as they can overwrite it in the modules.config. The problem is when it's hard-coded into the output path.
I also added the `-C` bash flag to the shell directive in the template so that should also prevent accidental clobbering.

mahesh-panchal (approved):

sites/docs/src/content/docs/guidelines/components/modules.md
jfy133: > The problem is when it's hard-coded into the output path.

I don't think I follow... isn't the suggestion here to technically embed a hardcoded string?

mahesh-panchal: No, the example clearly shows prefix having a default of `"${meta.id}_sorted"`. If I wanted something different, I just update in the config
```
    ext.prefix = { "${meta.id}_mysorted" } 
```
The command should still look like:
```
mycommand --input $file > ${prefix}.out
```

SPPearce: Yes, it is setting a default, but can be entirely overwritten in the usual way

jfy133: Ah I see. I'm still not particularly happy (I don't see why a pipeline developer can't do this themselves), but I won't die on this hill (I feel strongly about, but not enough to fight everyone).

SPPearce: > Ah I see. I'm still not particularly happy (I don't see why a pipeline developer can't do this themselves), but I won't die on this hill (I feel strongly about, but not enough to fight everyone).

Yes, it can be done, but it shouldn't have to be. 
It also means you don't have to add a config for nf-test to work, to simplify there

jfy133: Disagree on both points (1. it should, pipeline devs should be very aware of the output files, 2. I don't think we should sacrifice flexibility just to avoid a small config file) but like I said - 'small mound' :laughing: Maybe ask for one more opinion

SPPearce: > Disagree on both points (1. it should, pipeline devs should be very aware of the output files, 2. I don't think we should sacrifice flexibility just to avoid a small config file).
> 
> Bbut like I said - 'small mound' 😆 Maybe ask for one more opinion and you can merge

I mean, we could say that pipeline devs should make each module from scratch because that would be purer. I don't understand your objection here. This sacrifices no flexibility whatsoever, it defines a default prefix that can then be overwritten in exactly the usual way.

mahesh-panchal: I also don't understand the objections.

nvnieuwk: I follow James here. IMO it's better to have a common way of doing things in modules so you don't have to check how it's done in every module. If I know every module has the `meta.id` as default, then I don't have to check the default every time. Adding different defaults could become confusing for some developers. But also not a hill I'm willing to die on :)

mahesh-panchal: My experience with using nf-core modules, when written by someone else is usually that I'll have to check the code anyway to know what the output channels are named at the least, and what inputs I'm expecting. If I then see the 
```
if ( file.name == "${prefix}.ext" ){
```
I'll know there's a possibility for a filename collision. However, this is where I would expect the module to have a sensible default not to have filename collision if I just plug and play. More often than not though, the current state is that once I use it, I'll discover the default is not actually sensible and have to modify my modules.config to deal with it. This to me is a time waster. If we did use "sensible defaults", rather than "meta.id", then I'll see when I inspect the module to change `prefix` if I wanted it differently, but I'd rather not assume that `${prefix}.ext` will result in a filename collision if I don't set my own prefix anyway. That's at least my experience, which is why I'm for this update.

nvnieuwk: Maybe we should move this discussion to the next maintainers meeting? This seems like a topic that could fit very well in the meetings

mashehu:
What is the state of this docs update, @SPPearce ?

SPPearce:
> What is the state of this docs update, @SPPearce ?

@jfy133  and @nvnieuwk are still against it as far as I know, but I still don't understand why. 
The suggestion was to discuss at a maintainers meeting but I was on leave for the one last week.

mashehu:
> > What is the state of this docs update, @SPPearce ?
> 
> @jfy133 and @nvnieuwk are still against it as far as I know, but I still don't understand why. The suggestion was to discuss at a maintainers meeting but I was on leave for the one last week.

ah, yes, remember now. let's try to bring it up in the next meeting then.

nvnieuwk:
Sorry for being difficult :grin: let's talk about it in the next meeting indeed :)

sites/docs/src/content/docs/guidelines/components/modules.md
```diff
@@ -390,13 +390,19 @@ Channel names MUST follow `snake_case` convention and be all lower case.
 Output file (and/or directory) names SHOULD just consist of only `${prefix}` and the file-format suffix (e.g. `${prefix}.fq.gz` or `${prefix}.bam`).
 
 - This is primarily for re-usability so that other developers have complete flexibility to name their output files however they wish when using the same module.
-- As a result of using this syntax, if the module has the same named inputs and outputs then you can add a line in the `script` section like below (another example [here](https://github.com/nf-core/modules/blob/e20e57f90b6787ac9a010a980cf6ea98bd990046/modules/lima/main.nf#L37)) which will raise an error asking the developer to change the `args.prefix` variable to rename the output files so they don't clash.
+- As a result of using this syntax, if the module could _potentially_ have the same named inputs and outputs add a line in the `script` section like below (another example [here](https://github.com/nf-core/modules/blob/e20e57f90b6787ac9a010a980cf6ea98bd990046/modules/lima/main.nf#L37)) which will raise an error asking the developer to change the `args.prefix` variable to rename the output files so they don't clash.
```
nvnieuwk, line 393:
```suggestion
- As a result of using this syntax, if the module could _potentially_ have the same named inputs and outputs add a line in the `script` section like below (another example [here](https://github.com/nf-core/modules/blob/e20e57f90b6787ac9a010a980cf6ea98bd990046/modules/lima/main.nf#L37)) which will raise an error asking the developer to change the `ext.prefix` variable to rename the output files so they don't clash.
```
:scream:

jfy133:
#small-mound-i-feel-strongly-about-but-not-enough-to-fight-everyone

mahesh-panchal:
The issue is more we still don't understand the objection. Since you're experienced, at least I feel like, there is something we're not understanding about your viewpoint. 

We've established though that the developer still has control.

jfy133:
> The issue is more we still don't understand the objection. Since you're experienced, at least I feel like, there is something we're not understanding about your viewpoint.
> 
> We've established though that the developer still has control.

I can't remember the depths of the discussion anymore, but mostly it's a development style. 

It firstly irks me because I firstly hate pipelines that tag loads of extra suffixes at the end (`_sorted_indexed_sorted_indexed`), however shouldn't happen here because of course it's just meta.id + prefix/suffix every time.

But the main reason is: I strongly feel that developers should think very carefully about how output is presented to users. Automating output names with defaults to making the development experience 'easier' at the expense user experience is not a good practice in my personal opinion. Sure, it _can_ be overridden by an organised pipeline developer, however most people are too busy and will take any opportunity to skip steps if they can. By allowing default prefixes it'll make people not think about this, and thus not think carefully about what goes in the results directory etc and just result in big mess and harder and less attractive to use the output. This of course would be mitigated with good documentation, but that still remains poor across all of bioinformatics (/rant).

I _personally_ prefer having a hard error when having a name conflict, as it _forces_ the developer to think carefully about the name, then should it even be  presented to the user, and then logically where should it go in the output directory etc etc. Rather than 'hoping' they'll do some TLC.

But I do recognise that this is a personal opinion/development style about a relatively minor point so I won't block it for that reason when I've been outvoted ;).

mahesh-panchal:
> It firstly irks me because I firstly hate pipelines that tag loads of extra suffixes at the end (`_sorted_indexed_sorted_indexed`), however shouldn't happen here because of course it's just meta.id + prefix/suffix every time.

Agreed. Perhaps this should be a pipeline linting check (warning) or an nf-test check. Split filenames on underscores/periods/non-alpha-numeric characters and check the number of unique parts against total (duplicate words) and that the total is not more than say 7 (ultra-long names).

> But the main reason is: I strongly feel that developers should think very carefully about how output is presented to users. Automating output names with defaults to making the development experience 'easier' at the expense user experience is not a good practice in my personal opinion. Sure, it _can_ be overridden by an organised pipeline developer, however most people are too busy and will take any opportunity to skip steps if they can. By allowing default prefixes it'll make people not think about this, and thus not think carefully about what goes in the results directory etc and just result in big mess and harder and less attractive to use the output. This of course would be mitigated with good documentation, but that still remains poor across all of bioinformatics (/rant).
 
I think we all agree on the first sentence. I disagree that this change is at expense of user experience though. While it makes developer experience better by not automatically resulting in a filename collision, the user should see at most one `_something` tacked on ( maybe we need to be explicit about how the default prefix can differ from `meta.id`, for example the default prefix cannot be based on `file.baseName` ). So then even in a chain of these modules with defaults changed from `meta.id` because they might commonly result in collision, there should be at most one `_something`. 

Conversely though, are we currently making the user experience better by how we're setting the `meta.id`? For many developers it's just a key to make something unique for `join`ing. A busy developer is just as likely to skip proper value setting of `meta.id` too.

> I _personally_ prefer having a hard error when having a name conflict, as it _forces_ the developer to think carefully about the name, then should it even be presented to the user, and then logically where should it go in the output directory etc etc. Rather than 'hoping' they'll do some TLC.

I think this is the key part though. Putting in developer roadblocks means the developer has to directly act. I don't think it'll force them to put TLC into it though. Workflow design ( including naming outputs, etc) is still a skill imo. 

Sorry, I guess this wall of text wasn't necessary, but maybe there is one thing we still need to do, and that's define how much the prefix can differ from `meta.id`.

SPPearce:
Huh, I thought this had actually been merged a while ago.
Can we merge this now?

PR #2695, opened 2024-08-21
Evidence stats page

edmundmiller:
This is the initial rewrite I threw together yesterday(emphasis on how quick this was to set up and get hacking on).
 
I pulled the database off mysql and tossed it in motherduck(This could also probably be a csv on hugging face or S3 if MotherDuckd wants to get greedy)
Scripts and some data: https://huggingface.co/datasets/nf-core/stats/blob/main/db/README.md(this was the hardest part getting connected to the DB :sweat_smile:)

Threw the credentials in the Evidence template and it was off to the races.

https://nf-core-stats.evidence.app/

### Some questions answered

1. At first I started out with a separate repo because I couldn't get it to `npm run dev` when I tossed it in the website repo. It's kinda of a Evidence bug, kinda an issue with npm workspaces. I tried switching to pnpm, and it just worked. That's definately a seperate debate, but just calling it out, so bear with me if that's a blocker this is a PoC.

2. Just deployed it with their cloud offering because it was a click. They have netlify instructions or you just build it yourself.

3. It runs as a static site generator. I think we just stick with that while I prove out recreating the current stats page. I'm hoping there's some way to just injest the charts or components, but I think worst case scenario, we can just wrap the html in Astro, like what's happening with the old site currently. Just need to pull in the theming.

Main point is it's the kind of dev workflow that I think will let the community hack on it. It's simple, and quick to get up to speed, and I'll be excited to see what people build with it!


edmundmiller:
Arguement for why this isn't in nf-core/website

- Evidence is a completely different build system, not using Astro
  - Just going to wrap it as an iframe
- There's also the whole data-engineering side. That could live in the nf-core/ops monorepo, but it might be better to just start yet another monorepo.

ewels (approved):
Given how badly outdated the current one now is, happy to merge this and incrementally improve 👍🏻

mashehu (changes requested):
needs to use nf-core styling, incl. dark theme 😎

mashehu (approved):

mashehu:
Tbh i don't think this is in a state to replace the current stats page. Too many placeholders, too little different stats and it will feel quite out of place with the current styling.

ewels (changes requested):
Some stuff that'd be good to sort before merging:

- [ ] Side navbar is missing on mobile
- [ ] Links on home don't correspond to side nav
- [ ] Home > Views is the same as Code Stats > Repository Traffic? Delete Home > Views.
- [ ] Add note to empty pages explaining that we know they're empty.

ewels:
> [!CAUTION]
> Refused to display 'https://nf-core-stats.evidence.app/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.

This may be a problem..?

edmundmiller:
The latest deployment seems to work @ewels!
image


> Tbh i don't think this is in a state to replace the current stats page. Too many placeholders, too little different stats and it will feel quite out of place with the current styling.

@mashehu I imported all the static data from the current stats page into some CSVs and recreated the plots. Feels way less half-finished now. 

As for styling, I'd love some help on that; the docs cover it. I don't know how the website styling works here. https://docs.evidence.dev/reference/themes-and-layouts/

ewels:
Can we change the dark mode background colour to 

CleanShot 2025-02-03 at 14 09 22@2x

ewels (approved):
letsgo

PR #2713, opened 2024-09-04
Add single cell analysis hackathon group

nictru:
I changed the description a bit, since the text on slack was more written as an internal explanation while this version is more friendly towards participants. It's a bit longer than the others, if that's an issue I can shorten it.


ewels (approved):
Great, thanks!

PR #2714, opened 2024-09-05
Use auto overflow for profile names

nictru:
This should prevent displaying scrollbars when they are not needed, like here:
image

The auto overflow should still display the scrollbar whenever needed, as described [here](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x#auto).

I think this looks more beautiful, but if there was a rationale behind the original implementation, let me know.


ewels (approved):
Nice, thanks!

PR #2732, opened 2024-09-20
Add new troubleshooting page on why a process is retrying

jfy133:
As requested in the thread here: https://nfcore.slack.com/archives/C031QH57DSS/p1726754158971039

@netlify /docs/usage/troubleshooting/retries


jfy133:
Thanks to @Matvi for the suggestion :)

christopher-hakkaart:
It might be hard to find (which isn't ideal), but I think there is some very similar documentation under [Crash halfway](https://nf-co.re/docs/usage/troubleshooting/crash_halfway), which also links out to [Quick reference](https://nf-co.re/docs/usage/getting_started/configuration#tuning-workflow-resources), which has a section on tuning resources.

I would be reluctant to add another page describing the same thing (or something that is very similar), can we modify the titles on these other pages to make them more findable?

PR #2775, opened 2024-10-08
Minor fix (NOTE in list)

chriswyatt1:
Cannot have a NOTE declaration within a list, 

Must be in this format:

`> **NOTE**: Use _exactly_ the same version as in the code (e.g. `1.0.0`) - **do not prefix with v** (e.g. not `v1.0.0`).`

Maybe there is another way, but this works.

@netlify /docs/checklists/pipeline_release


mashehu:
sorry, reverted your change, the problem was not the element, it was the indentation. While working on this also saw that somehow a lot of the checklist progress logic disappeared. added it back now.

chriswyatt1:
Ah ok, thanks. Didn't realise you could fix with indentation

PR #2889, opened 2024-12-07
Update metromap diagram components cheatsheet

LouisLeNezet:
In this PR the metromap component cheatsheet is updated.
Every icons is standardized to 10px multiple.
New icons are added and files from Maxime Garcia are integrated.
The library for drawio is added for an easy use of all components.

@netlify /docs/guidelines/graphic_design/workflow_diagrams


jfy133 (approved):
LGTM although the updated table isn't rendering for me

PR #2895, opened 2024-12-11
Adding references-assets to ignored_repos.yaml

maxulysse:
I want to store https://github.com/nf-core/references/tree/dev/assets outside of the #references repo, because it's data that is disconnected to the pipeline I could then update one without updating the other


mashehu (approved):

maxulysse:
thanks @mashehu

PR #2913, opened 2025-01-08
Update ignored_repos.yaml

maxulysse:
references is actually a pipeline


itrujnara (approved):
It definitely looks like a pipeline

mashehu (approved):

PR #2945, opened 2025-01-16
Leaflet map: style updates

ewels:
* 100% width instead of 90% on Leaflet component
* `mt-5` on `div` for EventMap component

Would like to doublecheck centre point and zoom level too, make sure we're not cutting folks off. Though I seem to remember that it's difficult to get a good balance.

@netlify /events/2025/hackathon-march-2025

### DIFF
--- a/sites/main-site/src/components/Leaflet.svelte
+++ b/sites/main-site/src/components/Leaflet.svelte
@@ -68,7 +68,6 @@
 <style lang="scss">
     .map {
         height: 480px;
-        width: 90%;
     }
     @media (max-width: 767.98px) {
         // md-breakpoint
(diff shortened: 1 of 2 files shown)


ewels:
Yeah, New Zealand is still cut off 🙈 

At leat we have West Coast US now though:

CleanShot 2025-01-16 at 12 16 57@2x

Will see if I can tweak it a bit more.

ewels:
Looks like the map is already zoomed out as far as it'll go 🫠 

Not entirely sure what we can do to get NZ in as well.. @mashehu any ideas?

mashehu:
played a bit with zoom and zoom center and NZ is visible now (alaska is there twice, but I don't think we will have a site there, so shouldn't be too confusing). there is a weird non-linear jump from 1.49 to 1.5 zoom, so this is the best I could do.

mashehu:
@nf-core-bot fix linting

ewels:
Gaah I didn't see the `minZoom` attribute 🙈 

Why back to 90% width and smaller height? Is that to avoid too much alaska or something?

PR #3003, opened 2025-01-30
CZ Biohub -> Seanome

olgabot:
Also need to upload the Seanome logo as an svg


mashehu (approved):

ewels:
@olgabot do we need to remove CZI or could we just add Seanome as a new entry?

Or does CZI have nothing to do with nf-core any more? We don't really have a policy for removing folks from this page. I'm inclined to leave in place even if only a past contributor..

mashehu (approved):

PR #3004, opened 2025-01-30
Umiami patch2

glichtenstein:
Venue confirmed. The LAB Miami co-working space, up-to 20 people. The hosting Institution is the Sylvester Comprehesive Cancer Center. 

@netlify /events/2025/hackathon-march-2025/university-of-miami

### DIFF
--- a/sites/configs/package-lock.json
+++ b/sites/configs/package-lock.json
@@ -1,0 +1,7428 @@
+{
+    "name": "@nf-co.re/configs",
+    "version": "0.0.1",
+    "lockfileVersion": 3,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "@nf-co.re/configs",
+            "version": "0.0.1",
+            "dependencies": {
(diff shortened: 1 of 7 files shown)


ewels:
Not quite sure what's happened here @glichtenstein - looks like you've been having a bit of a fight with the Prettier markdown formatting? 😅 

There shouldn't be any changes in the lock files, I've ported your changes over to https://github.com/nf-core/website/pull/3008 and fixed the syntax so that we can get this in.

For next time, I'd recommend using pre-commit to handle these kinds of problems. It's a one time setup:

```bash
pip install pre-commit  # install pre-commit globally
pre-commit install  # activate pre-commit for this repo clone
```

After that, every time you run `pre-commit` then the markdown syntax will be checked. If anything is wrong, the commit will be aborted and the changes made. Then you can do `git add .` to stage the changes, run `git commit` again and it should pass. Push that and the CI should not complain.

PR #2996, opened 2025-01-28
add frontmatter option to hide event export options

mashehu:
@netlify /config.ts

PR #3013, opened 2025-01-30
don't show time for full day events, simplify date handling

mashehu:


ewels:
Looks good! Listing component needs to be updated I think though (also still has the export button):

CleanShot 2025-01-30 at 11 46 16@2x

ewels:
Great! For the export button, I meant that this component wasn't updated in https://github.com/nf-core/website/pull/2996

But looks good, let's merge and can tackle that in another PR 👍🏻

PR #3018, opened 2025-01-30
Add argentina to countries map

nschcolnicov:
Argentina initials don't render in Hackathon [main page](https://nf-co.re/events/2025/hackathon-march-2025):
image

Not sure if this is the only reason why it isn't rendering, but I'm not sure how to test these changes to verify that it fixes the error. @mashehu since you authored the PR that introduced the automated table generation, could you verify if there is anything else needed? I'm happy to implement any further changes


ewels (approved):
Looks good, thanks!

ewels:
Note that there's probably a better or at least more comprehensive solution to this, we were just doing this in a rush.

PR #3031, opened 2025-02-03
Fix old hackathons, reorganise components for new projects

ewels:
Turns out that our fancy new hackathon project listing broke old hackathon event pages: https://nf-co.re/events/2024/hackathon-barcelona#list-of-projects

I added back the old components to fix those, then reorganised the various components for the new hackathon project listing into a directory to keep them together.

@netlify /events/2025/hackathon-march-2025


ewels:
The git diff is getting very confused because of the filenames, but basically:

* `sites/main-site/src/components/event/HackathonGroup.astro`
    * Previous version of this component, from [here](https://github.com/nf-core/website/blob/43ce44d840d2d22d52aadb7609e04e6cae62e0ce/sites/main-site/src/components/event/HackathonGroup.astro)
* Everything else: moved into a `HackathonProjects` directory to keep them together, with renamed files.

I think that these are the only meaningful changes.

mashehu (approved):

PR #3056, opened 2025-02-10
hard-code height of stats iframe

mashehu:
because some of the subpages have a height longer than the initial screen, this is the only solution for now.

@netlify /stats


ewels (approved):
Should remove the `vh-100` class, but otherwise LGTM

sites/main-site/src/pages/stats.astro
```diff
@@ -19,3 +19,8 @@ import PageLayout from "@layouts/PageLayout.astro";
             height="100%"></iframe>
     </div>
 </PageLayout>
+<style>
+    .iframe-container {
```
ewels, line 23: Probs no need for the !important if you remove the height class

Issue #2722, opened 2024-09-17
Pipelines: Add page listing modules


ewels:
Each module says which pipelines use it, but we don't have the inverse.

It'd be nice to have the inverse: a tab on the pipeline pages listing all modules that are used. Can be summary boxes, similar to the main modules listing page.

mashehu:
we are running out of space for tabs on <20" screens imo. how about adding it to the sidebar similar to how it is on the modules pages

lmReef:
I've been playing around and have a few notes/questions, I'm keen to work on this issue so let me know what you guys think

## Tabs
There seems to be enough space for another tab except at screen widths between `990px - 1090px` where it starts to look a bit janky with the release tag wrapping to a new line:

![Image](https://github.com/user-attachments/assets/413f5cbb-8436-4295-a809-7eabdd990768)

## Sidebar
Adding the info to the sidebar similar to the module pages works well when there are only a few modules/subworkflows to show. When there are quite a lot however, the sidebar ends up overflowing and the current `overflow-y` behavior is just `scroll` which I reckon isn't a great user experience.

I think it'd be better to limit the results shown in the sidebar and add a "show all" button that either
1. expands the list to show them all (basically as pictured below)
2. links to a new `modules` tab

else we could just leave it as-is with the `overflow-y: scroll` behavior for the time being

![Image](https://github.com/user-attachments/assets/4bcb6787-b03b-412f-b7ed-3badcab3fbbb)

![Image](https://github.com/user-attachments/assets/d369a719-e686-4769-b3df-3bb6cba5c8a9)

ewels:
Oh nice! Gut feeling is sidebar with a button to "show more" as you say. Sounds like an ok compromise.

Are you able to put in a PR? Should get a deployment preview that we can play with then to get a feel for it.

Thanks for this!

lmReef:
Opened a PR here #3094. Forgive the implementation, I've not used Astro before hah. I'll refactor it properly once we've decided what direction it's going

PR #3094, opened 2025-02-19
Show included modules on pipeline pages

lmReef:
Preview for #2722

image


mashehu:
hmm, maybe make the "show more" button a bit more subtle, e.g. similar to how we do it in the module cards
Screenshot 2025-02-20 at 09 12 44

lmReef:
So the implementation for this in Astro is looking more scuffed than it needs to be. I think it could be better to abstract this out from the other places its implemented (module/subworkflow [cards](https://github.com/nf-core/website/blob/main/sites/main-site/src/components/component/ComponentCard.svelte#L39-L103) and [pages](https://github.com/nf-core/website/blob/main/sites/main-site/src/components/sidebar/SidebarStats.astro#L20-L87)) into a new svelte component so we can reuse it across all 3 places.

This approach should make it easier to maintain and fix issues like [this](https://nf-co.re/modules/arriba/) going forward too.

I can do up a draft for now in svelte 5 and wait for #3047 before opening a new PR/updating this one. Let me know what you think.

Probably something along these lines with a few extras for handling the differences with pipelines:
``` svelte
<script lang="ts">
    const {
        tags,
        maxSliced = 5,
        type,
    }: { tags: string[]; maxSliced?: number; type: "modules" | "subworkflows" | "pipelines" } = $props();
    let expanded = $state(false);
</script>

<div class="tags-section mb-2">
    {#each expanded ? tags : tags.slice(0, maxSliced) as tag}
        <a
            href={`/${type}/${tag.replace("/", "_")}/`}
            class="badge fw-normal border border-info-subtle bg-info-subtle text-body me-2 text-decoration-none"
        >
            {tag}
        </a>
    {/each}
</div>
<span
    class="text-small cursor-pointer tags-show-more"
    title={`click to show all ${tags}`}
    on:click={() => (expanded = !expanded)}
    on:keydown={() => (expanded = !expanded)}
    role="button"
    data-bs-toggle="tooltip"
    data-bs-delay="500">+{tags.length - maxSliced} more {tags}</span
>
```

mashehu:
#3047 is merged so give the component approach a shot

lmReef:
awesome! I'll get started

lmReef:
I don't think the tab is necessary just tried it out as an option, happy to remove it and move along with the sidebar!

lmReef:
@mashehu seems I'm not authorized to push to main so requested review :+1:

PR #3183, opened 2025-03-13
hlatyping_project

mapo9:
Project regarding the nf-core/hlatyping pipeline for the hackathon march 2025. 

A project wrt to hla-typing was already proposed by @natanast and @sfragkoul, however they proposed to add an hla-typing tool to nf-core/airrflow. As they were probably unaware of the hlatyping pipeline and hlatyping is out-of-cope for airrflow, I added them as leaders to this project here and would suggest to remove the airrflow project and join forces in nf-core/hlatyping.  

I hope this is ok for you @natanast and @sfragkoul?

@netlify /hackathon-projects/hackathon-march-2025/picking-up-hlatyping


sites/main-site/src/content/hackathon-projects/hackathon-march-2025/picking-up-hlatyping.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Reviving nf-core/hlatyping
+category: pipelines
+slack: "https://nfcore.slack.com/archives/CFVDZCA23"
+intro_video: ""
```
mashehu, line 5:
```suggestion
```

sites/main-site/src/content/hackathon-projects/hackathon-march-2025/picking-up-hlatyping.md
```diff
@@ -0,0 +1,31 @@
...
+  mapo9:
+    name: Mark Polster
+    slack: "https://nfcore.slack.com/archives/D05SAA3LL6P"
+---
+
+This project aims to get nf-core/hlatyping back on track.
+
+## Goals
+
+1. Template update (last touched 2022 :o)
```
mashehu, line 24:
```suggestion
1. Template update (last touched 2022 😬 )
```

mapo9: awesome, thanks. didnt now how to add an actual emoji :D

mashehu:
waiting on a reply from @natanast or @sfragkoul before merging

natanast:
Hi @mapo9 ,
 
Thanks for reaching out and for your interest in our project for the hackathon! I appreciate your input regarding nf-core/hlatyping, but our initial goal was to develop a standalone module for arcasHLA that could be integrated into multiple pipelines—we used airrflow as an example.
 
I understand your concern that HLA typing might not fit within airrflow, but the module itself would still be useful for different workflows. Once developed, it can be added to nf-core/hlatyping and potentially other pipelines in the future.
 
Regarding the hackathon project, it will be nice to work together under one project. We would like to propose a new title that emphasizes the development of new modules rather than specific pipelines, such as 'Introducing HLA Modules."

mapo9:
Hey :)
wrt to the module, there is already an nf-core module for arcasHLA , I am not so familiar with the tool, do you require more than this to do typing?
You can find the module [here](https://nf-co.re/modules/arcashla_extract/) if you havent seen it already.
But I agree with you totally that the modules built during the hackathon should be nf-core modules and not local for specific pipelines!

I personally dont see that any hla typing modules will be part of any other pipeline as the use-case is quite specific and in nf-core we try not to have too much overlap between pipelines.

I am happy to change the name of course to get the emphasis away from only the pipeline itself. But I would focus on getting the pipeline to the newest template etc before I join you creating new modules. 
So how about we simply name it "HLA typing modules and pipeline"  or something in this direction?

What do you think?

natanast:
Thanks for the clarification! I wasn’t aware of the existing arcasHLA module—I'll definitely check it out to see if it covers everything we need for HLA typing. 
"HLA Typing Modules and Pipeline" sounds perfect! :)

mapo9:
Ok, great. Then I think we got it for the new project! Looking forward to hack!!

@mashehu, I changed the name now and removed the old project in this PR. If you are happy, I think we can merge.

mashehu (approved):

PR #3191, opened 2025-03-13
Initial attempt at a new Weekly Helpdesks page

ewels:


sites/main-site/src/pages/weekly_helpdesk.mdx
```diff
@@ -0,0 +1,28 @@
+---
+layout: "@layouts/MarkdownTocLayout.astro"
+title: Weekly Helpdesks
+subtitle: Join us to get help with your nf-core pipelines
+md_github_url: https://github.com/nf-core/website/blob/main/sites/main-site/src/pages/weekly_helpdesk.md
```
mashehu, line 5:
```suggestion
md_github_url: https://github.com/nf-core/website/blob/main/sites/main-site/src/pages/weekly_helpdesk.mdx
```

sites/main-site/src/pages/weekly_helpdesk.mdx
```diff
@@ -0,0 +1,28 @@
...
+- APAC: Wednesdays 3pm NZDT
+- EMEA: Thursdays 10am CET
+
+:::tip
+
+In theory, this should show the above times in your local time zone,
+but take with a pinch of salt as it often doesn't work (eg. daylight savings) -
+if in doubt, believe the times above!
+
+- Tuesdays <LocalDateTime date={new Date(1741716000000)} date_options={{ hour: "2-digit", minute: "2-digit" }} client:visible />
```
mashehu, line 24: this will not work, because it is a fixed date, so will never switch to DLS times. 
use something like https://date-fns.org/v4.1.0/docs/nextWednesday instead

mashehu: I would either use date-fns or just hard coded times.

ewels: Had a quick play in Cursor, but tied myself in knots. If anyone else fancies having a go, please take over. Otherwise can comment out this block for another time.

mashehu: I think I found a solution

ewels: Nice! Shows as "UTC" for me for some reason, instead of CET?

CleanShot 2025-03-14 at 01 38 13@2x

I kind of preferred it in a separate tip admonition to be honest. Was a bit easier to find then, took me a while to see where this new output was when I was looking just now.

mashehu: Ah, yes it's UTC because that string is server rendered 🤦‍♂️

maxulysse (approved):

PR #3195, opened 2025-03-14
Update ignored_repos.yaml

maxulysse:
Remove repo that has changed name.
`nf-co.re` has been `website` for a while, I do think we can remove it from this list


mashehu (approved):

mashehu:
same could be said about `nf-core.github.io`, but I actually don't care how tidy this file is...

PR #3200, opened 2025-03-14
Add mention to AWS full test on the release checklist

JoseEspinosa:
@netlify /docs/checklists/pipeline_release


FriederikeHanssen (approved):
caveat: the full size from dev may not be available? but they should run on dev

PR #3281, opened 2025-03-26
Revert Maxime's newer assewrtion example but not really working example to my original one

jfy133:
Because grep only looks in the first level directory, doesnt go inside, so if you have subdirectories you just get a list of those with `listFiles()` and the grep of filenames/patterns fails 

@netlify /docs/contributing/nf-test/assertions


maxulysse (approved):
To be fair this works well when you don't have weird stufff.

But I will finish my functions with nft-utils, and then re-update this part

PR #3288, opened 2025-03-30
Make sticker page load stickers from logos automatically at build time.

FloWuenne:
# Description

Instead of using a static defined json file to include stickers, this approach will use a fetchStickers function to get stickers from the logos repository at build time. I tried to use astro's content loader to achieve this but hit a couple of roadblocks and couldn't make it work. I think this is a reasonable intermediate for the time being!

@netlify /stickers


FloWuenne:
Nice, gotta check out that content  collection code 👀 !

FloWuenne:
Nice work refactoring @mashehu! 
giphy

FloWuenne:
From my point of view looks good and you can merge whenever you want. Seems like all functionality is maintained!

PR #3294, opened 2025-04-01
Add suffix for correct reference to config file

Joon-Klaps:
Currently the URL to the config file is incorrect. The generated URL seems to miss the `.config` suffix. 

For example on https://nf-co.re/configs/aws_tower/ the `See config file on GitHub` refers to 
`https://github.com/nf-core/configs/tree/master/conf/pipeline/demultiplex/aws_tower`

Not certain if the URL should be updated (this PR) or another argument for `${config}` should be created instead such as `${config.fileName}`


mashehu (approved):
great find, thanks!

PR #3307, opened 2025-04-09
Update state-of-nf-core-CI.md

maxulysse:
@netlify /blog/2025/state-of-nf-core-CI


mashehu (commented):
That's part is just talking about general CI, which is already part of the template

mashehu:
fixed in #3308

PR #3336, opened 2025-05-05
add bytesizes to pipelines

maxulysse:
@netlify /pipelines


ewels:
Hmm, can't see these showing up in the deployment preview. @mashehu is that some quirk of how the previews render?

maxulysse:
> Hmm, can't see these showing up in the deployment preview. @mashehu is that some quirk of how the previews render?

I have an idea, let me try it

maxulysse:
ok so it's not the quotes

mashehu (approved):
Tested it locally

PR #3338, opened 2025-05-06
catch incorrectly parsed directives

mashehu:


mashehu:
The build needs some time (and it isn't coming up in the deploy previews, because we don't include pipeline pages there). Now the changes are live and the table looks good.

jonasscheid:
Great! I see it, thanks!!

PR #3372, opened 2025-05-28
Sticker groups: Singular

ewels:
Double plural "Hackathons Stickers" doesn't make sense. Should be "Hackathon Stickers"

@netlify /stickers


ewels:
I think that this is going to break the stickers page because the plurals are used in file paths.

PR #3378, opened 2025-06-02
Add `run-nf-test` action repo

edmundmiller:
Or should we just change the `setup-nf-test` repo to `nf-test-action`?


mashehu (approved):
I think I prefer having two different GHA, but please bring this up on the nf-core core slack channel

PR #3410, opened 2025-06-23
Revert "Revert "nf-utils => nf-core-utils""

mashehu:
Reverts nf-core/website#3409


edmundmiller:
@mashehu Why the revert? Apologies if it broke a build, I was a little slow on the rename, but it's done now.

https://github.com/nf-core/nf-core-utils

Also, the plugin has been released, so no going back.

mashehu:
it's a revert of the revert. we had some problems because the repo wasn't renamed, while the yaml file was already updated. next time, we should do it the other way around

PR #3527, opened 2025-10-02
Add timeline mermaid diagram for core team

jfy133:
Because we :metal: 

@netlify /about/governance


christopher-hakkaart:
I do love a good graph, but it is very small. Additionally, I wonder if this information is superfluous to the page's purpose, which is primarily to describe the teams, their responsibilities, and list their members.

Would it be better to have another section on the profile button, showing `2018-2025` or whatever? Maybe bury the graph in a dropdown?

PR #3597, opened 2025-10-28
correct time for mag bytesize

FranBonath:
@netlify /events/2025/bytesize_mag_V5


mashehu (approved):
what time zone is that?

PR #3636, opened 2025-11-10
Add section on pre-existing pipeline contributions

itrujnara:
We have seen several submissions where teams want to include a pre-developed pipeline, and so far, we have no clear guidance on what they should expect. This PR proposes an additional section in the pipeline creation overview to set expectations for this type of contribution.

@netlify /docs/tutorials/adding_a_pipeline/overview


ewels (approved):
Nice addition 👍🏻 

I wonder if it would make sense to add any of these concepts to actual guidelines as well..

PR #3727, opened 2026-01-23
 [docs-v2] suggestions for 'running' sections

jfy133:
Got to mid-way through 'nextflow for your system'.

Also need to add TODOs on these thoughts:

- Running > Reference Genomes: I've moved the 'local genome' section first, as I think that is the most inclsuive and common method (iGenomnes only works specifically for essnetial model organisms) (TODO  not necessarily)
- Running > Reference GEnomes > RefGenie: The RefGenie instructions I'm suspicious of, it refers to the tool maknig a Nextflow config but I see no reference to this on the refgeni website (but note that the instructions are the same on the current docs site), [we need to find who wrote those... bgit glame isn't helping] (TODO ✅)
- Running > Running pipelines offline: I don't understand what it means by 'copy the plugins' from where do oyu see the list of plugins? [again need the author to clarify]  (TODO - no longer present, maybe now removed)
- Running > Running pipelines offline > Transfer reference genomes offline: link to Reference genomes is dead  (TODO ✅)
- `running/configuration/overview` > I feel the 'Nextflow loads configuration parameters sequentially and overwrites previous values. The loading order is:' should go first, because this applies to everything (e.g. you can give a `-c` even if you've also given a `-profile`)   (TODO ✅)
- `running/configuration/overview`: Is `Parameters in custom.config files will not override defaults in nextflow.config. Use -params-file with YAML or JSON format instead.` correct? I'm pretty sure `nextflow.coinfig` is even more base that `base.conf`   (TODO ✅)

@netlify /docs/contributing/documentation

### DIFF
--- a/sites/docs/src/content/docs/running/advanced-topics/managing_work_directory_growth.md
+++ b/sites/docs/src/content/docs/running/advanced-topics/managing_work_directory_growth.md
@@ -5,7 +5,7 @@ shortTitle: Managing work directory growth
 ---
 
 The management of intermediate files generated during Nextflow pipeline execution is a challenge for some workflows.
-As pipelines increase in complexity and scale, work directories can rapidly consume available storage, potentially leading to pipeline failures.
+As pipelines increase in complexity and scale, work directories (containing intermediate files) can rapidly consume available storage, potentially leading to pipeline failures.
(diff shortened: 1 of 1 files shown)


jfy133:
To finish review:

- [ ] run-pipelines.md
- [ ] advanced-topics/google-colb.md
- [ ] advanced/topics/managing_work_directory_growth.md

PR #3778, opened 2026-02-09
Add nf-core/variantbenchmarking as hackathon project

kubranarci:
@netlify /hackathon-projects/hackathon-march-2026/variantbenchmarking


georgiakes:
I think it's a bit confusing that way. Is this how it's supposed to be?

PR #3862, opened 2026-03-01
Modifying hackathon Seoul local hub information

jhlee0637:
I guess if I change the country name to 'Korea', I'll see the flag icon in [the list](https://nf-co.re/events/2026/hackathon-march-2026#registered-hubs).

image

Also  I changed some information about the Seoul Local hub.

@netlify /events/2026/hackathon-march-2026/seoul


ewels:
There's no magic here I think (off the top of my head, on my phone right now), just type in the emoji. See other pages for comparison.

Issue #10, opened 2018-08-07
Embed a google maps showing location of all contributors


ewels:
We now have addresses for each contributing institute on the _About_ page. It would be cool to have a google maps embedded that uses the GMaps API to show the global distribution of groups.

pditommaso:
wow! really nerd !! 🤓🤓

ewels:
This was done in #33

PR #3976, opened 2026-03-11
docs :add advisory example image

beumjin:
Added an advisory example image to replace the TODO comment. #10

@netlify /docs/community/advisories


public/images/advisory_example.png
mashehu: this is quite wide scree, can you maybe make the window a bit smaller when you make the screenshot?

PR #3993, opened 2026-03-11
Add permitted keys for ext map and rational

### LINKED ISSUE #3412: Document use of `task.ext` fields

`task.ext` fields outside singularity_pull_docker,prefix,suffix,args,when, and use_gpu are being added to modules for additional inputs rather than being documented in the meta.yml.

Add a description of what the purpose of task.ext variable is for. Users to override developer settings, on a per process level. 

List the permitted fields, and what they're used for. 

e.g. https://github.com/nf-core/modules/issues/8716#issuecomment-3004941384

Describe why more shouldn't be used.

**Describe alternatives you've considered**

**Additional context**

mahesh-panchal:
Adds which keys are permitted in the `ext` map, and the rational behind using the `ext` map. 

Closes #3412

@netlify /docs/contributing/components/ext_args

### DIFF
--- a/sites/docs/src/content/docs/contributing/components/ext_args.md
+++ b/sites/docs/src/content/docs/contributing/components/ext_args.md
@@ -46,14 +46,21 @@ so the script becomes:
 tool subtool -T -K test.bam > test.log
 ```
 
-The following table lists the available keys commonly used in nf-core modules.
+## Permitted ext keys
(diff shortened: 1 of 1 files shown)


mashehu (approved):
could you also make a PR against the `docs-v2` branch with these changes please

PR #3997, opened 2026-03-11
Added table of posters

jess789550:
Removed download links as they weren't working for me.

Replaced with actual SVG.

Is this ok?

@netlify /docs/community/brand/presentations


mashehu:
symlink is fine (/public is actually already symlinked into sites/docs for example)

PR #4021, opened 2026-03-12
Clarify custom config precedence and nf-core note

PeterTSmith1997:
Add a warning that parameters in custom.config do not override nextflow.config and recommend using -params-file with YAML/JSON. Add a note that nf-core uses base.config for resource defaults. Remove the duplicated warning and added more links in "Additional resources" section to clean up the configuration options page.

@netlify /docs/running/configuration/configuration-options

### DIFF
--- a/sites/docs/src/content/docs/running/configuration/configuration-options.md
+++ b/sites/docs/src/content/docs/running/configuration/configuration-options.md
@@ -80,7 +80,7 @@ If not, follow the repository instructions or the tutorial to add your cluster.
 See [Run configuration](https://training.nextflow.io/latest/nextflow_run/03_config/) for a guided configuration tutorial.
 
 :::note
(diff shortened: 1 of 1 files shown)


sites/docs/src/content/docs/running/configuration/configuration-options.md
```diff
@@ -79,6 +79,15 @@ If not, follow the repository instructions or the tutorial to add your cluster.
 
 See [Run configuration](https://training.nextflow.io/latest/nextflow_run/03_config/) for a guided configuration tutorial.
 
+:::warning
```
jfy133: I realise in retrospect that my comment is not clear what I was going for about 'moving this BEFORE'... to the point I can't remember what I was trying to say 😅 

I think this can go back to the original place for the time being (sorry about that @PeterTSmith1997 )

PR #4023, opened 2026-03-12
[docs-v2] remove outdated docs to refgenie

jfy133:
I proposed to remove this because:

1. The instructions didn't work for me
1.  I found no references to Nextflow nor nf-core on the RefGenie website, so I couldn't fix 1
2. The plugin doesn't seem to have been updated since 2021
3. The #refgenie channel has had only 2 messages since 2024, so does not seem to be used very often 

I consulted with the core team member previously working with refgenie, and we voted to remove the docs now, until we can have someone from RefGenie to help us overhaul them: https://nfcore.slack.com/archives/CE6SDBX2A/p1773323326029159

@netlify /docs/running/reference-genomes


ewels (approved):
Can add back in the future if requested / functioning

PR #4078, opened 2026-03-22
[docs-v2] Move prettier page to environment setup, fix URLs on getting overview.md

### LINKED ISSUE #4077: docs: move code formatting page outside of documentation (docs-v2)
Currently the documentation on pre-commit/prettier etc. is sitting in `documentation > code formatting`.

However pre-commit/prettier runs not just on markdown files but all files (e.g. checking for trailing new lines etc).

We should move it to a more logical place. 

Arguably, `getting started > environment setup` would be a better place for this. However having to run `pre-commit install` on each new repo is something people would likely forget. 

But it would be more visible under `environment setup` than `documentation` IMO

jfy133:
Close #4077 

Rationale:

- pre-commit/prettier is not just for markdown files, so doesn't make sense under `developing/documentation/`
- moved it to the other environment setup pages under 'getting started', as it is more natural place to sit (i.e., setting up your dev environment

Alternative

- Understandably, @christopher-hakkaart wants to keep getting-started rather slim
- We could instead make a new section under `developing` called 'Dev. environment' or 'Tools for developing' or similar
- We then replicate the original 'env set upo' getting-started pages by linking the same pages into the new section, and add additional pages like the `prettier` one 

Also fixes internal links on getting started/en set up over view page

@netlify /docs/get_started/environment_setup/overview


christopher-hakkaart (commented):
Don't care about adding prettier, but some suggestions around the framing (running, developing, contributing). See how it feels, otherwise, feel free to merge.

christopher-hakkaart (approved):

PR #4138, opened 2026-04-13
Updates documentation on dynamic ext.args in nf-tests

### LINKED ISSUE #4133: [Module testing] "module_args" method does not support calling nextflow variable like "meta.id"
Testing `ext.args` with method described here https://nf-co.re/docs/specifications/components/modules/testing#configuration-of-extargs-in-tests

Won't work if some of your passed args contains nextflow variable like `meta.id` (eg: `--sample ${meta.id}`)

It is particularly problematic for testing `ext.prefix` injection, because you very often do something such as `ext.prefix = "toolName_${meta.id}"`
<br>

I guess having `meta.id` work in `params` section of "main.nf.test" is complicated from `nf-test` devs point of vue

This could be at least mentionned in the docs, as a limitation of this method
<br>

**Describe alternatives you've considered**
If you need to call nextflow variable, you have to put your `ext.prefix` or `ext.args` directly in `nextflow.config` (as you would do usually for a pipeline)
<br>

**Additional context**
Answer from `nf-test` dev about this issue : https://github.com/askimed/nf-test/issues/360

mahesh-panchal:
- Updates docs to use `.minus()` instead of `.trim()` in `ext.args`.
- Adds info on dynamic parameter inputs in nf-tests. Resolves #4133

@netlify /docs/developing/components/ext-args

### DIFF
--- a/sites/docs/src/content/docs/developing/components/ext-args.md
+++ b/sites/docs/src/content/docs/developing/components/ext-args.md
@@ -71,12 +71,11 @@ Set `ext.args` based on parameter settings:
 ```groovy
 process {
     withName: '.*:FASTQC_UMITOOLS_TRIMGALORE:UMITOOLS_EXTRACT' {
-        ext.args   = [
-                params.umitools_extract_method ? "--extract-method=${params.umitools_extract_method}" : '',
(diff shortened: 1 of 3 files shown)


mahesh-panchal:
@nf-core-bot fix linting

mashehu (approved):

PR #4176, opened 2026-04-28
Update module specification stub section for simplicity

famosab:
Relates to https://github.com/nf-core/tools/issues/4231

@netlify /docs/specifications/components/modules/general

### DIFF
--- a/sites/docs/src/content/docs/specifications/components/modules/general.md
+++ b/sites/docs/src/content/docs/specifications/components/modules/general.md
@@ -609,7 +609,7 @@ Ideally, the stub block should reproduce as much as possible the number of files
 Use the syntax in the following example for stub files output as gzip compressed:
 
 ```bash
-echo | gzip > ${prefix}.txt.gz
+echo | gzip > "${prefix}.txt.gz"
 ```
 
 :::info{title="Rationale" collapse}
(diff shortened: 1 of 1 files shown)


sites/docs/src/content/docs/specifications/components/modules/general.md
```diff
@@ -609,7 +609,7 @@ Ideally, the stub block should reproduce as much as possible the number of files
 Use the syntax in the following example for stub files output as gzip compressed:
 
 ```bash
-echo "" | gzip > ${prefix}.txt.gz
+echo | gzip > ${prefix}.txt.gz
```
LouisLeNezet: Doesn't that fail with the new nf-core modules lint ?

famosab: yes thats why its updated here, see https://github.com/nf-core/tools/issues/4231

LouisLeNezet: I've proposed this change a few years ago, but this never took.
Happy to approve !

LouisLeNezet (approved):

sites/docs/src/content/docs/specifications/components/modules/general.md
mashehu:
```suggestion
echo | gzip > "${prefix}.txt.gz"
```
based on shell-check recommendations

SPPearce:
To be clear, I think we should allow both types in the linting, because we have been using both types throughout the repo.

PR #4188, opened 2026-04-29
blog: Add blog cover image

christopher-hakkaart:
- Green apples 🍏 

@netlify /blog/2026/new-core-members


ewels:
Blog already has a header image, so I guess this was done in another PR.

PR #4210, opened 2026-05-13
Add documentation for non-argument module inputs

itrujnara:
This PR adds a new section in module input/output specifications to describe best (and discouraged) practices for module inputs that are not tool arguments. This change aims to make existing practices explicit and promote consistency rather than create new rules. It is based on my perception of the consensus, so please let me know if anything is inaccurate.
The changes have been preliminarily checked for compliance with the documentation style guide with GPT-5.4-mini. 

@netlify /docs/specifications/components/modules/input-output-options


sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,19 @@ When one and only one of multiple argument are required:
 
   :::
 
+## Non-argument `val` channel inputs
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
+
+- If a module implements multiple subcommands of a tool, the subcommand SHOULD be provided through a channel.
```
mashehu: isn't the rule that subcommands should be their own module?

itrujnara: Generally yes, but there are some modules where a subcommand or subsubcommand is passed as an input (specifically, per Codex and confirmed manually: `HELITRONSCANNER_SCAN`, `LONGSTITCH`, `COOLER_CLOAD`, `MIDAS_RUN`, `QSV_CAT`, `UMICOLLAPSE`, `SAMBAMBA_DEPTH`, `VAMB_BIN`, `ANNOSINE`). I am not sure whether this is an accepted practice or unnoticed abuse. I don't feel too strongly about this point, so I am happy to leave it out if there is strong opposition.

itrujnara:
@nf-core-bot fix linting

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
mashehu: I would say unnoticed deviations

itrujnara: I'll wait a bit for differential opinions, but if others agree with you I'll remove it since I don't want to start a civil war here

jfy133: Yeah I agree, this sounds like things that slipped through review.

The only case I can imagine this being valid if it all subcommands share exactly the same options and arguments ,(so they apply to every subcommand), but then this would be very unlikely as they you might as well just control this via another argument.

itrujnara: Alright, changed it. I used SHOULD, since there might be very specific cases where this makes sense.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,18 @@ When one and only one of multiple argument are required:
...
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
+
+- If a module can output arbitrary or numerous file formats, the output format SHOULD be provided through a channel. The module SHOULD NOT perform bespoke string manipulation on input paths.
+  :::info{title="Rationale" collapse}
+  Modules can encounter numerous input name scenarios. Custom string operations necessarily make assumptions about the name of the file (for example that the name of a compressed file has at least two dots). Providing an explicit format input returns full control to the pipeline developer and reduces the risk of unexpected behaviour.
+  :::
+- If a module contains an optional pipe (for example: compression, sorting), the pipe SHOULD be controlled with a Boolean input channel.
+
+Non-standard `ext` fields (example: `ext.suffix`) SHOULD NOT be used to control module behaviour. `val` inputs SHOULD NOT be used for subcommands, separate modules SHOULD be used instead.
```
mahesh-panchal, line 94: Module linting prevents non-standard keys now. It throws an error if someone tries to use a non-approved ( coded into nf-core tools ) key for either `ext` or `meta`. This should probably still be here though.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,18 @@ When one and only one of multiple argument are required:
...
 
+## Non-argument `val` channel inputs
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
+
+- If a module can output arbitrary or numerous file formats, the output format SHOULD be provided through a channel. The module SHOULD NOT perform bespoke string manipulation on input paths.
+  :::info{title="Rationale" collapse}
+  Modules can encounter numerous input name scenarios. Custom string operations necessarily make assumptions about the name of the file (for example that the name of a compressed file has at least two dots). Providing an explicit format input returns full control to the pipeline developer and reduces the risk of unexpected behaviour.
+  :::
+- If a module contains an optional pipe (for example: compression, sorting), the pipe SHOULD be controlled with a Boolean input channel.
```
mahesh-panchal, line 75: Adding code examples would help clarity tremendously.

itrujnara:
@nf-core-bot fix linting

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
itrujnara: Added

itrujnara:
@nf-core-bot fix linting

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,45 @@ When one and only one of multiple argument are required:
 
   :::
 
+## Non-argument `val` channel inputs
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
+
+- If a module can output arbitrary or numerous file formats, the output format SHOULD be provided through a channel. The module SHOULD NOT perform bespoke string manipulation on input paths.
```
muffato: Here is a use case that I've definitely seen across multiple modules:

```suggestion
- If a module can output multiple file formats, the output format SHOULD be provided through a channel. The module SHOULD NOT infer the output format from the input path.
```

If there are other cases, I would add them all to the list – even if their sum is equivalent to "The module SHOULD NOT perform bespoke string manipulation on input paths" !

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,45 @@ When one and only one of multiple argument are required:
...
+
+  ```nextflow
+  input:
+  ...
+  val compress
+
+  script:
+  compress_cmd = compress ? "gzip" : "cat"
+  """
+  tool $in_file | $compress_cmd > $out_file
```
muffato: "useless use of cat" :)
```suggestion
  compress_cmd = compress ? "| gzip" : ""
  """
  tool $in_file $compress_cmd > $out_file
```

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,45 @@ When one and only one of multiple argument are required:
...
+  :::
+
+  :::info{title="Example" collapse}
+
+  ```nextflow
+  # incorrect
+  out_file = "${prefix}.${in_file.name.split('.')[1]}.gz"
+  # correct
+  out_file = "${prefix}.${out_ext}.gz"
+  ```
```
muffato, line 69: Under my updated rule, this pattern would only be discouraged for tools that support multiple output formats from the same input.
This would leave tools that support multiple input formats but always output data in the same format as their input. In this case, the format is redundant with the input file name. It would be _convenient_ if the module could infer the format automatically, and it would make sense for the module to derive it itself. As long as we feel certain about it.
This may mean having a recommended recipe in the docs for extracting the extension of a file, discarding an optional `.gz`

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,45 @@ When one and only one of multiple argument are required:
...
+
+  :::
+
+- If a module contains an optional pipe (for example: compression, sorting), the pipe SHOULD be controlled with a Boolean input channel.
```
muffato, line 75: I generally agree with having booleans to control output compression, but a module that already has an input `val` to define the output format could perhaps combine both ? e.g. allowing `fa.gz` as the output format.

muffato, line 75: Also, the wording here applies to inputs too. I would be disappointed if the modules were forced to be made "dumb" and unable to recognise a compressed file by its `.gz` extension. Especially when nf-core recommends compressing the output files, sort of forcing modules to support optionally compressed inputs for chaining.

muffato (commented):
I'm not entirely sold on removing all `getExtension()`. I feel this proposal would just shift the name parsing to (sub-)workflows.

That said, I agree with removing ambiguity from the modules. Modules should offer choice when there is choice (instead of picking one). And when modules are trying too hard to guess, when there's a risk they may get it wrong, modules should require the caller to be explicit about what they want.

Here are some further comments.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
itrujnara: It's not a very common approach for now, but I don't see any problem with it. Added a bullet point for this.

itrujnara: This is slightly problematic for `.gz` specifically, since it can mean both normal and blocked gzip, and some tools (e.g. `gunzip`) are only compatible with one of them. The most bulletproof to deal with it is `file`, but I understand that looking at the path makes more sense in most modules

muffato: Your comment is spot on and very timely. This morning, we've just been hit by a pipeline failing because the input file was regular gzip, not block gzip.
There's a bgxip module in nf-core that uses `htsfile` indeed to only skip recompression if the input file is already a _block_ gzip.

Some tools use `.bgz` as a file extension, but I fear it's not common enough to make it a recommended guideline.

itrujnara: As it stands, `.bgz` is not a standard and many tools (modules and pipelines alike) don't accept it, so I would stay away from it for the time being. Both `file` and `htsfile` can distinguish the two, but it's still rather annoying. I think we should leave the details out of the guidelines for now, since the ultimate implementation is quite context-dependent (i.e., `.vcf.gz` is likely to be a block gzip, but `.fq.gz` is probably not)

mahesh-panchal (approved):
I'm good with this. 👍🏽

muffato (approved):

mashehu (approved):

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,49 @@ When one and only one of multiple argument are required:
 
   :::
 
+## Non-argument `val` channel inputs
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
+
+- If a module can output multiple file formats, the output format SHOULD be provided through a channel. The module SHOULD NOT infer the output format from the input path.
+  :::info{title="Rationale" collapse}
```
jfy133, line 58:
```suggestion

  :::info{title="Rationale" collapse}
```

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,49 @@ When one and only one of multiple argument are required:
...
+  Modules can encounter numerous input name scenarios. Custom string operations necessarily make assumptions about the name of the file (for example that the name of a compressed file has at least two dots). Providing an explicit format input returns full control to the pipeline developer and reduces the risk of unexpected behaviour.
```
jfy133, line 59:
```suggestion
  Modules can encounter numerous input name scenarios.
  Custom string operations necessarily make assumptions about the name of the file (for example that the name of a compressed file has at least two dots).
  Providing an explicit format input returns full control to the pipeline developer and reduces the risk of unexpected behaviour.
```

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,49 @@ When one and only one of multiple argument are required:
 
   :::
 
+## Non-argument `val` channel inputs
+
+Input channel `val` declarations MAY be used to control behaviours of the module that cannot be expressed with arguments of the underlying tool.
```
jfy133, line 55:
```suggestion
Input channel `val` declarations MAY be used to control optional but critical behaviours of the module that cannot be expressed with arguments of the underlying tool.
```

Is this what you mean? 

At the moment reading this I'm not sure how this differs from the much more concise point 1 of the previous section (although we could have a an example):

> Mandatory non-file inputs are options that the tool MUST have to be able to be run.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
jfy133: This could be an example moved to the previous section

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
jfy133: I don't fully understand this point - what do you mean by 'nuermous input name scenarios'? and 'custom string operations' - for what? Is this sentence meant to be a justification here why you should have an explicit extension `val` channel to control this (as in the alternative methods cenarios are bad)?

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,49 @@ When one and only one of multiple argument are required:
...
+  out_file = "${prefix}.${in_file.name.split('.')[1]}.gz"
+  # correct
+  out_file = "${prefix}.${out_ext}.gz"
+  ```
+
+  :::
+
+- If the output format of a module is necessarily the same as the input format, the module MAY infer the output format from the input path.
+
+- If a module contains an optional pipe (for example: compression, sorting), the pipe SHOULD be controlled with a Boolean input channel.
```
jfy133, line 75: OK the more I read about this, this feels mostly more about how to define output names, in which case I think the title of the section should be changed accordingly and rephrased to be more active of what you SHOULD do (a lot of this seems to be more defensive).

E.g., a seciton on output naming would be more like:

1. Should use `${prefix}`
2. Avoid a default prefix as far as possible (pipeline developers should control this)
3. Should not include custom strings except if necessary to define extension (controlled by input channel) 

Or something like that

jfy133: Also technically this also overlaps with this section actually: https://nf-co.re/docs/specifications/components/modules/naming-conventions#command-file-output-naming

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
itrujnara: This stems from my observations while reworking the bgzip modules. Some implementations assumed that the input name would always be something like `variants.vcf.gz`, so the module could "fish out" the output extension by splitting on dots and taking the second item from the end. However, it is not guaranteed that it will always be the case, and files could have names like `database.gz`, where `database` is likely not the desired extension. At least in my opinion, the module should be agnostic to the exact file name and use inputs to delegate the details to the pipeline. I hope I have explained it clearly enough in the draft.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
itrujnara: This is not quite what I mean. This section aims to codify existing practice for inputs like `val action`, `val compress`, and `val out_fmt`, which are common in modules, but lack appropriate documentation. These input are not really "an argument to the tool", but are still an important argument to the module.

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
itrujnara: This is not meant to be a section on outputs, but rather on inputs. The current documentation largely suggests that every input to the module should be an input to the tool, which is clearly not the case. The involvement of output names is due to the drama of `val out_fmt` vs. `ext.suffix` (the former being the preferred practice according to maintainer discussions).

itrujnara: Basically what I want here is to move ideas from random Slack threads to the documentation

jfy133: >  The current documentation largely suggests that every input to the module should be an input to the tool, which is clearly not the case. 

I don't think this is necessarily true: https://nf-co.re/docs/specifications/components/modules/input-output-options#required-val-channel-inputs ('essential for executuion of the tool' does not necessarily reuqire this to be an input to the tool itself - but we could make this clearer if yo uwanted?)

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
jfy133: HRm OK, I think you've made the explainations a bit too generic...

So does it ultimately boil down to:

If a module can emit mutually exclusive output formats of the same output file (e.g. bam/sam/cram), this SHOULD be explicitly defined using a dedicated input `val` channel?

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
jfy133: Ah I think I'm starting to follow now.

I guess then just a single additional sentence of:

'This can include parameters to control execution of the tool, outside of the tool parameters itself - such as an optional post-execution compression step (to adhere to nf-core specification XYZ)' or something like this?

itrujnara:
@jfy133 could you please clarify whether you suggest to:
1. keep the new section, but phares it differently, or
2. scrap the new section and instead add information to the section above?

PR #4213, opened 2026-05-13
add params.json file (and build function) and pipelines_params page

### LINKED ISSUE #1251: Create a global parameters list
**Create a global parameters list**

Terminology between pipelines and shared assets can differ. To help preserve shared content and familiarity between pipelines, subworkflows and modules, it would be beneficial to create a reserved ontology. For example, parameter names such as --bwa_index and --bwa should be reserved.

A reserved ontology list needs to be created. There might be example from elsewhere we could use to start. We could also scrape all JSON schema files and build a big list (link to it in the writing pipelines tutorial). Final product could be a list with clear descriptions that can be used by developers to guide naming conventions.

mashehu:
Closes #1251 

@netlify /pipeline_params


mashehu:
@nf-core-bot fix linting

PR #4214, opened 2026-05-14
Fix poster examples links

vinisalazar:
Was searching for poster templates and came across this page: https://nf-co.re/docs/community/brand/presentations but the internal links were broken.

This PR fixes that

Thanks

@netlify /docs/community/brand/presentations


vinisalazar:
Preview is not working, will see if the relates to how paths are resolved in the build directory

vinisalazar:
Not sure why only 1 image preview is rendering. Could be a build issue? Download links are working though.

piplus2:
The problem likely resides in the image sizes. Can you try to set a smaller size for the png previews?

mashehu:
my guess is this is a deploy preview issue. I will merge and see if the same problem is happening in production

mashehu (approved):
Thanks for fixing this

mashehu:
nope didn't work. will try to fix this later

piplus2:
Can you try reducing the size of the pngs? Do we need that big previews?

mashehu:
i can do that. the problem does not seem to be the size, but the cross origin policy is somehow violated for these images.

mashehu:
oh, it might be the file size actually (I thought you were talking about the image dimensions). they seem to be over the netlify limit...

piplus2:
Yes, I saw that error code, I don't have access to the pc at the moment, just seen from chrome logs.

mashehu:
seems to have fixed it: #4225

PR #4244, opened 2026-06-02
Add metadata for agent-discovery

ewels:
Discovered via https://isitagentready.com/nf-co.re


sites/main-site/netlify.toml
```diff
@@ -6,7 +6,23 @@ environment = { NODE_VERSION = "22" }
...
+for = "/*"
 [headers.values]
 Access-Control-Allow-Origin = "*"
+
+# Advertise machine-readable resources to agents/crawlers from the homepage (RFC 8288 / RFC 9727).
+# Expose-Headers lets cross-origin browser JS actually read the Link header (it is not CORS-safelisted).
+[[headers]]
+for = "/"
+[headers.values]
+Link = '''</.well-known/api-catalog>; rel="api-catalog"; type="application/linkset+json", </llms.txt>; rel="describedby"; type="text/plain", </docs/>; rel="service-doc"; type="text/html", </sitemap-index.xml>; rel="sitemap"; type="application/xml"'''
```
mashehu: this should be in `BaseHead.astro` no?

sites/main-site/public/.well-known/api-catalog
```diff
@@ -0,0 +1,45 @@
+{
```
mashehu, line 1: is there a script to generate this file?

sites/main-site/netlify.toml
ewels: `BaseHead` is the HTML headers right? This are the HTTP request headers, not part of the HTML at all.

sites/main-site/public/.well-known/api-catalog
ewels: I don't think that there's any way to discover them really, so any changes would still have to be manually specified - I'm not sure that adding a script would really help much?

I can look into adding a CI to make sure it doesn't get left behind by accident.

sites/main-site/netlify.toml
```diff
@@ -6,7 +6,23 @@ environment = { NODE_VERSION = "22" }
 command = "npm run build -w sites/main-site"
 [images]
 remote_images = ["https://raw.githubusercontent.com/.*", "https://images.unsplash.com/.*", "https://avatars.githubusercontent.com/.*"]
+# Allow cross-origin reads of all resources (the whole site is public and unauthenticated).
+# Set Access-Control-Allow-Origin in exactly one rule to avoid duplicate/merged header values.
 [[headers]]
-for = "/_astro/*"
+for = "/*"
```
mashehu: this feels still a bit too risky for my taste. we still have some credentials on a server.

sites/main-site/netlify.toml
mashehu: ah, yes.
But I don't think we need all these points. sitemap is already in the header. all the other thinks are already ointed out in api-catalog, so no need to have them twice. crawlers should already be happy with api-catalog, from what i found.

 
```suggestion
Link = '''</.well-known/api-catalog>; rel="api-catalog"; type="application/linkset+json"'''
```

sites/main-site/netlify.toml
ewels: It only affects hot-loading via JS though, it doesn't really affect visibility at all. So we shouldn't have anything kicking around that we don't want loaded.

sites/main-site/public/.well-known/api-catalog
ewels: Added in 395d76164

.github/workflows/linting.yml
mashehu, line 1: instead of adding it to the linting.yml please add it to the `pre-commit-config.yml` then it gets automatically picked up here and should be fast enough

sites/main-site/netlify.toml
ewels: I'll narrow it down a bit anyway though. I was preferring simplicity so it was a conscious choice (also CORS just generally annoys me 😆), but it's an easy change.

sites/main-site/netlify.toml
```diff
@@ -6,23 +6,50 @@ environment = { NODE_VERSION = "22" }
...
+
+[[headers]]
+for = "/components.json"
+[headers.values]
+Access-Control-Allow-Origin = "*"
+
+[[headers]]
+for = "/pipeline_names.json"
+[headers.values]
+Access-Control-Allow-Origin = "*"
```
mashehu: the public folder is symlinked in all sub-sites, afaik. No need for this I think

sites/main-site/netlify.toml
ewels: Does it help to remove them? They're never going to change, and they're simply additional information for a pre-fetch of headers only? I don't really see any complexity cost here, it's not something we're going to have to maintain.

sites/main-site/netlify.toml
ewels: Fair, removed again in 155fe3070

sites/main-site/netlify.toml
mashehu: just want to keep the things as minimal as possible.

ewels: Generally speaking I'd agree, but here I don't think that there's any cost for keeping the more verbose version, and a tiny bit of benefit.

mashehu (approved):

PR #4246, opened 2026-06-08
Add an `AGENTS.md`

ewels:
Minimal `AGENTS.md`, added after asking Claude to debug phantom CORS errors. Hopefully helps sessions get going a bit faster.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,35 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents (Claude Code and others) when working with code in this repository.
+
+The source for **https://nf-co.re** — the nf-core community website. Built with **Astro** (static site generation), **Svelte** (interactive components) and **Bootstrap** (CSS), managed as **npm workspaces** with **Node 22**.
+
+## Architecture
+
+The repo is a **monorepo of independent Astro sites**, one per subdirectory of `sites/*`, each an npm workspace ([background blog post](https://nf-co.re/blog/2024/new-website-structure)):
+
+- `main-site` — the apex `nf-co.re`: homepage, blog, events, community/governance pages, and shared components/layouts that other sites import.
+- `docs` — documentation.
+- `pipelines`, `modules-subworkflows`, `pipeline-results`, `configs`, `stickers` — data-driven sites that pull content from the **GitHub API** at build time.
+
+You usually work on **one sub-site at a time**, not the whole repo.
+
+Two cross-cutting things are non-obvious and worth knowing before editing build config:
+
+- **Each sub-site deploys as its own separate Netlify site**, and the apex `main-site` stitches them together: it **proxies** many URL paths to the other sites via rewrite rules in `sites/main-site/public/_redirects` (e.g. `/usage/*` and `/contributing/*` are served by `docs`).
+- **Assets load cross-origin.** Each sub-site sets `build.assetsPrefix` (in its `astro.config.mjs`) to its own Netlify subdomain, so a page served under `nf-co.re` loads its JS/CSS/fonts from a different origin. This is why each sub-site's `netlify.toml` serves `/_astro/*` with an `Access-Control-Allow-Origin` CORS header.
+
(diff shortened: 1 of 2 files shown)


mashehu (commented):
can't we just symlink this to contributing.md? All this information is there already and we don't need to remember to keep things in sync

ewels:
You mean [`.github/CONTRIBUTING.md`](https://github.com/nf-core/website/blob/277a5c41a19950c052a14da2a4df4c679f853ad8/.github/CONTRIBUTING.md)? The two files look very different to me?

That file is mostly about how to add yourself to the list of contributors, and how to fork the repo and make PRs. None of that is relevant to an LLM that will already be checked out on a branch with a task to do (actually it'll make its performance actively worse).

It doesn't include any information about the build architecture, testing, Astro setup etc (arguably most Human contributors don't need to know any of this, as they can just edit a markdown file and see the deployment preview).

ewels:
Side note: playing around with how to get Claude to read `AGENTS.md` - the cleanest I can figure out is to add `.claude` to `.gitignore`  and then have the following file in there:

`.claude/CLAUDE.md`
```
@../AGENTS.md
```

I think this should get Claude to automatically read `AGENTS.md` without cluttering up the repo for folks using other LLMs.

mashehu:
Ah, sorry, was thinking of the wrong repo. For the website the readme has the same instructions as the agents.md

ewels:
That's closer, but it still doesn't mention tests etc. and is 300+ lines instead of 30. Also it's not automatically loaded into context like this file.

This isn't that much to keep updated is it?

Would be a good addition to this file to say read the `README.md` for instructions on adding events / advisories etc..

AGENTS.md
```diff
@@ -0,0 +1,39 @@
+# AGENTS.md
+
+This file provides guidance to AI coding agents (Claude Code and others) when working with code in this repository.
+
+The source for **https://nf-co.re** — the nf-core community website. Built with **Astro** (static site generation), **Svelte** (interactive components) and **Bootstrap** (CSS), managed as **npm workspaces** with **Node 22**.
```
mashehu, line 5:
```suggestion
The source for **https://nf-co.re** — the nf-core community website. Built with **Astro 6** (static site generation), **Svelte 5** (interactive components) and **Bootstrap 5.3** (CSS), managed as **npm workspaces** with **Node 22**.
```
if you mention one version, we should mention all, especially because in my experience it keeps producing outdated syntax for svelte. pain to keep updated though

AGENTS.md
```diff
@@ -0,0 +1,39 @@
...
+```sh
+npm install --workspaces                 # install dependencies for all sub-sites
+npm run dev --workspace sites/main-site  # dev server for one sub-site (astro dev)
+npm run build --workspace sites/docs     # production build for one sub-site
+npm run test-all                         # Playwright tests across all sub-sites
+```
+
+Tests are **Playwright** (`npm run test --workspace sites/<name>`, or `npx playwright test <file>` within a sub-site for a single test).
+
+**GitHub token:** the data-driven sub-sites (`pipelines`, `pipeline-results`, `configs`, `modules-subworkflows`, `stickers`) hit the GitHub API during the build. Add a `.env` at the repo root with a `GITHUB_TOKEN` (a personal access token with `public_repo` scope) to avoid rate limits.
```
mashehu, line 35: i don't think it can add a correct github_token automatically...

ewels: If it's an agent running locally it should be able to, using the gh CLI (`gh auth token` off the top of my head).

maxulysse:
We had a CLAUDE.md file originally in sarek, before we even thought about doing an AGENTS.md file, so we have this for now: https://github.com/nf-core/sarek/blob/dev/CLAUDE.md
I do not really recommend, as we don't really want to do that for every single agent, but for now it's good enough for us

maxulysse (approved):

ewels:
I have this in my `~/.claude/CLAUDE.md` file now, seems to work ok-ish:

```md
# Agents file

If no `CLAUDE.md` file is found in the root of the repo, check if there is a root-level
`AGENTS.md`. If there is, read it and treat it as if it were a `CLAUDE.md` file.
```

PR #4295, opened 2026-07-01
Fix newsletter month seeding so the current month's issue always appears

ewels:
The newsletter dated the 1st of month M reports the *previous* calendar
month's content, but getNewsletterMonths seeded candidate months from the
month each item was *dated in* rather than the month that *reports* it. A
newsletter therefore only appeared if something happened to be dated in its
own month, so the July issue (reporting June) would not show on the 1st
unless unrelated July-dated content existed.

Seed backward-looking content (blog posts, advisories, releases, new
pipelines) from month X into the newsletter dated X+1, and events into both
their own month and the previous month (matching the forward-looking
"upcoming events" section). newsletterMonthHasContent still gates out any
candidate whose sections would all be empty.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XUD4bQJx4nFhMGTtSqiwvK

### DIFF
--- a/sites/main-site/utils/newsletter.ts
+++ b/sites/main-site/utils/newsletter.ts
@@ -112,44 +112,41 @@ export function getNewsletterMonths(
     const monthSet = new Set<string>();
     const add = (year: number, month: number) => monthSet.add(`${year}-${month}`);
 
-    // A newsletter is *dated* the 1st of month M but reports the content it renders
-    // from other months, so we must seed the month that *reports* each item — not the
-    // month the item is dated in — or a newsletter can be missing its own content.
+    // A newsletter is *dated* the 1st of month M but reports content from adjacent months,
+    // so we seed the newsletter(s) that *report* each item rather than the month of the item.
     //
-    // Backward-looking content (blog posts, advisories, releases, new pipelines) from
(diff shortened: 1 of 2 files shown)


sites/main-site/utils/newsletter.ts
```diff
@@ -104,31 +110,46 @@ export function getNewsletterMonths(
...
+        const [{ year, month }] = adjacentMonths(date.getUTCFullYear(), date.getUTCMonth() + 1, 1, 1);
+        add(year, month);
+    };
+    // Events are forward-looking: an event in month X appears in the newsletters dated
+    // X (this month) and X-1 (next month's "upcoming events").
+    const seedFromEvent = (date: Date) => {
+        add(date.getUTCFullYear(), date.getUTCMonth() + 1);
+        const [{ year, month }] = adjacentMonths(date.getUTCFullYear(), date.getUTCMonth() + 1, 1, -1);
+        add(year, month);
+    };
```
mashehu: should be one function

mashehu (approved):
code comments could also be a bit less wordy

PR #4312, opened 2026-07-17
fix: make pagination reactive to lastPage changes

dialvarezs:
When searching modules and subworkflows, pagination is only computed on mount and when clicking its buttons.
This causes odd behaviour with the search field: the paginator keeps showing the pre-search page count (up to `maxPages`), and clicking a page higher than the filtered `lastPage` renders a blank listing.

This PR fixes it by making the page buttons reactive to `lastPage`, and clamps `currentPage` when `lastPage` shrinks.


mashehu (approved):
thanks for this fix!

PR #4342, opened 2026-08-01
Fix newsletter proposal HTML: duplicate pipelines and wrong accepted status

ewels:
Done on my phone in a campsite, so please forgive the AI blurb description. 'cc @jfy133 who reported.

---

Two bugs in the newsletter "New pipelines & proposals" section:

- A pipeline proposal was marked "accepted" whenever its issue was closed
  on GitHub with state_reason "completed". But nf-core/proposals closes
  rejected/withdrawn proposals as "completed" too, so e.g. dragen (#137,
  labelled "proposed") showed up as accepted. Acceptance is actually
  tracked with the "accepted" label, so key on that instead. A proposal
  closed without the label and not opened that month now drops out
  entirely, as intended.

- When a pipeline proposal is accepted, its repo is created the same
  month and already appears under "New pipeline repositories", so the
  pipeline was listed twice (e.g. pathogenepidemiology). Drop the
  proposal entry when a matching new repo is present for that month.

Update the getProposalsForMonth unit test to cover label-based
acceptance and the completed-but-not-accepted case.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01X2KaU4537UFKLW6Kw1rSSt

@netlify /blog/2026/tools-4_1_0


ewels:
@nf-core-bot fix linting

sites/main-site/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "extends": "astro/tsconfigs/base",
+    "extends": "../../node_modules/astro/tsconfigs/base.json",
```
ewels: @mashehu this looks a bit dodgy to me? Wanted to flag it. Otherwise the rest of the PR is fairly uncontroversial I think.

ewels: 🤖 explanation:

---

The Playwright CI job was failing before this — not on a test, but at config-load time:

```
Failed to resolve "extends" path "astro/tsconfigs/base"
referenced from sites/main-site/tsconfig.json
```

So `playwright test` died before running a single test. Cause: Playwright 1.62's tsconfig loader resolves `extends` through the package's `exports` map, and `astro`'s `package.json` doesn't expose the `./tsconfigs/base` subpath. `tsc`, `astro` and the editor use classic/their-own resolution and find the file fine, which is why only Playwright trips over it. It's made worse by npm workspaces hoisting `astro` up to the repo-root `node_modules`, so it isn't in `sites/main-site/node_modules` for a naive lookup either.

Pointing `extends` straight at the hoisted file sidesteps the exports map and resolves for every tool. It's ugly but deterministic (workspaces always hoist astro to root).

Cleaner alternatives if we'd rather not hardcode the path:

1. **Bump Playwright** to a version whose loader resolves the bare package `extends` again — smallest conceptual change, but a dependency bump to validate across the suite.
2. **Give Playwright its own tsconfig**: leave `tsconfig.json` extending `astro/tsconfigs/base` for astro/editors, and point Playwright at a tiny dedicated tsconfig that just carries the `@utils/*` etc. path aliases (no astro extends). Keeps the shared file untouched; adds one small config file.
3. **Skip the extends for Playwright entirely** — the config only needs the path aliases, not astro's compiler options.

jfy133 (approved):

sites/main-site/tsconfig.json
mashehu: I am not too sure about the relative path. Will need to double check that it doesn't break other builds.

ewels: Yup you're right, I think they'd be borked. Reverted and did option (2) instead in 771681b693dc279898a98585043a35c1d84a19b8

ewels: Ok should be good now I think @mashehu ?

sites/tsconfig.playwright.json
mashehu, line 1: why does playwright need compilerOptions with path aliases? that feels wrong? if it just needs to have some dummy tsconfig I would just make an empty one.

ewels: The test file `sites/main-site/tests/newsletter.spec.ts` does `import { … } from "@utils/newsletter"`, and without the paths entries in `tsconfig`, Playwright's TS resolution will fail with a module-not-found error before any test runs.

ewels: ok if I use relative paths in the newsletter test file then it can be empty. I thought that going back to relative paths like this would break stuff but Claude assures me that it should be fine. Will push now.

ewels: Done in 6edd141e0

ewels: Nope, it broke everything as I suspected. Reverting..

ewels:
Now hanging during playwright browser install for some reason 🙈

PR #4348, opened 2026-08-11
Update meta-omics metro map: add seqsubmit, drop proteinfold

vagkaratzas:
- New "4. DATA UPLOAD" column with nf-core/seqsubmit, fed from mag's fasta output alongside the other secondary-analysis branches
- Remove nf-core/proteinfold (not a meta-omics pipeline)
- Refresh the nf-core/proteinfamilies logo
- Convert the Inkscape flowRoot captions (fastq/fasta/profiles) to plain SVG text so they render outside Inkscape
- Drop the superseded 2024/2025 map files

@netlify /special-interest-groups/meta-omics


erikrikarddaniel (approved):
👍

PR #4352, opened 2026-08-14
Document GPU configuration flow across pipelines, modules, and profiles

mahesh-panchal:
## Summary
- Clarify that `accelerator` is a map (`request`/`type` keys), not a plain integer, and document how a GPU request flows: module -> pipeline config -> institutional profile -> user override.
- Add a "GPU resource requests" section to the institutional profiles guide, including a note on `--gres` vs `--gpus` on SLURM, and update the example `gpu` profile to read `task.accelerator` via a closure instead of hardcoding `--gres=gpu:1`.
- Add a new `running/configuration/gpu-pipelines.md` guide aimed at pipeline users.
- Remove the now-unused `ext.use_gpu` key from the `ext-args` reference.

@netlify /docs/running/configuration/gpu-pipelines


mahesh-panchal:
@nf-core-bot fix linting

sites/docs/src/content/docs/developing/components/ext-args.md
```diff
@@ -54,7 +54,6 @@ This is the list of permitted keys:
 | ext.prefix                            | File name prefix for output files.                                                           |
 | ext.prefix2                           | Second file name prefix for output files.                                                    |
 | ext.when                              | Boolean expression to determine when a module runs.                                          |
-| ext.use_gpu                           | Determines whether the module uses GPU settings.                                             |
```
pontus, line 57: I haven't done a survey, but since this is a list for allowed keys, I'm not comfortable simply removing it outright as it might make things technically non-compliant. My preference would be to mark as deprecated with a link to the pipeline notes (but then I don't have track of procedures this normally follows).

mahesh-panchal: There are 3 nf-core modules that still use it, and it looks like they should be updated to use task.accelerator. In those three modules it's only used in the container directive. edit: It's also used in the python script of one of the modules. 

The docs don't point to using it for something and it seems to have quietly been replaced by task.accelerator.

sites/docs/src/content/docs/developing/components/gpu-modules.md
```diff
@@ -24,6 +31,19 @@ A label-only alternative (e.g., requiring a `process_gpu` label) would not work
 Pipelines also set GPU container flags via `containerOptions` in their process config.
 Use `containerOptions` (not global `docker.runOptions`) to scope GPU flags to GPU processes only.
 
+## How a GPU request reaches the scheduler
```
mashehu: this section is useful, but feels a bit too in-depth for general docs imo. maybe add them inside a [collapsed admonition](https://nf-co.re/docs/developing/documentation/website-markdown#collapsing-admonitions)?

sites/docs/src/content/docs/developing/institutional-profiles/configuration.md
```diff
@@ -145,6 +145,43 @@ Other useful directives for the process scope:
...
+    }
+  }
+}
+```
+
+Using a closure that reads `task.accelerator{:groovy}` (rather than hardcoding a fixed `--gres=gpu:1`) keeps the request count under the pipeline/user's
+control and avoids reserving a GPU for jobs that never asked for one.
+Adjust the scheduler flag for your system: SLURM uses `--gres=gpu:N`, LSF uses `-gpu num=N`, and PBS Pro uses `-l select=1:ngpus=N`.
+
+:::note{title="SLURM: --gres vs --gpus"}
```
mashehu:
```suggestion
:::note{title="SLURM: --gres vs --gpus" collapse}
```

sites/docs/src/content/docs/developing/components/ext-args.md
mahesh-panchal: I've updated the remaining nf-core modules here: https://github.com/nf-core/modules/pull/12669

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+    --outdir results \
+    --remove_ribo_rna --ribo_removal_tool ribodetector --use_gpu_ribodetector \
+    -profile docker
+```
+
+Here, `--use_gpu_ribodetector` is enough on its own — `-profile docker` (or `singularity`) is the same profile you would use for a CPU-only run.
+The GPU container flags (`--gpus all` for Docker, `--nv` for Singularity/Apptainer) are applied automatically, and only to the GPU-enabled step.
+Other steps in the same run are unaffected and continue to run normally on CPU-only nodes, which matters on a shared or mixed cluster.
+
+Pipelines using this pattern often also expose an override parameter (for example, `--gpu_container_options`) so you can replace the auto-selected container flag if your site needs something different (a specific device index, a bind mount, etc.).
```
mashehu:
```suggestion
```
the different pipelines should document this. out of scope for the general docs imo

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+nextflow run nf-core/methylseq \
+    --aligner bwameth \
+    -profile gpu \
+    --input samplesheet.csv --genome GRCh38
+```
+
+:::caution{title="`-profile gpu` applies to the whole run"}
+In this pattern, the `gpu` profile might set the container engine's GPU flags globally (`docker.runOptions`, `singularity.runOptions`, etc.), so every process in the run gets them — not just the GPU-accelerated one - check the pipeline configuration.
+
+Avoid combining `-profile gpu` with `conda`/`mamba` (most GPU tools don't support it), and check the pipeline's usage docs for whether you also need to set the `accelerator` directive yourself in a custom config — some pipelines only set it automatically on cloud/Kubernetes executors (`awsbatch`, `google-batch`, `k8s`, `hq`), leaving it up to you elsewhere.
```
mashehu:
```suggestion
Check the pipeline's usage docs for whether you also need to set the `accelerator` directive yourself in a custom config — some pipelines only set it automatically on cloud/Kubernetes executors (`awsbatch`, `google-batch`, `k8s`, `hq`), leaving it up to you elsewhere.
```

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+
+The one place order still genuinely matters is a container engine's global `runOptions` (`docker.runOptions`, `singularity.runOptions`, etc.). Unlike `containerOptions` on a process, `runOptions` has no per-process scoping and can't be written as a closure on `task.accelerator` — it applies to every container the engine runs, full stop. That's also why the "applies to the whole run" caution above exists: older Pattern 2 pipelines like sarek set the GPU flag via `runOptions` rather than a process-scoped `containerOptions`. If your institutional profile also sets `runOptions` (for example, to add a bind mount), whichever profile is listed last fully replaces the other's value rather than combining with it — so you could silently lose either the GPU flag or the bind mount. This is specifically a `runOptions` problem: `containerOptions`, as used in the [institutional profile examples](../../developing/institutional-profiles/configuration#gpu-resource-requests), doesn't have it, since it's scoped per-process and only applies where you attach it. Check your institution's own `docs/<institution>.md` for the profile order it recommends for GPU runs.
+:::
+
```
mashehu: I feel like this is already in the general configs/institutional configs docs and is not so gpu specific.

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+```groovy title="custom-gpu.config"
+process {
+    withName: 'RIBODETECTOR' {
+        accelerator = [request: 2, type: 'a100']
+    }
+}
+```
+
+`accelerator{:groovy}` takes a map — `request` is the GPU count (`accelerator = 2{:groovy}` is shorthand for `accelerator = [request: 2]{:groovy}`), and `type` names a specific GPU model where your cluster or cloud executor supports it.
```
mashehu: duplication of previous infomartion
```suggestion
```

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+
+If your institution has a profile from [nf-core/configs](https://github.com/nf-core/configs), setting `accelerator` here is usually enough on its own: its `clusterOptions` is typically already written as a closure that reads `task.accelerator` and builds the scheduler flag for you (see [GPU resource requests](../../developing/institutional-profiles/configuration#gpu-resource-requests) for the pattern institutional profiles use, including why the flag is `--gpus`, not `--gres`, once a `type` is involved).
```
mashehu: again handled in other sections
```suggestion
```

mashehu (commented):
just some nitpicks, can be ignored if you disagree

sites/docs/src/content/docs/developing/components/gpu-modules.md
```diff
@@ -24,6 +31,19 @@ A label-only alternative (e.g., requiring a `process_gpu` label) would not work
...
+2. **The pipeline's `conf/base.config`** tags the process with `label 'process_gpu'{:groovy}` (or targets it with `withName`).
+3. **The pipeline's `conf/modules.config`** sets the default allocation, typically `accelerator = 1{:groovy}`, on that label or process name — often behind a pipeline parameter (for example, `accelerator = { params.use_gpu_ribodetector ? 1 : null }{:groovy}`) so users can opt in without needing a separate profile.
+4. **An institutional profile** (from [nf-core/configs](https://github.com/nf-core/configs)) intercepts `task.accelerator` inside a `clusterOptions` closure and translates it into the local scheduler's syntax, for example `--gres=gpu:${task.accelerator.request}{:groovy}` on SLURM. This is also where a site defines `task.accelerator.type{:groovy}` (which GPU model a request maps to on that cluster) and where `containerOptions` gets the runtime-specific flag (`--nv`, `--gpus all`) scoped to the same label. See [Writing process configuration](/docs/developing/institutional-profiles/configuration#process-scope) for the pattern institutional profiles use.
```
pontus:
```suggestion
4. **An institutional profile** (from [nf-core/configs](https://github.com/nf-core/configs)) intercepts `task.accelerator` and applies it to the resources requested - e.g. a `slurm` executor would typically use a `clusterOptions` closure to translate the request into the local scheduler's syntax (including considering `task.accelerator.type` if appropriate) and `containerOptions` to provide needed runtime options, whereas  for example a Kubernetes setup should use `accelerator` automatically and e.g. a cloud VM executor might use the machine type to supply the needed resources. See [Writing process configuration](/docs/developing/institutional-profiles/configuration#process-scope) for the pattern institutional profiles use.
```

sites/docs/src/content/docs/developing/institutional-profiles/configuration.md
```diff
@@ -145,6 +145,43 @@ Other useful directives for the process scope:
 - `beforeScript`: Load modules or set environment variables before job execution
 - `scratch`: Specify temporary directory for intermediate files
 
+### GPU resource requests
+
+nf-core pipelines tag GPU-capable processes with `label 'process_gpu'{:groovy}` and set Nextflow's
+[`accelerator{:groovy}` directive](https://docs.seqera.io/nextflow/reference/process#accelerator) when a GPU is
+requested (see [GPU-capable modules](/docs/developing/components/gpu-modules)).
+Your institutional profile only needs to translate that into your scheduler's own GPU request syntax, and scope the
+container runtime's GPU flag to the same processes:
```
pontus:
```suggestion
nf-core pipelines ensure GPU-capable processes have the [`accelerator{:groovy}` directive](https://docs.seqera.io/nextflow/reference/process#accelerator) set, (e.g. through `label 'process_gpu'{:groovy}` or directly using `withName` selectors) when
requested (see [GPU-capable modules](/docs/developing/components/gpu-modules)).
Your institutional profile only needs to translate that into your scheduler's own GPU request syntax, and scope the
container runtime's GPU flag to the same processes:
```

To my reading, it's easy to think `process_gpu` is magical with the current text, which it really isn't.

sites/docs/src/content/docs/developing/institutional-profiles/configuration.md
```diff
@@ -145,6 +145,43 @@ Other useful directives for the process scope:
...
+    clusterOptions = { task.accelerator ? "--gres=gpu:${task.accelerator.request}" : null }
+    containerOptions = {
+        if (!task.accelerator) { return null }
+        if (workflow.containerEngine == 'docker') { return '--gpus all' }
+        if (workflow.containerEngine in ['singularity', 'apptainer']) { return '--nv' }
+        return null
+    }
+  }
+}
+```
```
pontus, line 170: There may be cases where defining details for `process_gpu` is all that can be done, but I think they're rather few. Listing this here as such probably hurts more more than it helps to my mind (*much* better if the general process considers `task.accelerator` properly, e.g. like this but without the `process_gpu` label).

sites/docs/src/content/docs/running/configuration/gpu-pipelines.md
```diff
@@ -0,0 +1,121 @@
...
+title: Running GPU-accelerated pipelines
+subtitle: Enable and configure GPU acceleration for supported pipeline steps
+shortTitle: GPU pipelines
+---
+
+Some nf-core pipelines can run specific steps on a GPU instead of a CPU, for example,
+tools like NVIDIA Parabricks, GPU-accelerated basecallers, or machine-learning-based classifiers.
+GPU support is opt-in and pipeline-specific: there is no single flag that turns on GPU acceleration
+across all of nf-core. Always check the pipeline's own `docs/usage.md` for the exact parameter names
+before you run.
```
pontus, line 11:
```suggestion
before you run.

Currently, GPU acceleration typically assumes you have Nvidia/CUDA capable cards (e.g. AMD GPUs are unfortunately currently not supported).
```

sites/docs/src/content/docs/developing/components/gpu-modules.md
```diff
@@ -16,6 +16,13 @@ For the normative requirements, see the [resource requirements](/docs/specificat
 Modules detect whether a GPU has been requested with `task.accelerator{:groovy}`.
 A module never sets the `accelerator` directive itself; the pipeline controls GPU allocation by setting `accelerator = 1{:groovy}` in its process config (for example, via a `process_gpu` label or a `withName` block).
 
+:::note{title="accelerator is a map"}
+The [`accelerator{:groovy}` directive](https://docs.seqera.io/nextflow/reference/process#accelerator) is a map with `request` and `type` keys.
+`accelerator = 1{:groovy}` is shorthand for `accelerator = [request: 1]{:groovy}`.
+Modules only ever need `task.accelerator.request{:groovy}` — the GPU count, the same role `task.cpus{:groovy}` plays for cores.
```
CormacKinsella:
```suggestion
Modules only ever need `task.accelerator.request{:groovy}` (the GPU count). In addition to its role in detecting GPU requests, it can serve the same role that `task.cpus{:groovy}` does for CPU cores.
```

sites/docs/src/content/docs/developing/components/gpu-modules.md
```diff
@@ -16,6 +16,13 @@ For the normative requirements, see the [resource requirements](/docs/specificat
...
+`task.accelerator.type{:groovy}` (the GPU model, e.g. `'nvidia-tesla-a100'`) is infrastructure-specific: it is set by an institutional profile or a user's local config, never hardcoded by a pipeline or module.
+:::
+
 :::info{title="Why detection lives in the module, not a label"}
 Placing GPU allocation in the pipeline config lets users control it through their pipeline config or profiles.
 A label-only alternative (e.g., requiring a `process_gpu` label) would not work for modules that support both CPU and GPU modes, such as [`ribodetector`](https://github.com/nf-core/modules/tree/master/modules/nf-core/ribodetector), so the choice is left to the pipeline author.
```
CormacKinsella: I wonder if this text could be improved, it reads as misleading as it stands. Possibly also needs a PR for the (I think 5) modules that are CPU + GPU capable.

My reading of it is it _would work_ to use a label, but it's currently just a decision to have these modules default to CPU, even when the `gpu` profile is used. GPU opt-in is required right now.

In my view it would not be unreasonable to default to GPU with the `gpu` profile and CPU without. In a mixed pipeline developers would just need to opt-out for specific processes: `task.accelerator` `null` using `withName`.

sites/docs/src/content/docs/developing/components/ext-args.md
mahesh-panchal: Would you prefer this is removed from this PR and made in a separate PR? The remaining uses in nf-core modules are now gone.

sites/docs/src/content/docs/developing/components/gpu-modules.md
mahesh-panchal: I thought this was ok, but admittedly it took me a bit to wrap my head around what it meant, so adjusting the wording is probably a good call.

mahesh-panchal:
@nf-core-bot fix linting

pontus (approved):
Great work, thanks!

sites/docs/src/content/docs/developing/components/gpu-modules.md
```diff
@@ -16,14 +16,39 @@ For the normative requirements, see the [resource requirements](/docs/specificat
...
 Pipelines also set GPU container flags via `containerOptions` in their process config.
 Use `containerOptions` (not global `docker.runOptions`) to scope GPU flags to GPU processes only.
 
+:::note{title="How a GPU request reaches the scheduler" collapse}
+
+A module only ever reads `task.accelerator`; it never decides when a GPU is actually available.
+That decision flows through several configuration layers before it reaches the module:
+
+1. **The module** checks `task.accelerator` to pick a container/binary and reads `task.accelerator.request` for the GPU count (this page).
+2. **The pipeline's `conf/base.config`** assigns a global default to the processes with `label 'process_gpu'{:groovy}`.
```
CormacKinsella:
```suggestion
2. **The pipeline's `conf/base.config`** assigns a global default to the processes with `label 'process_gpu'{:groovy}` (requires use of the `gpu` profile).
```

CormacKinsella (approved):

PR #4359, opened 2026-08-19
Add Cloudflare Turnstile to the newsletter sign-up form

ewels:
The newsletter sign-up endpoint has been used as a spam relay: attackers submit harvested third-party addresses at roughly 300/day and the backend emails each victim a confirmation request, generating ISP spam complaints against nf-co.re's sending reputation. A CAPTCHA at submit time is the only control that engages, because rate limiting cannot distinguish the attacker (who runs slowly, from many IPs, with a fresh address each time) from a real visitor.

This adds a Cloudflare Turnstile widget to the sign-up form. The matching server-side verification is in nf-core/newsletter.

## What changes

Only `sites/main-site/src/components/newsletter/NewsletterSignup.astro`. That single component backs all three entry points — the homepage, `/newsletter`, and `/newsletter/subscribe` — and nothing else in the repo posts to the subscribe endpoint, so there was no second place to patch.

The widget renders explicitly once Cloudflare's script loads, the submit button stays disabled until the widget produces a token, and the token is sent as `cf-turnstile-response` in the same JSON body as the email address. A `400 {"error": "captcha_failed"}` from the backend shows a retry message and calls `turnstile.reset()`, since Turnstile tokens are single-use and a spent one cannot be resubmitted.

No new dependency — this is a script tag and a form field.

## Configuration

The site key is public and read from `PUBLIC_TURNSTILE_SITE_KEY`, matching the `envPrefix: ["PUBLIC_"]` convention already in `astro.config.mjs`. It needs setting as a build-scoped environment variable on the `nf-core-main-site` Netlify project; Astro inlines it at build time, so a redeploy is required after any change to the value. The secret key does not belong in this repo — it lives in AWS SSM, since anything `PUBLIC_`-prefixed is served to every visitor.

Deploy previews need `netlify.app` added to the widget's hostname list in Cloudflare, otherwise verification fails on preview URLs.

## Safe to merge before the backend enforces

With `PUBLIC_TURNSTILE_SITE_KEY` unset the form renders and behaves exactly as it does today — no script tag, no container, no disabled button. Verified by running a full production build twice, once with the key unset and once with a Cloudflare test key, and inspecting the generated HTML in both cases.

The backend only enforces once its own secret is configured, so merging this first is the correct order: the form starts sending a token the backend initially ignores, and enforcement is switched on afterwards. Doing it the other way round would reject every genuine sign-up.

## Testing note

Playwright could not run in this environment at all — its config fails to resolve `astro/tsconfigs/base`, reproduced on the unmodified codebase and on both Node 22 and 26, so it is pre-existing and unrelated. Verification here was the two full builds described above rather than an end-to-end run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JSUGrKz9VzGoe4S1Hr9soj

### DIFF
--- a/sites/main-site/src/components/newsletter/NewsletterSignup.astro
+++ b/sites/main-site/src/components/newsletter/NewsletterSignup.astro
@@ -22,6 +22,13 @@ const {
 // Live newsletter service /subscribe endpoint.
 // TODO: move behind a stable custom domain (e.g. api.nf-co.re) when one exists.
 const SUBSCRIBE_URL = "https://e5veor8z45.execute-api.eu-west-1.amazonaws.com/subscribe";
+
+// Cloudflare Turnstile CAPTCHA, to stop the endpoint being used as a spam relay
+// (harvested addresses submitted here get a confirmation email from us).
+// The site key is public by design. Until it's set (local dev, early preview
+// deploys) the form degrades gracefully to its pre-Turnstile behaviour — the
+// backend only enforces the check once its matching secret key is configured.
+const TURNSTILE_SITE_KEY = import.meta.env.PUBLIC_TURNSTILE_SITE_KEY;
 ---
 
 <div class="nl-signup">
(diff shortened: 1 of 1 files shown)


sites/main-site/src/components/newsletter/NewsletterSignup.astro
```diff
@@ -44,16 +51,70 @@ const SUBSCRIBE_URL = "https://e5veor8z45.execute-api.eu-west-1.amazonaws.com/su
                 autocomplete="email"
                 required
             />
-            <button class="btn btn-success" type="submit">
+            <button class="btn btn-success" type="submit" disabled={!!TURNSTILE_SITE_KEY}>
```
mashehu, line 54: did you really want `!!` here?

mashehu: ah, figured out what it does now. javascript is weird.

PR #4371, opened 2026-08-25
Add bytesize event for nf-core/nft-utils

maxulysse:
Louis Le Nezet will present nf-core/nft-utils v1.1.1 on Wednesday, September 2nd.

@netlify /events/2026/bytesize_nft-utils


FranBonath (approved):
It will need to be very clear that the bytesize changed week day.

PR #4382, opened 2026-08-31
Document random seed input requirements

itrujnara:
Added section on random seed input handling for modules, as discussed at the August 2026 maintainers' meeting and [on Slack](https://nfcore.slack.com/archives/C043UU89KKQ/p1787926629794649).

@netlify /docs/specifications/components/modules/input-output-options


sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,17 @@ When one and only one of multiple argument are required:
 
   :::
 
+### Random seed
+
+If the main underlying tool accepts a random seed as an argument, the module SHOULD define an optional `val` input for the random seed.
+
+- If the user provides a value, it MUST be appropriately provided to the tool.
+- If the user provides no value (`[]`), the module MUST NOT pass any seed argument to the tool.
```
mahesh-panchal, line 58: I would add here that pipeline users can still supply random seed values via `ext.args` if necessary overriding the default pipeline behaviour.

mahesh-panchal (approved):
I'm happy with this, but do get more input.

SPPearce (approved):

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
LouisLeNezet: Good question, maybe an error should be given if user pass seed through ext.args and val(seed) ?

LouisLeNezet: With an example like
```
if (seed && args.contains("--seed")) {
    error("Seed should be passed either through external arguments or through seed input channel")
}
```

mahesh-panchal: I would say it's the users preference that should take precedence. I can't think of a reason a pipeline developer should maintain control of the seed.

itrujnara: I have updated with the user-first version, let me know if you prefer the other approach

sites/docs/src/content/docs/specifications/components/modules/input-output-options.md
```diff
@@ -50,6 +50,22 @@ When one and only one of multiple argument are required:
...
+- If a random seed is provided through both the input and `ext.args`, the module SHOULD use the value from `ext.args`.
+
+:::info{title="Tip"}
+Since most tools apply duplicate arguments in the order they are provided, you can usually achieve this by passing `$args` after the input-based seed argument.
```
mahesh-panchal: I'm not sure if this is true. I think the check should be in the groovy code. 
```
def seed_flag = random_seed ? "--seed ${random_seed}": ""
if (task.ext.args.contains("--seed")){
  // extract the value
  seed_flag = "..."
}
```

pinin4fjords (approved):

ewels:
@nf-core-bot fix linting

PR #4385, opened 2026-09-02
Fix relative links between pipeline docs pages

### LINKED ISSUE #4384: Relative links between a pipeline's docs pages 404 on the site
**Describe the bug**

Relative links between a pipeline's own docs pages (the `[text](usage.md#anchor)` form that renders correctly on GitHub) resolve to a 404 on the website. Docs pages are served, and linked from the site's own navigation, at a URL ending in a slash, but a cross-file link is rewritten to a still-relative href, so it resolves under the current page's directory instead of beside it.

**To Reproduce**

1. Go to https://nf-co.re/mag/dev/docs/output/
2. Find "see the [note on long read filtering](usage.md#a-note-on-long-read-filtering)" (source: [`docs/output.md#L143`](https://github.com/nf-core/mag/blob/dev/docs/output.md?plain=1#L143))
3. The rendered href is `usage#a-note-on-long-read-filtering`, relative
4. Clicking it goes to https://nf-co.re/mag/dev/docs/output/usage → **404**

The correct page is https://nf-co.re/mag/dev/docs/usage (200). The site's own sidebar links to it as `/mag/dev/docs/usage/`, with the trailing slash, which is what makes the relative href resolve one level too deep. The same happens on the release channel: `/<pipeline>/<version>/docs/output/` is also slash-terminated, so `.../2.1.0/docs/output/usage` 404s too.

This is invisible in PR review, because GitHub renders the same source link correctly.

**Expected behavior**

A cross-file docs link resolves to the sibling page in the same pipeline **and the same version** — from `/mag/dev/docs/output/` to `/mag/dev/docs/usage/`, and from `/mag/4.0.0/docs/output/` to `/mag/4.0.0/docs/usage/`.

The version-preserving part is the reason this is worth fixing rather than working around. The workaround available to a pipeline today is an absolute `https://nf-co.re/<pipeline>/usage#anchor` link — which is what [nf-core/rnaseq](https://github.com/nf-core/rnaseq/blob/dev/docs/output.md) does, and what we have just switched [nf-core/phyloplace#79](https://github.com/nf-core/phyloplace/pull/79) to — but that always lands on the released docs, so a reader browsing `/dev/` is silently taken out of the version they were reading, and an anchor for an unreleased section does not resolve at all until the next release. Our reviewer raised exactly that objection, and he was right; there is just no relative form that currently works.

**Additional context**

Two plugins rewrite these links, and neither produces an absolute path.

Anchored links go through `link()` in [`bin/satteri/mdast-github-markdown.ts`](https://github.com/nf-core/website/blob/main/bin/satteri/mdast-github-markdown.ts):

```ts
const MDX_ANCHOR_REGEX = /\.mdx?#/;
...
} else if (MDX_ANCHOR_REGEX.test(node.url)) {
    ctx.setProperty(node, "url", node.url.replace(MDX_ANCHOR_REGEX, "#"));
}
```

`usage.md#a-note-on-long-read-filtering` becomes `usage#a-note-on-long-read-filtering` — which is the href on the live page, so this is what is deployed.

Links with no anchor go through `bin/satteri/hast-markdown-links.ts` instead, which turns `usage.md` into `usage/`. Also relative, so also 404.

**Proposed fix**

Emit a root-relative URL — `/{pipeline}/{version}/docs/{path}/#{anchor}` — resolved against the directory of the page being rendered. Everything needed is already in scope in the mdast plugin: `createGitHubMarkdownPlugin({ repo, ref, parentDirectory })` is constructed per page inside `renderPipelineMarkdown()`, from a route that has `pipeline`, `version` and the doc file's `parentDir`. So this is a change to the existing `link()` visitor rather than new plumbing, and it wants:

- a match on the path portion (`/\.mdx?($|#|\?)/`) so anchored and plain links take the same branch, instead of one branch here and one in the hast plugin;
- resolution of the link against `parentDirectory`, so a `../` in a link from a docs subdirectory keeps working;
- a new option for the URL segment rather than reuse of `ref`. `renderPipelineMarkdown` is passed `version_tag_name`, so on `/{pipeline}/latest/docs/...` a link built from `ref` would move the reader off the `latest` alias onto a pinned version. `Astro.params.version` is the right value and would need threading through.

I would like to have a go at this — could you assign it to me?

**Browser**

- OS: Linux
- Browser: not browser-specific; reproduced with `curl` against the rendered HTML

_(Investigated with the help of Claude Code.)_

ewels:
Closes #4384.

Relative links between a pipeline's own docs pages (`[text](usage.md#anchor)`, which renders correctly on GitHub) 404 on the website. Docs pages are served at trailing-slash URLs, so a still-relative href resolves one level too deep: on `/mag/dev/docs/output/`, `usage#a-note-on-long-read-filtering` goes to `/mag/dev/docs/output/usage` instead of `/mag/dev/docs/usage/`. It's invisible in PR review because GitHub renders the same source link fine.

## History

We used to handle this, and the handling was dropped by accident.

`6f3caa3a4` (June 2023, "fix relative links in pipeline docs") added a regex pass over the raw markdown in the docs page component, before render:

```js
// make relative links to other markdown files absolute to current Astro.url.pathname
content = content.replaceAll(/\[([^\]]+)\]\((?!http)(?!#)(.*?)\)/g, (match, p1, p2) => {
    const link = [...new Set([...url.split('/').filter(i => i), ...p2.split('/').filter(i => i)])].join('/');
    return `[${p1}](/${link})`;
});
```

`96d2e1903` ("move cached markdown files into content collections on build") then replaced the fetch-and-string-munge step with `entry.render()`, and this pass was not carried over into the remark plugins that took over. Everything since — `remark-github-markdown.js`, the `rehype-urls` `urlTransformer`, and today's satteri plugins — only strips the `.md` extension. That leaves the href relative, which is the bug. So this is a restoration rather than a new feature.

## What this does

The rewrite goes back in as part of `createGitHubMarkdownPlugin` in `bin/satteri/mdast-github-markdown.ts`, where the anchored-link branch already lived, so anchored and plain links take the same path instead of one branch here and one in `hast-markdown-links.ts`:

```ts
function markdownLinkUrl(url: string): string {
    // Resolve "../" etc. against the directory of the file being rendered.
    const resolved = new URL(url, `https://n/${parentDirectory ? parentDirectory + "/" : ""}`);
    const path = resolved.pathname.slice(1);
    const page = path.replace(/\.mdx?$/, "");
    if (page === "README") return `/${repo}/${version}/${resolved.hash}`;
    if (!page.startsWith("docs/")) return `${baseRepoUrl}${path}${resolved.search}`;
    return `/${repo}/${version}/${page}/${resolved.search}${resolved.hash}`;
}
```

Three things worth calling out:

**Version, not ref.** `version` is a new option, threaded through `renderPipelineMarkdown()` from `Astro.params.version`, and it is deliberately not `ref`. `renderPipelineMarkdown` is passed `version_tag_name`, so a URL built from `ref` would move a reader on `/mag/latest/docs/` onto a pinned version. Preserving the channel is the reason to fix this properly rather than tell pipelines to hardcode `https://nf-co.re/<pipeline>/usage#anchor` — that workaround always lands on the released docs, so a reader browsing `/dev/` is silently taken out of the version they were reading, and an anchor for an unreleased section doesn't resolve at all until the next release.

**`new URL` for resolution**, so a `../` in a link from a docs subdirectory keeps working. The 2023 version joined deduplicated path segments, which mangled repeated segments and did not understand `../`.

**Markdown with no page on the site goes to GitHub.** Previously only an exact `CHANGELOG.md` / `CITATIONS.md` / `.github/CONTRIBUTING.md` matched, so `../CHANGELOG.md` from a docs page fell through to a relative 404. Now the path is resolved first, so the prefixed forms are caught too.

Also deleted the dead `url` variable at the top of the docs route — the last leftover of the 2023 version, unused since the pass it fed was removed.

## Testing

`npm run test-unit` (new root script, plain `node --test`) runs `bin/satteri/mdast-github-markdown.test.ts`: sibling links, anchored links, `../` from a docs subdirectory, README links from the pipeline index, non-docs markdown, and absolute/anchor-only links left alone. There was no unit test setup in the repo before; the existing Playwright suites need a deployed preview, and CI only runs them for `sites/main-site` changes, so a link-rewriting regression like this one had nothing to catch it.

Rendering `see the [note](usage.md#a-note-on-long-read-filtering)` through the real satteri processor (same plugin sets as `renderPipelineMarkdown`, `parentDirectory: "docs"`) gives:

```html
<p>see the <a href="/mag/dev/docs/usage/#a-note-on-long-read-filtering">note</a> ...
```

**This is not visible on the deploy preview.** `sites/main-site/astro.config.mjs` proxies `/:pipeline/:version/*` to `https://nf-core-pipelines.netlify.app` — the hardcoded production host, not a preview one — and the pipelines sub-site does not publish a deploy preview of its own. So `deploy-preview-4385--nf-core-main-site.netlify.app/mag/dev/docs/output/` serves the live pipelines site and is byte-identical to nf-co.re, whatever this branch does. Any change to a pipeline page is unpreviewable on a PR today, which seems worth fixing separately.

@netlify /[pipeline]/[version]

### DIFF
--- a/bin/satteri/mdast-github-markdown.test.ts
+++ b/bin/satteri/mdast-github-markdown.test.ts
@@ -26,6 +26,15 @@ test("markdown links point at the same pipeline and version", () => {
 
 test("files without a page on the website still go to GitHub", () => {
     assert.equal(rewriteLink("../CHANGELOG.md", "docs"), "https://github.com/nf-core/mag/blob/2.5.4/CHANGELOG.md");
+    // anchors survive: the reader lands on the release section, not the top of the changelog
+    assert.equal(
+        rewriteLink("../CHANGELOG.md#v250", "docs"),
+        "https://github.com/nf-core/mag/blob/2.5.4/CHANGELOG.md#v250",
+    );
+    assert.equal(
+        rewriteLink("CITATIONS.md#bowtie2", ""),
+        "https://github.com/nf-core/mag/blob/2.5.4/CITATIONS.md#bowtie2",
+    );
     assert.equal(rewriteLink("CITATIONS.md", ""), "https://github.com/nf-core/mag/blob/2.5.4/CITATIONS.md");
     assert.equal(rewriteLink("../assets/foo.csv", "docs"), "https://github.com/nf-core/mag/blob/2.5.4/assets/foo.csv");
 });
--- a/bin/satteri/mdast-github-markdown.test.ts
+++ b/bin/satteri/mdast-github-markdown.test.ts
@@ -1,0 +1,36 @@
+// Link rewriting checks for createGitHubMarkdownPlugin: run with `npm run test-unit`.
+// Pipeline docs pages are served at trailing-slash URLs, so a relative href to a
+// sibling doc resolves one level too deep and 404s — see nf-core/website#4384.
+import assert from "node:assert/strict";
+import test from "node:test";
+import { createGitHubMarkdownPlugin } from "./mdast-github-markdown.ts";
+
+const ctx = { setProperty: (node: any, key: string, value: any) => (node[key] = value) };
+
+function rewriteLink(url: string, parentDirectory: string): string {
+    const plugin: any = createGitHubMarkdownPlugin({ repo: "mag", ref: "2.5.4", version: "dev", parentDirectory });
+    const node: any = { url };
+    plugin.link(node, ctx);
+    return node.url;
+}
+
+test("markdown links point at the same pipeline and version", () => {
+    assert.equal(rewriteLink("usage.md#a-note", "docs"), "/mag/dev/docs/usage/#a-note");
+    assert.equal(rewriteLink("usage.md", "docs"), "/mag/dev/docs/usage/");
+    assert.equal(rewriteLink("../usage.md", "docs/subdir"), "/mag/dev/docs/usage/");
+    assert.equal(rewriteLink("subdir/page.mdx", "docs"), "/mag/dev/docs/subdir/page/");
+    // README (parentDirectory "") linking into the docs pages
+    assert.equal(rewriteLink("docs/usage.md", ""), "/mag/dev/docs/usage/");
+    assert.equal(rewriteLink("../README.md", "docs"), "/mag/dev/");
+});
+
+test("files without a page on the website still go to GitHub", () => {
+    assert.equal(rewriteLink("../CHANGELOG.md", "docs"), "https://github.com/nf-core/mag/blob/2.5.4/CHANGELOG.md");
+    assert.equal(rewriteLink("CITATIONS.md", ""), "https://github.com/nf-core/mag/blob/2.5.4/CITATIONS.md");
+    assert.equal(rewriteLink("../assets/foo.csv", "docs"), "https://github.com/nf-core/mag/blob/2.5.4/assets/foo.csv");
+});
+
+test("absolute and anchor-only links are left alone", () => {
+    assert.equal(rewriteLink("https://nf-co.re/usage.md", "docs"), "https://nf-co.re/usage.md");
+    assert.equal(rewriteLink("#anchor-only", "docs"), "#anchor-only");
+});
--- a/bin/satteri/mdast-github-markdown.ts
+++ b/bin/satteri/mdast-github-markdown.ts
@@ -15,7 +15,7 @@
 import { defineMdastPlugin } from "satteri";
 
 const SPECIAL_FILES_REGEX = /^(\.github\/CONTRIBUTING\.md|CITATIONS\.md|CHANGELOG\.md|\.config)$/;
-const MDX_ANCHOR_REGEX = /\.mdx?#/;
+const MD_LINK_REGEX = /\.mdx?($|#|\?)/;
 const ADMONITION_REGEX = /^\[!(NOTE|TIP|IMPORTANT|WARNING|CAUTION)\]\s*(.*)/;
 const IMG_SRC_REGEX = /<img(.*?)src="(.*?)"/g;
 const SOURCE_SRCSET_REGEX = /<source(.*?)srcset="(.*?)"/g;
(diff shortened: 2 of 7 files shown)


bin/satteri/mdast-github-markdown.ts
```diff
@@ -74,14 +74,37 @@ export interface GitHubMarkdownOptions {
...
+     *
+     * Markdown files outside docs/ have no page on the website, so they go to GitHub.
+     */
+    function markdownLinkUrl(url: string): string {
+        // Resolve "../" etc. against the directory of the file being rendered.
+        const resolved = new URL(url, `https://n/${parentDirectory ? parentDirectory + "/" : ""}`);
+        const path = resolved.pathname.slice(1);
+        const page = path.replace(/\.mdx?$/, "");
+        if (page === "README") return `/${repo}/${version}/${resolved.hash}`;
+        if (!page.startsWith("docs/")) return `${baseRepoUrl}${path}${resolved.search}`;
```
erikrikarddaniel: `resolved.search` is carried through here but `resolved.hash` is not, so an anchor into a markdown file outside `docs/` is silently dropped:

- `../CHANGELOG.md#v250` → `https://github.com/nf-core/mag/blob/2.5.4/CHANGELOG.md`
- `CITATIONS.md#bowtie2` → `https://github.com/nf-core/mag/blob/2.5.4/CITATIONS.md`

The reader lands at the top of a long changelog instead of the release section. `SPECIAL_FILES_REGEX` is `^…$`-anchored, so as soon as such a link carries an anchor or a `../` it falls through to this branch rather than being caught earlier.

Not a regression — the old `.mdx?#` → `#` rewrite left these relative and broken too — but cheap to carry over while you are in here.

```suggestion
        if (!page.startsWith("docs/")) return `${baseRepoUrl}${path}${resolved.search}${resolved.hash}`;
```

(`page === "README"` just above drops `resolved.search` for the same reason. Much harder to hit, so only worth it if you are touching the line anyway.)

package.json
```diff
@@ -9,7 +9,8 @@
         "build-component-json": "node bin/components.json.js",
         "check-api-catalog": "node bin/check-api-catalog.js",
         "update": "npx npm-check-updates --interactive --format group --workspaces",
-        "test-all": "for dir in sites/*; do (cd $dir && npm run test); done"
+        "test-all": "for dir in sites/*; do (cd $dir && npm run test); done",
+        "test-unit": "node --test bin/satteri/*.test.ts"
```
erikrikarddaniel, line 13: Nothing invokes this. I checked all 12 files under `.github/workflows/` at this ref and none references `test-unit`, so the suite will not run on PRs.

Worth closing, because the rationale for adding it — that a link-rewriting regression like this one had nothing to catch it — still holds while the script is manual-only. `linting.yml` already provisions Node 24, which is new enough to run `node --test` against `.ts` directly, so it should be a couple of lines there.

erikrikarddaniel (approved):
So, here's my Claude's small comments. It's overall verdict: "It's a real fix, not a drive-by close. And it's better than what we proposed".
I note that there's a test failing though.

bin/satteri/mdast-github-markdown.ts
ewels: Fixed in 0e8abc2c1 — took the suggestion, and carried `resolved.search` through the `README` branch on the line above for the same reason. Two assertions added for the anchored cases (`../CHANGELOG.md#v250`, `CITATIONS.md#bowtie2`).

package.json
ewels: Right, it was manual-only. Added a `unit` job to `linting.yml` in 0e8abc2c1: Node 24, `npm ci`, `npm run test-unit`. It has `permissions: contents: read` and `persist-credentials: false`, which the other jobs in that file predate — worth doing to them too, but not in this PR.

bin/satteri/mdast-github-markdown.test.ts
ewels, line 1: New convention for tests. @mashehu shout if you don't like this and we can move / delete. For now it seems like it doesn't hurt and is additive so will leave it.

PR #4386, opened 2026-09-02
Fix broken links for 'Edit' button in page header

ewels:
The "Edit" button in page headers goes to a 404 on GitHub from many pages. Enough that I no longer bother trying it for any pages. This is sad as really like this feature.

I set Claude off on a hunt to try to track down anywhere it was broken and fix it.

### DIFF
--- a/sites/docs/src/pages/docs/[...doc].astro
+++ b/sites/docs/src/pages/docs/[...doc].astro
@@ -156,11 +156,7 @@ if (currentGroup.length > 0) {
 }
 
 const docType = Astro.props.doc.data.type;
-const isMDX = Astro.props.doc.filePath?.endsWith(".mdx");
-const md_github_url =
-    "https://github.com/nf-core/website/blob/main/sites/docs/src/content/docs/" +
-    Astro.props.doc.id +
-    (isMDX ? ".mdx" : ".md");
+const md_github_url = "https://github.com/nf-core/website/blob/main/sites/docs/" + Astro.props.doc.filePath;
 // remove leading and trailing slash
 const splitUrl = `${Astro.url.pathname}/`.replace(/^\/|\/$/g, "").split("/");
 // check if markdown file starts with a heading (markdown or html)
--- a/sites/docs/src/pages/docs/nf-core-tools/api_reference/[version]/[...slug].astro
+++ b/sites/docs/src/pages/docs/nf-core-tools/api_reference/[version]/[...slug].astro
@@ -85,21 +85,24 @@ headings.map((heading) => {
     heading.text = heading.text.replaceAll(/\./g, ".\u200B");
 });
 
-// try to figure out github URL to the source file from file path
-let md_github_url = "";
-const base_url = `https://github.com/nf-core/tools/blob/${version_tag_name}/nf_core/`;
-const file_name =
-    entry.filePath?.replace("src/content/api_reference/", "").split("/").pop()?.replace(".md", ".py") || "";
-// check if entry.id contains "pipeline_lint_tests"
-if (entry.id.includes("pipeline_lint_tests")) {
-    md_github_url = base_url + "lint/" + file_name;
-} else if (entry.id.includes("module_lint_tests") || entry.id.includes("subworkflow_lint_tests")) {
-    // append to base url either to modules/lint or subworkflows/lint
-    const component_type = entry.id.includes("module_lint_tests") ? "modules" : "subworkflows";
-    md_github_url = base_url + component_type + "/lint/" + file_name;
-} else if (entry.id.includes("api")) {
-    md_github_url = base_url + file_name;
-}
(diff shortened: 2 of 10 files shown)


ewels:
> [!NOTE]
> AI generated comment:


## Validation

Every link this PR touches was checked mechanically rather than spot-checked by eye. Method and numbers below, including what is still broken and why.

### Website source links (commit 1)

Built `main-site` and `docs`, extracted every generated `github.com/nf-core/website/blob/main/...` URL, and checked each against the file on disk. Exact check, not sampled.

| | count | broken before | broken after |
|---|---|---|---|
| main-site | 552 | most (see below) | 0 |
| docs | 195 | ~20 | 0 |

The before-state per page type: all blog posts (63) and special interest groups (17) had no file extension appended; every about page had both a wrong path prefix and no extension; all 128 hackathon projects appended `.mdx` to files that are `.md`; 74 `.mdx` events had `.md` appended. In docs, `nf-core-tools/CLI/*` was lowercased to `cli/*` by slugification and 404s on case-sensitive GitHub. Advisories worked by luck (all `.md`).

### Pipeline pages (commit 2)

These routes are SSR, so there is no build output to grep. Instead I reproduced the template's URL expression against `pipelines.json` with the same inputs, generating the 514 URLs for `/latest/` pages, and sampled 40 across pipelines and doc files.

Before: `40/40` returned 404. After: `40/40` returned 200.

```
404  https://github.com/nf-core/rnaseq/blob/latest/README.md
200  https://github.com/nf-core/rnaseq/blob/3.26.0/README.md
```

`version_tag_name` was already being computed a few lines above in both files for fetching the content; the edit URL just was not using it. The docs page additionally now uses the real filename from `release.doc_files` rather than re-appending `.md`, which removes a latent break the slug logic already anticipates. No pipeline currently ships `.mdx` docs, so that part is not a live fix.

### API reference source links (commit 3)

Fetched the git tree of nf-core/tools for all 51 referenced tags via the API and checked all 2170 generated links against the real file lists. Exact check.

| | before | after |
|---|---|---|
| broken source links | 1239 (57%) | 191 (9%) |

Also removed: 168 links to `index.py` files that have never existed at any tag, since section index pages have no python source. The button is now hidden on those.

### API reference internal links (commit 3)

The version dropdown swapped the version into the current path without checking the target version has that page. Confirmed against the live site:

```
404  https://nf-co.re/docs/nf-core-tools/api_reference/1.10.1/api/pipelines/create
404  https://nf-co.re/docs/nf-core-tools/api_reference/2.14.1/api/pipelines/create
```

Checked every internal `api_reference` href in the built output against the built pages:

| | unique links | broken |
|---|---|---|
| dropdown, before | 6882 | 4296 |
| sidebar, before | 2296 | 0 |
| all, after | 2452 | 0 |

Sidebar links were not 404s before, so they do not appear in that table. They were pinned: all 58 on the live `/latest/` page return 200, but they point at `/4.1.0/`, so one click drops you out of `latest`. Verified after the fix that all 58 stay on `/latest/`, and that `aria-current="page"` marks exactly one entry. The old `isCurrent` used a substring match, so `lint` was highlighted while reading `lint_utils`.

### What is still broken, and why

The 191 remaining API reference source links cannot be resolved from this repo. Two causes:

**Packages, 125 links.** A page documenting a package needs `__init__.py`, not `.py`. Nothing in the generated docs distinguishes a package from a module, and it is not stable per name either, so it cannot be hardcoded: `pipelines/download` is a module at 3.0.0 and a package at 4.1.0.

| path | links |
|---|---|
| `nf_core/lint.py` | 31 |
| `nf_core/modules.py` | 31 |
| `nf_core/pipelines/create.py` | 20 |
| `nf_core/pipelines/lint.py` | 20 |
| `nf_core/subworkflows.py` | 14 |
| `nf_core/pipelines/download.py` | 9 |

**Stale docs, 66 links.** The doc page exists but its source was renamed or removed upstream at that tag: `base_config` and `modules_config` folded into `configs.py` (44), `pipelines/utils.py` removed (19), `nf_core_yml.md` is a hand-written config reference rather than a module doc (2), `actions_ci` removed (1).

Both classes are 404s today as well, so nothing regresses. There is a comment in the code noting the package limitation.

### Approach that was tried and rejected

Deriving the module path from each page's `nf_core.x.y` heading instead of its directory looks more authoritative but measures worse: 208 broken versus 191. The headings are stale at 3.0.x, where they say `nf_core.bump_version` while the source had already moved to `nf_core/pipelines/`. They are also mutated in place by the zero-width-space pass above, and Astro reuses the `headings` object across renders of the same entry, so U+200B leaks into the URL on `/latest/` pages.

### Not covered

`configs`, `nft-utils` and `contributors.yaml` were reviewed and are correct, so they are untouched. The configs page can only be built when a matching `docs/<name>.md` exists, so that link cannot dangle by construction.

mashehu (approved):
Just checked on the phone, but looked fine

PR #4391, opened 2026-09-07
Organise `ignored_repos.yaml` into two groups and sort

ewels:
Split `ignore_repos` into early-stage pipelines (sorted by repo creation date, oldest first, each annotated with its creation date) and non-pipeline repos (sorted alphabetically). No repos added or removed.

Hopefully this makes it a bit easier to keep tabs on the early-stage pipelines so that we can check whether they have progressed enough to be removed from the list and shown on the website.


ewels:
Minor change with no functional difference. Self-merging as CI is passing.

